PR #10, opened 2020-06-19
Fix initial offline state

graywolf336:
- Removed the default thumbnail from git
- When cleaning up the assets, remove the thumbnail as well
- Generate a default thumbnail when starting up from the logo

Feel free to reject this pull request, I just found this to be a clean way of doing it


gabek:
haha whoops, we just crossed paths in fixing that initial thumbnail copy.  I also changed the format to jpg since it seems a bit faster to write.  If you want to pull those changes I think removing the thumb at launch makes sense.

graywolf336:
That is actually quite comical we crossed paths like that 😂  just merged and switched to the jpg.

PR #17, opened 2020-06-19
Update the module path

graywolf336:
This was messing me up while working on the restructure 😂  at first I was like "why is it pulling in `video-stream` but then I checked this file and it made sense haha


gabek:
Haha, whoops!  Thanks!

PR #18, opened 2020-06-20
Project restructure

### LINKED ISSUE #16: Project Restructure
What would be your thoughts on the following project structure:

```bash
├── config 
│   └── config.go
├── controllers (http controllers)
│   ├── chat.go
│   └── static.go
├── core (the logic lives here)
│   ├── ffmpeg/
│   │   ├── offline.go
│   │   └── start.go
│   ├── rtmp/
│   │   ├── handler.go
│   │   └── server.go
│   ├── storage/ (implementations)
│   │   ├── ipfs.go
│   │   └── aws.go
│   ├── storage.go (interface)
│   ├── core.go
│   ├── stats.go
├── models (definitions for how data will be handled)
│   ├── stat.go
│   ├── server.go
│   └── etc.go
├── utils
│   └── utils.go
├── static (stays the same)
│   └── logo.png (etc)
├── webroot (stays the same)
│   ├── img/
│   ├── js/
│   ├── static/
│   ├── styles/
│   ├── vendor/
│   └── index.html
├── go.mod
├── go.sum
├── main.go
├── config.sample.yaml
├── README.md
├── Dockerfile
└── .gitignore
```

This is a massive difference from the current structure of the project and I am willing to start the PR to move it and work on the restructuring. Curious to hear your thoughts

graywolf336:
Closes #16 

@gabek would love your feedback on this! And would like for you to try it before you merge. I tried my best to keep everything how it was, but I did take a little bit of liberty with small things and will happily answer any questions you have.

...yes, I did get too caught up in the idea and spent my evening doing it. 😊


gabek:
That's so awesome!  I'll totally check it out and give it a spin.

core/ffmpeg/verifyInstall.go
```diff
@@ -0,0 +1,32 @@
...
+		return fmt.Errorf("error while verifying the ffmpeg path: %s", err.Error())
+	}
+
+	if stat.IsDir() {
+		return errors.New("ffmpeg path can not be a folder")
+	}
+
+	mode := stat.Mode()
+	//source: https://stackoverflow.com/a/60128480
+	if mode&0111 == 0 {
```
gabek, line 27: Nice touch!

core/playlist/monitor.go
```diff
@@ -0,0 +1,161 @@
...
+
+var (
+	_storage models.ChunkStorageProvider
+	variants []models.Variant
+)
+
+//StartVideoContentMonitor starts the video content monitor
+func StartVideoContentMonitor(storage models.ChunkStorageProvider) error {
+	pathToMonitor := config.Config.PrivateHLSPath
+
```
gabek, line 29:
```suggestion
_storage = storage
```

core/storageproviders/ipfsStorage.go
```diff
@@ -38,27 +42,32 @@ type IPFSStorage struct {
 	gateway       string
 }
 
-func (s *IPFSStorage) Setup(config Config) {
+//Setup sets up the ipfs storage for saving the video to ipfs
+func (s *IPFSStorage) Setup() error {
 	log.Println("Setting up IPFS for external storage of video. Please wait..")
 
-	s.gateway = config.IPFS.Gateway
+	s.gateway = ownconfig.Config.IPFS.Gateway
```
gabek, line 49: I know in other places we're using the default `config.Config`, and from pure aesthetic reasons something like this `ownconfig.Config` looks nicer.  I guess we could do something also like `config.Current` or something too.

main.go
```diff
@@ -1,121 +1,33 @@
...
+	log.Println(core.GetVersion())
 
-		if path.Ext(r.URL.Path) == ".m3u8" {
-			clientID := getClientIDFromRequest(r)
-			stats.SetClientActive(clientID)
-			disableCache(&w)
-		}
-	})
+	//TODO: potentially load the config from a flag like:
+	//configFile := flag.String("configFile", "config.yaml", "Config File full path. Defaults to current folder")
```
gabek, line 16: 👍

gabek:
There's too many things to go through and mention how awesome they are, so in general, **THANK YOU**!  

Sorry about the couple changes in master that cause a conflict with these updates.  Should be small updates, though.

There was only one real issue, and I committed that missing assignment to `monitor.go`.

You went above and beyond just re-organizing, that's very cool and so appreciated!  It's so nice to have everything in the respective modules and passing around errors properly.

I figure instead of going through those TODOs and answering the questions or fixing I'd just do that separately as to not hold this reorganization up.

core/storageproviders/ipfsStorage.go
graywolf336: I use the `config.Config` convention pretty often, so I hadn't thought twice about it. But I do agree with you here, the `ownconfig.Config` makes sense. The reason I switched in this file is because it already imported another `config` package. 🤔 

I'm conflicted on which one I like better as what you suggested makes sense, `config.Current`

.gitignore
```diff
@@ -15,10 +15,11 @@
 vendor/
 
 #owncast specific
-config/config.yaml
-config/stats.json
```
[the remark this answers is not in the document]
graywolf336: I think that would be a good route to go. Another thought too is that instead of using json files, could switch to an embedded database such as [boltdb](https://github.com/etcd-io/bbolt). Then expose the data via endpoints. I've used boltdb in several projects and have one that is in production in several business on site (actually using raspberry pi's). If you're interested in going that route, we could explore our options for an embedded database instead of flatfiles.

core/ffmpeg/verifyInstall.go
graywolf336: Only thing I am worried about is whether or not this works on Windows too or if we should add a special handler for Windows 🧐

core/playlist/monitor.go
graywolf336: Ah, good catch!

graywolf336:
You're most welcome!! I am genuinely interested in this project to act as a self-hosted rtmp to hls converter for a project of mine. We currently stream to restream and have a mobile app and if we had an HLS stream we could have a "native stream" inside our app (we use a service for the App due to several things out of my control). So, I will be using just the service side of it and not the UI. :)

I will merge master down and resolve the conflicts so we can see about getting this merged shortly. :)

.gitignore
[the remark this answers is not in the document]
graywolf336: Let me know your thoughts on this as doing more persistent storage opens a wide range of options. The primary one that comes to mind is sessions. A session would represent one streaming time frame and could have the ability to see stats, messages, and other details about it.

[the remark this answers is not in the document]
gabek: Bolt is nice and straightforward, I've used it once previously as well.  The reason I wanted to use parsable flat files was simply to allow flexibility for people who might want to use that data for something and not need a dependency to read it.  Like stats, for example, I don't know what somebody would want to do with it, but I figured I'd make it easy to have those values accessible instead of requiring an API  to consume it.  But maybe that's solving for a problem that doesn't exist :)  It's something to think about, though, depending on how much stuff needs to be persisted going forward.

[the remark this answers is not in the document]
graywolf336: From that perspective, it makes sense...but at the same time, if I'm hosting this I think it would be easier to consume an api than to read the json files. My thought is that most people will host this on another server aside from their other ones since it would require more processing power. Which in that instance, it would be easier to consume an api than to read the files themselves. 🤔 Like you said, though, it is something to think about and evaluate in the future.

graywolf336:
Is there anything that needs to be done for this to be merged? Or you wanting to test it more?

gabek:
Awesome!  Let's merge and push forward.  Super exciting, thanks again!

PR #23, opened 2020-06-23
Fix a memory leak with the chat

graywolf336:
Essentially, the for loop of the read listener wasn't being returned out of and that caused the read listener to never be let up and released to the garage collector.


gabek:
Nice catch!

PR #27, opened 2020-06-24
New video transcoder

gabek:
Big change here that still needs some usage and testing.

In general this adds a new Transcoder that is far more customizable, and can handle different combinations of HLS variants, including the ability of having one lower-quality transcoded variant, along with the non-transcoded video that is passed along from the input.

The Transcoder is now standardized to allow for customizing input, output, quality per variant, and resizing (while keeping aspect ratio as long as you only supply a new height *or* width).

I also feel like this refactor is the next step to looking into https://github.com/gabek/owncast/issues/9, where we try and pulls status of generated files directly from ffmpeg instead of polling the filesystem.


core/rtmp/handler.go
```diff
@@ -76,8 +76,8 @@ func (h *Handler) OnPublish(timestamp uint32, cmd *rtmpmsg.NetStreamPublish) err
 	}
 	h.flvEnc = enc
 
-	//TODO: why is this turned into a goroutine?
-	go ffmpeg.Start()
+	transcoder := ffmpeg.CreateTranscoderFromConfig()
+	go transcoder.Start()
```
gabek, line 80: This is run under a Goroutine because the execution is blocking, `ffmpeg` will keep running.  Let me know if I'm missing something or you have some other suggestions!

core/ffmpeg/transcoder.go
```diff
@@ -0,0 +1,285 @@
...
+	variant.SetVideoBitrate(strconv.Itoa(quality.Bitrate) + "k")
+	variant.SetAudioBitrate(strconv.Itoa(quality.AudioBitrate) + "k")
+	variant.SetVideoScalingWidth(quality.ScaledWidth)
+	variant.SetVideoScalingHeight(quality.ScaledHeight)
+	variant.SetVideoFramerate(quality.Framerate)
+
+	return variant
+}
+
+// CreateTranscoderFromConfig will return a new Transcoder, populated by the config
```
gabek, line 139: I don't really like this implicit initialization via the config, instead of the config being explicitly passed in, but we're pulling the config elsewhere, so I thought I'd go that route, but I don't know if it should be kept.

config-example.yaml
```diff
@@ -1,18 +1,25 @@
...
 webServerPort: 8080
-enableOfflineImage: true
 
 videoSettings:
   chunkLengthInSeconds: 4
   streamingKey: abc123
-  encoderPreset: superfast # https://trac.ffmpeg.org/wiki/Encode/H.264
-  passthrough: true # Enabling this will ignore the below stream qualities and pass through the same quality that you're sending it
-  offlineImage: doc/logo.png # Is displayed when a stream ends
+  offlineContent: static/offline.m4v # Is displayed when a stream ends
```
gabek, line 9: This isn't a great name.  It's not clear that it's the content that is shown when the stream is offline.

config-example.yaml
```diff
@@ -1,18 +1,25 @@
...
-    - bitrate: 1000 # in k
+    # Pass through the exact video and audio that you're streaming.
+    - full:
+      videoPassthrough: true
+      audioPassthrough: true
+
+    # Transcode the video to a lower bitrate and resize
+    # - low:
+    # bitrate: 700
+    # scaledWidth: 600
```
gabek, line 20: You can now use `scaledWidth`, `scaledHeight` or both.  If you use one or the other it'll keep the aspect ratio.

config/config.go
```diff
@@ -15,28 +15,41 @@ import (
...
-type streamQuality struct {
-	Bitrate int `yaml:"bitrate"`
+type StreamQuality struct {
+	// Enable passthrough to copy the video and/or audio directly from the
+	// incoming stream and disable any transcoding.  It will ignore any of
+	// the below settings.
+	IsVideoPassthrough bool `yaml:"videoPassthrough"`
+	IsAudioPassthrough bool `yaml:"audioPassthrough"`
+
+	Bitrate      int `yaml:"bitrate"`
```
gabek, line 43: I left this as `bitrate` instead of `videoBitrate` because the config files already use `bitrate`, though I should probably just rename it.

gabek:
On my remote box with ffmpeg 4.2.1 I'm getting segfaults when I added a third bitrate, but with the exact same command on ffmpeg 4.2.2 on macOS it's fine.  I updated to 4.2.3 and it fixed it, so 4.2.2+ might have to be a required version?  Or maybe just 4.2.1 has a bug in it.

config/config.go
graywolf336: Yeah, renaming this to have the `video` prefix would probably be user friendly even though there is `audioBitrate`.

core/ffmpeg/transcoder.go
```diff
@@ -0,0 +1,285 @@
...
+	command := t.GetString()
+
+	_, err := exec.Command("sh", "-c", command).Output()
+	if err != nil {
+		log.Panicln(err, command)
+	}
+
+	log.Println("Video transcoder started.")
+
+	return err
```
graywolf336, line 71: I would return `nil` here as the system will panic if the err is not nil

core/ffmpeg/transcoder.go
graywolf336: I think you could just name it `NewTranscoder` unless you plan to create another way that someone can create a valid Transcoder to be used. Unless I am missing something, we currently assume everything is done via the config file.

gabek:
@graywolf336 One change I'd like to make is right now we move to an "offline" state when the user disconnects from RTMP.  When really what we want is to go offline when the transcoder exists.  There's an amount of seconds from the time the user stops broadcasting and when we stop distributing the video, so we should take that into account.

The requirements for this would be that we need to catch this "transcoder exited" from the call site of `Start()`, and not just fire something from the transcoder itself, because we use the transcoder to create the offline state, and that would create a loop of start/exits.

Would it make the most sense just to create a go channel around the go routine of `Start()` to catch an exit?

gabek:
I'm thinking about this more, and I think we should maybe leave it as is for the time being.  It's really two different things:

* When do we show the user we're offline?  When the transcoder is done.
* When do we, server-side, start firing offline logic?  When RTMP disconnects.

We should keep it simple for now and just show the user that we're "offline" a little early and add complexity around this later.

graywolf336:
Got a question for you and it's more related to the code style. I've noticed that on the Transcoder in particular you're exporting the properties of the trancoder while also generating functions which are setters for the transcoder. This is a different style than I've seen and experienced when it comes to go code. Usually if there is a setter function then what is being set is marked as private. Is there any particular reason for the style you're going with?

gabek:
No reason whatsoever.  Purely because I started with public properties and then decided it was best to create setters in case we want to intercept bad values or have other side effects.  I'll change them to private, there's no reason for them to be exported.  Thanks!

graywolf336:
@gabek thanks for the explanation! I figured that is why you was going with the setter functions, especially since it would be purposeless to set values when the transcoder has already started.

gabek:
I'm going to merge this in.  It'll break the old config file, but might as well push forward and catch any issues.  Though I've been running and testing on the goth.land test instance and it's been working well.

PR #55, opened 2020-07-12
Correct indenting

mattdsteele:
Ran into this when trying to use the example config directly


gabek:
Oh awesome, thanks Matt!

PR #67, opened 2020-07-17
Use endpoint for chat history instead of websocket

gabek:
This uses the `/chat` endpoint for bulk fetching of past chat messages instead of getting one message at a time of the history over the websocket.

### DIFF
--- a/webroot/js/message.js
+++ b/webroot/js/message.js
@@ -246,12 +246,14 @@ class MessagingInterface {
 	disableChat() {
 		if (this.formMessageInput) {
 			this.formMessageInput.disabled = true;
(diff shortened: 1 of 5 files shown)


webroot/js/message.js
```diff
@@ -261,4 +265,14 @@ class MessagingInterface {
 			jumpToBottom(this.scrollableMessagesContainer);
 		}
 	}
+
+	async getChatHistory() {
```
gingervitis, line 269: I would actually keep this function in app.js, where it handles all the other data fetching for the app.
I was hoping to just keep this object, mainly for enabling functionality in the DOM.

Issue #65, opened 2020-07-16
Support chat sidebar on tablets


gabek:
This is purely for selfish purposes because I have an iPad.  I would love to have the layout for an iPad, in landscape, be the same as a desktop browser since there's ample space.

gabek:
@gingervitis did this already too.

PR #68, opened 2020-07-17
Plz can we have desktop layout on wide enough tablets?

gabek:
Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2020-07-16 at 21 03 17

I think it looks ok?

I'm not saying we should go just by width, but I just wanted to test it.  But if it's ok to do this we can keep it.

For #65


gingervitis:
I'll refactor the markup/css for layout a bit..

PR #69, opened 2020-07-18
Show in the UI how long the user has been streaming for.

gabek:


webroot/js/app.js
```diff
@@ -223,8 +223,26 @@ class Owncast {
     }
 
     this.streamStatus = status;
+    this.setCurrentStreamDuration();
```
gingervitis, line 226: I would just put this call in this.handleOnlineMode().

webroot/js/app.js
```diff
@@ -223,8 +223,26 @@ class Owncast {
...
   };
   
+  setCurrentStreamDuration() {
+    // If we're offline then don't update any of the UI.
+    if (!this.streamStatus.online) {
```
gingervitis, line 231: If we only do this in handleOnlineMode, we won't need this here.

webroot/js/app.js
```diff
@@ -252,6 +270,8 @@ class Owncast {
     clearTimeout(this.disableChatTimer);
     this.disableChatTimer = null;
     this.messagingInterface.enableChat();
+
+    setInterval(this.setCurrentStreamDuration.bind(this), 1000);
```
gingervitis, line 274: We should create a new timer instance up top in the constructor for this interval counter., with all the other Timers.

We should make a `this.streamDurationTimer = null` in the contructor. 
Then here `this.streamDurationTimer = setInterval...`;

gingervitis: We'll have to make sure that in `handleOfflineMode()` we `clearInterval(this.streamDurationTimer)` .

gabek:
@gingervitis You told me to leave this and let you do it, so that's what I'm doing :P

PR #80, opened 2020-07-21
Add: Config Key when changing ffmpeg.

jeyemwey:
This is just a quick edit which saves the newbie the clicks to [configuration-example-full.yaml](https://github.com/gabek/owncast/blob/master/doc/config-example-full.yaml).


gabek:
Nice, thank you Jannik!

PR #91, opened 2020-07-28
Support separate S3 serving endpoint

### LINKED ISSUE #90: Different S3 serving endpoint (e.g. CDN support)
DigitalOcean supports a [built-in CDN](https://www.digitalocean.com/docs/spaces/how-to/enable-cdn/) when using their S3-compatible Spaces storage. After enabling, use of the CDN just requires changing the endpoint; e.g.:

* Origin: https://bucket.region.digitaloceanspaces.com
* Edge: https://bucket.region.cdn.digitaloceanspaces.com

It's also possible to configure a custom domain (using a subdomain you own, for example).

I can try to add a new configuration parameter to support this (something like `s3.servingHostname`), but wanted to get your thoughts before making the change.

Simplest approach would be to just generate and return the custom URL within `s3Storage.Save`, but if you think there's value in pulling this up more generally, we could probably move it into `monitor.go`. This would potentially let other storage providers use the same param. I don't think this would be valuable for local serving, but maybe for other providers in the future? I dunno.

mattdsteele:
Fixes #90

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -86,7 +86,7 @@ type s3 struct {
 	Secret          string `yaml:"secret"`
 	Bucket          string `yaml:"bucket"`
 	Region          string `yaml:"region"`
-  ACL             string `yaml:"acl"`
+	ACL             string `yaml:"acl"`
 }
 
 func (c *config) load(filePath string) error {
--- a/config/config.go
+++ b/config/config.go
@@ -79,12 +79,13 @@ type files struct {
 
 //s3 is for configuring the s3 integration
 type s3 struct {
-	Enabled   bool   `yaml:"enabled"`
-	Endpoint  string `yaml:"endpoint"`
-	AccessKey string `yaml:"accessKey"`
-	Secret    string `yaml:"secret"`
-	Bucket    string `yaml:"bucket"`
-	Region    string `yaml:"region"`
+	Enabled         bool   `yaml:"enabled"`
+	Endpoint        string `yaml:"endpoint"`
+	ServingEndpoint string `yaml:"servingEndpoint"`
+	AccessKey       string `yaml:"accessKey"`
+	Secret          string `yaml:"secret"`
+	Bucket          string `yaml:"bucket"`
+	Region          string `yaml:"region"`
 }
 
 func (c *config) load(filePath string) error {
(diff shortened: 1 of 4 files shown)


core/storageproviders/s3Storage.go
```diff
@@ -85,6 +88,8 @@ func (s *S3Storage) GenerateRemotePlaylist(playlist string, variant models.Varia
 			fullRemotePath := variant.GetSegmentForFilename(line)
 			if fullRemotePath == nil {
 				line = ""
+			} else if s.s3ServingEndpoint != "" {
+				line = fmt.Sprintf("%s/%s/%s", s.s3ServingEndpoint, config.Config.PrivateHLSPath, fullRemotePath.RelativeUploadPath)
```
gabek, line 92: Could you use `config.Config.GetPrivateHLSSavePath()` instead of `config.Config.PrivateHLSPath`?  It takes into account defaults if people don't have it specified in their config.

gabek:
I tried it using the `transfer acceleration` endpoint of S3, uploading to the same place, but using `https://gabevideo.s3-accelerate.amazonaws.com` as the replacement downloading host, and it worked great.

The only change that needs to be made is using `config.Config.GetPrivateHLSSavePath()` instead of `config.Config.PrivateHLSPath`.  It uses defaults for config values that aren't explicitly set.

Thanks, Matt!

config/config.go
```diff
@@ -79,13 +79,14 @@ type files struct {
...
-	Region    string `yaml:"region"`
-	ACL       string `yaml:"acl"`
+	Enabled         bool   `yaml:"enabled"`
+	Endpoint        string `yaml:"endpoint"`
+	ServingEndpoint string `yaml:"servingEndpoint"`
+	AccessKey       string `yaml:"accessKey"`
+	Secret          string `yaml:"secret"`
+	Bucket          string `yaml:"bucket"`
+	Region          string `yaml:"region"`
+  ACL             string `yaml:"acl"`
```
graywolf336, line 89: Spacing issue 👀

mattdsteele: This is what I get for using the Web UI for merge 😶 I'll fix when I'm at my personal machine.

PR #92, opened 2020-07-28
Chat updates

gabek:


gabek:
I'll test this more later this week.

gabek:
Tested this updated chat branch with a bunch of people in a stream tonight and all seemed to work well.  Lots of chat happened and nobody seemed to be disconnected, not get messages, or not be able to send messages.  It's better than it was.  Success!

PR #97, opened 2020-08-02
Fix handling newlines in messages

nforro:
It's not possible to make a newline when already holding Shift, for example to type an uppercase character. Fix that.


gabek:
Just tested before and after, and it works!

For anybody else looking to test try:

<shift>TEST<ENTER>
Will create a newline.

gingervitis:
👍

gabek:
@nforro Thanks so much for your help and sending over this change! 🥇

Issue #48, opened 2020-07-12
Use Websocket for Status Communication


graywolf336:
Maybe it would be useful to use the websocket connection to be notified of the status updates? Instead of continuously polling?

gabek:
Agreed!  It's been something I've been thinking would be a worthwhile update, but just never filed anything for it.

graywolf336:
I'll see what I can accomplish this week :)

gingervitis:
This sounds good and makes more sense to me too.

Note to self: For the JS refactor, I've moved a lot of the logic around that handles Websocket and messaging functionality.  I'll make sure to keep track of this so I don't freak out when I get a merge conflict!

graywolf336:
@gingervitis if you would like, I can wait until your refactoring gets merged to master? That way it makes things easier for you

gingervitis:
Works for me @graywolf336 . I'm just not sure when I'll finish merging it in. I'm aiming for end of this coming weekend.

graywolf336:
Works for me!

PR #98, opened 2020-08-03
Add username tab completion

nforro:
This pull request adds tab completion on `@` character. It cycles through the list of matching usernames, Twitch style (or so I believe, it's been a while since I last used Twitch).
The list of usernames is taken from the current message history. That should be good enough for most purposes, I think.

I'm opening this as a draft to collect feedback. I'm not sure if there isn't a better way to pass the list of messages, for example.

### DIFF
--- a/webroot/js/message.js
+++ b/webroot/js/message.js
@@ -23,10 +23,10 @@ class Message {
 			emoji: true, 
 			openLinksInNewWindow: true, 
 			tables: false, 
-			strikethrough: false, 
 			simplifiedAutoLink: false,
 			literalMidWordUnderscores: true,
 			strikethrough: true,
+			ghMentions: false,
 		}).makeHtml(this.body);
 		const linked = autoLink(markdownToHTML, {
 			embed: true,
--- a/webroot/js/message.js
+++ b/webroot/js/message.js
@@ -51,6 +51,7 @@ class MessagingInterface {
 		this.messageCharCount = 0;
 		this.maxMessageLength = 500;
 		this.maxMessageBuffer = 20;
+		this.chatUsernames = [];
 
 		this.onReceivedMessages = this.onReceivedMessages.bind(this);
 		this.disableChat = this.disableChat.bind(this);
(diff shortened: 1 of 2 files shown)


gabek:
This is awesome, thanks for sending it along!

I agree that the chat history is good enough for this purpose.

Some thoughts I had off the top of my head:

Instead of the messaging interface reducing the messages to usernames at the time they try to auto-complete, would it make sense to keep a list of usernames and debounce update it when messages come in?  Then the `app` can pass those usernames to `messaging`, without `messaging` having to reach back into the vue app.  This could also allow for something like displaying the number of people who are chatting, or even a userlist if we ever wanted either of those things (not something we need right now, though).

I'd love to hear what @gingervitis thinks.

gingervitis:
With what the current messaging setup looks like, I think the more straight forward way to do this would be to do something like-

Within the `MessagingInterface` class we could capture all the author names into a local array from the newMessages in the `onReceivedMessages()` method, so we don't have to parse through the message list at the vueApp level and reduce it every time we type.   

We can gather all the possible names from _all_ the newMessages when the app first loads, then on every subsequent message that comes in, we can just look for the author from the latest one.

nforro:
Thanks for the suggestions, I've updated the commit.

nforro:
Adn I added another commit that appears to resolve issue #100.

gabek:
Awesome!

I see you're setting the class on the entire message if `isHighlighted`.  Do we want the entire message highlighted, or just the username?  I was experimenting with doing this in`formatText()` and just highlighting the username:

https://github.com/gabek/owncast/commit/ed8b4100bd2d1c5de5d8d3d88c684beb74f822e4#diff-9ef4ffb517508e5dbb07b9537d2b6fb8R38

I'd also prefer to not query the DOM to find the username, but I think the current username isn't available elsewhere right now.  I was thinking of passing it into `formatText()`.

I'll defer to what @gingervitis thinks :)

nforro:
> Do we want the entire message highlighted, or just the username?

I'm fine with either. I did it this way because it seemed cleaner to me. Perhaps we could even do both?

> I was experimenting with doing this in`formatText()` and just highlighting the username:
> 
> [ed8b410#diff-9ef4ffb517508e5dbb07b9537d2b6fb8R38](https://github.com/gabek/owncast/commit/ed8b4100bd2d1c5de5d8d3d88c684beb74f822e4#diff-9ef4ffb517508e5dbb07b9537d2b6fb8R38)

Beware that this will match also parts of other words or other users' names (it's not that uncommon to have usernames where one is a substring of the other).

> I'd also prefer to not query the DOM to find the username

I don't like that any more than you do. There has to be a better way to do it, but I'm not familiar with *Vue.js* at all, so I'd be glad for pointers.

gingervitis:
Yeah it makes sense to do the formatting in `formatText()`.

I'm not that familiar with Vuejs either, also not a fan of digging into the DOM. I'd also like to try and keep all the Messaging functionality related things within MessagingInterface if possible.

It may be worth moving the web socket functionality out of app.js and into MessagingInterface class instead, so everything is handled more locally there. (plus the web socket only seems to be used for messaging right now anyway). From there we can have more flexibility to manipulate the incoming messages and format them as needed.

We'll just have hack a way for the MessagingInterface class to go back out and update the vueApp.messages queue. :/

gabek:
I'd like to split out the websocket into its own module that can be passed to other components, since it won't just be for messaging forever.  #48 is already on the table to use it for status updates, so it's best to split it out.

gabek:
With regards to matching other parts of words with the username, I actually did that on purpose because I wanted to match things like "@name", "name:" "name," etc.  And since it's just highlighting it's no big deal unless you're somebody who makes their name the letter "a" or something dumb.

As for the username thing, I say we just go forward with the less than optimal solution of grabbing the value from the DOM and re-address it in any future web refactor.

So I vote that we just highlight the username, not the entire message, and if you do the highlighting in `formatText()`.

gabek:
Just because it was mentioned in here, #104 was just merged in where Websocket is split out of the UI.  It shouldn't make a difference with this PR, but I wanted to bring it up anyway :)

gabek:
It's working great!  I'll wait for @gingervitis to sign off on it, but it does what it says.

gingervitis:
Works for me, thanks for working on this @nforro !

PR #104, opened 2020-08-05
Websocket refactor: Pull it out of the UI and support callbacks

gabek:
Regardless with how we go forward with the web discussion, this change needed to happen.

Right now the websocket is embedded into the app + chat UI components.  This pulls the websocket out of the UI completely and any component can register itself as a `listener` to get callbacks for websocket events.

Right now the only listener is `app` because it updates `messages`.

## How to get callbacks.

Pass an object to `addListener`.  For example:
```
this.websocket.addListener(this);
```

And then implement a callback method such as `rawWebsocketMessageReceived(msg)` or `websocketDisconnected()`.  All callback methods are optional and objects (not strings) are passed along.

## How to send a websocket message

Pass the `send()` method to the component who needs it.  In this case `messagingInterface` needs it.  And then it can just fire `send(msg)` with an object (not a string).
```
this.messagingInterface.send = this.websocket.send;
```

## How to test

Just run it like normal, and it should function exactly the same as it has.

### DIFF
--- a/webroot/js/app.js
+++ b/webroot/js/app.js
@@ -1,6 +1,23 @@
 import Websocket from './websocket.js';
 import { MessagingInterface, Message } from './message.js';
 import SOCKET_MESSAGE_TYPES from './chat/socketMessageTypes.js';
+import { OwncastPlayer } from './player.js';
+
+const MESSAGE_OFFLINE = 'Stream is offline.';
+const MESSAGE_ONLINE = 'Stream is online';
+
+const TEMP_IMAGE = 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
+
+const LOCAL_TEST = window.location.href.indexOf('localhost:') >= 0;
+
+const URL_PREFIX = LOCAL_TEST ? 'http://localhost:8080' : ''; 
+const URL_CONFIG = `${URL_PREFIX}/config`;
+const URL_STATUS = `${URL_PREFIX}/status`;
+const URL_CHAT_HISTORY = `${URL_PREFIX}/chat`;
+
+const TIMER_STATUS_UPDATE = 5000; // ms
+const TIMER_DISABLE_CHAT_AFTER_OFFLINE = 5 * 60 * 1000; // 5 mins
+const TIMER_STREAM_DURATION_COUNTER = 1000;
 
 class Owncast {
   constructor() {
--- a/webroot/js/app.js
+++ b/webroot/js/app.js
@@ -128,11 +128,11 @@ class Owncast {
   // websocket for messaging
   setupWebsocket() {
     this.websocket = new Websocket();
-    this.websocket.addListener(this);
+    this.websocket.addListener('rawWebsocketMessageReceived', this.receivedWebsocketMessage.bind(this));
     this.messagingInterface.send = this.websocket.send;
   };
 
-  rawWebsocketMessageReceived(model) {
+  receivedWebsocketMessage(model) {
     if (model.type === SOCKET_MESSAGE_TYPES.CHAT) {
       const message = new Message(model);
       this.addMessage(message);
--- a/webroot/js/app.js
+++ b/webroot/js/app.js
@@ -1,14 +1,16 @@
+import Websocket from './websocket.js';
+import { MessagingInterface, Message } from './message.js';
+import SOCKET_MESSAGE_TYPES from './chat/socketMessageTypes.js';
+
 class Owncast {
   constructor() {
     this.player;    
 
-    this.websocket = null;
     this.configData;
     this.vueApp;
     this.messagingInterface = null;
 
     // timers
-    this.websocketReconnectTimer = null;
     this.playerRestartTimer = null;
     this.offlineTimer = null;
     this.statusTimer = null;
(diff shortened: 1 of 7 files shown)


gabek:
Safari had some issues and required me pulling constants out of `utils.js` and localizing some of the values to the modules themselves.  I don't think this is a bad thing, since `utils.js` had a bunch of constants for random things.  But this now is changing and touching more things than I wanted to :\

webroot/js/websocket.js
```diff
@@ -0,0 +1,137 @@
...
+
+  /*
+  onMessage is fired when an inbound object comes across the websocket.
+  We pass these messages along to listeners by firing the
+  RAW_WEBSOCKET_MESSAGE_RECEIVED type callback.
+  If the message is of type `PING` we send a `PONG` back and do not
+  pass it along to listeners.
+  */
+  onMessage(e) {
+    const model = JSON.parse(e.data);
```
gingervitis, line 110: Let's wrap this in a try/catch in case e.data is not parseable.

webroot/js/websocket.js
```diff
@@ -0,0 +1,137 @@
...
+ * rawWebsocketMessageReceived(msg): Will send all raw JSON objects that come across the socket.
+ * websocketConnected/websocketDisconnected: Be notified on the connectivity state of the socket.
+ * 
+ * This class should not know anything about views.  Note that error states are not being passed to
+ * listeners because listeners shouldn't know or care about specific socket-level errors.
+ */
+
+import SOCKET_MESSAGE_TYPES from './chat/socketMessageTypes.js';
+
+const LOCAL_TEST = window.location.href.indexOf('localhost:') >= 0;
```
gingervitis, line 17: Why not keep this one in utils.js?  Look repeated across files.

webroot/js/app.js
```diff
@@ -109,53 +127,17 @@ class Owncast {
...
-        const message = new Message(model);
-        this.addMessage(message);
-      } else if (model.type === SOCKET_MESSAGE_TYPES.NAME_CHANGE) {
-        this.addMessage(model);
-      }
-    };
-    this.websocket = ws;
-    this.messagingInterface.setWebsocket(this.websocket);
+    this.websocket = new Websocket();
+    this.websocket.addListener(this);
```
gingervitis, line 131: Hm interesting. I was wondering how "this" whole thing was a "listener". But I guess this works.

webroot/js/websocket.js
gabek: That's the thing Safari had issues with.  Once I moved to a module then Safari didn't have those values available globally anymore.  But it was ok in Chrome.  If we do a refactor with modules then we can move them to a constants module or whatever you prefer so they're not duplicated.

Also we should get rid of local test.

webroot/js/app.js
gingervitis: This still feels a bit strange to me, maybe syntactically.. 
Could we instead do something more explicit and in a typical format such as: `this.websocket.addListener(CALLBACKS.WEBSOCKET_DISCONNECTED, this.someCallback)`.

gingervitis: I know it's a little more wordy, but I think it would help to read the code better and be able to see where some of the methods are going.

gabek: We could, that would be more like Javascript events?  And I guess we could just use random event emitters at that point, but those make me cringe, but that's the javascripty way of doing it I suppose.  I wanted to build a standardized interface that any object could implement, it is much cleaner in my eyes since all objects will follow the same rules, standardizing the names of things.  If you're an object and you want to know when the websocket connects you implement `websocketConnected`, always.  I prefer this over some objects having `websocketConnected` and others having `socketOpened` and others having `chatStarted`.

Though I admit I don't think Javscript does interfaces and/or protocols, but that's what I'm trying to do here.  A shared contract between the websocket library and those that use it.  But if that's confusing I can change it.

gingervitis: Well since it's still all javascript here, structure-wise it's much more understandable. We don't have to use actual cringey event emitters. The only events happening are the websocket's events. The standard between each new class that uses this Websocket class would only be being aware of its CALLBACKS.[event name]s, which makes sense to me and makes it more easy to remember to implement a callback for.

gabek: Ok so how about this, I'll do the Javascripty thing and use events.

Two changes:
* `Websocket` will no longer call callbacks, but instead emit events on the listeners.
```
const event = new Event('websocketConnected');
listener.dispatchEvent(event)
```

And listeners will no longer implement the required interface functions, but instead just listen for events and run whatever it feels like.

```
this.addEventListener('websocketConnected', function (event) {});
```

Or whatever the syntaxes are.  Isn't there some kind of `.on('event')` syntax too?

gingervitis: Oh I was saying we didn't have emit events.

gabek: Seems like the `on` syntax is a jQuery thing.  The benefit I see over events (even though they're ugly) is `App` won't have to know the inner workings of `MessagingInterface`.  Otherwise if we explicitly make `App` set callbacks for `MesagingInterface` then `App` has to know methods inside `MesagingInterface`, and that feels leaky.

gabek: Ok updates with the
```
this.websocket.addListener('rawWebsocketMessageReceived', this.receivedWebsocketMessage.bind(this));
```
Approach is in!

webroot/js/websocket.js
gabek: Done!

webroot/js/app.js
gingervitis: Looks good to me 👍

Issue #50, opened 2020-07-12
Web Player Uses Lowest Bitrate


graywolf336:
When testing, it appears that the web player is using the lowest bitrate even though I have ample bandwidth for the high quality video

gabek:
I popped your stream into the hls validator, attaching it here just for reference.  Looking at it to see if there's anything in it that would point to why that bitrate is being ignored.  (Server host name redacted in file).

[graywolf.txt](https://github.com/gabek/owncast/files/4908927/graywolf.txt)

gabek:
This is my current video config on goth.land:

```
    - high:
      videoBitrate: 1200

    - low:
      videoBitrate: 700
      scaledWidth: 600
      encoderPreset: superfast
```

And the high quality is being played on the web.  What does your config look like?

graywolf336:
Here is my current config

```
  streamQualities:
    - full:
      videoPassthrough: false
      audioPassthrough: true

    - low:
      videoBitrate: 700
      scaledWidth: 600
      audioPassthrough: true
      encoderPreset: superfast
```

gabek:
I just threw your config in and I got this resulting master playlist, with the two variants.

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:BANDWIDTH=2996224,RESOLUTION=1192x746
0/stream.m3u8

#EXT-X-STREAM-INF:BANDWIDTH=950224,RESOLUTION=600x376,CODECS="avc1.64001e,mp4a.40.2"
1/stream.m3u8
```

In playback it's pulling variant 0, as it should be.
![image](https://user-images.githubusercontent.com/414923/87367059-e89d1e80-c52e-11ea-9fc7-63d0b3fb73a3.png)

So something is different with what's being generated.  Could you take a look at your master `stream.m3u8` playlist and see if there's anything in there that could be explaining it?

gabek:
As an aside, we should add a quality selector.  I'll wait until the JS updates, but after that it would be a good addition.

https://github.com/silvermine/videojs-quality-selector

graywolf336:
Yeah, I'll take a look later today and provide it

gabek:
As an aside that videojs quality selector doesn't support HLS variants anyway, it's just for inline video sources.  Lame.

gabek:
I've been trying to keep an eye on if this happens for me.  I've seen it drop qualities, and sometimes stay there for a few minutes, but it always makes its way back up as I'd expect.

[...]

PR #106, opened 2020-08-05
Transcoder updates for bitrate validator warnings 

gabek:
This change aims to fix the HLS validator warnings that arise due to spikes in bitrate and hopefully resolve the issue that #50 sees where it's dropping to the lower quality bitrate even when the higher one is perfectly fine.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -197,6 +197,15 @@ func (c *config) GetVideoStreamQualities() []StreamQuality {
 	return _default.VideoSettings.StreamQualities
 }
 
+// GetFramerate returns the framerate or default
+func (q *StreamQuality) GetFramerate() int {
+	if q.Framerate > 0 {
+		return q.Framerate
+	}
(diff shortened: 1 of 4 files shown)


graywolf336:
Do you want me to use this branch tonight?

gabek:
That would be great!  I don't think this is a silver bullet, but it should help if my guess of what might be happening is involved.

gabek:
Ok, hopefully the state of that branch should keep the bitrate spikes at bay.   I'll run [mediastreamvalidator](https://developer.apple.com/documentation/http_live_streaming/about_apple_s_http_live_streaming_tools against your stream tonight and see what it has to say.

gabek:
The good news is this fixes the HLS validator bitrate spike errors.  The bad news is it makes no difference with #50.  But hey, it's a net positive to keep players happy!  Merging it in.

PR #108, opened 2020-08-06
Merge of emoji + autolink + embed + etc

gabek:
This branch has the WIP emoji picker in it, so it's impossible to review because emoji picker isn't in master yet.  So wait until it is.

This also has an approach of string highlighting (for username highlighting) but it's not being used while we work through #98.

---

This change has the new version of:
* Autolinking and formatting URLs in chat messages.
* Embedding YouTube, Instagram, and image URL content.


gabek:
I changed this PR to merge into the `gek/emoji-picker` branch so it's at least readable as I continue to tweak.  @gingervitis if you have any changes in `gek/emoji-picker` locally could you make sure they're pushed up so I'm working off the current state of the chat stuff?

gingervitis:
Oops just saw this.
I had a bunch of changes, but they weren't in a working state yet. I just some sticker updates only.

gabek:
image

gabek:
Getting all these chat enhancements in.  I look forward to being able to pull out a lot of these util functions and put them in their own module once we organize code into modules.

* Reworked tab completion of usernames to be compatible with the new emojified input field (div instead of text input).
* Autolinking of URLs
* Embeds for Instagram, image URLs and YouTube.
* Emoji picker

* If the text is just an embeddable URL or photo then it's stripped from the text of the message.
* We need to resize the instagram embed to something.  Not sure what that should look like but we can figure it out.
* Emoji picker probably doesn't work with mobile.  We can deal with that separately.

gabek:
Filed #109 to fix placeholder text.

gabek:
Ginger is going to be in this code for the web refactor anyway, so merging it in and it can continue to be iterated on.

Issue #78, opened 2020-07-21
Show some basic statistics in the terminal


gabek:
I find myself wanting to know some basic values often enough it might be worth displaying these in the terminal.  I'm kind of thinking along the lines of what ngrok does.

![image](https://user-images.githubusercontent.com/414923/88082634-dc205380-cb36-11ea-8cfc-5a1cbea9ed1a.png)

For Owncast it would be nice to surface:

* Machine CPU usage
* Machine memory usage
* Number of current viewers
* Stream duration
* Video configurations currently in use

Maybe using something like https://github.com/gdamore/tcell or https://github.com/gizak/termui ?  This idea might be overkill, but it would be pretty user-friendly and useful.  Maybe I'll experiment sometime after the 0.0.1 release.

gabek:
After hearing feedback from different people on #111 I've decided to change gears and go with the following game plan:

1. Drop the idea of a full terminal performance/configuration dashboard.
1. Add some simple error events based on hardware utilization to alert the service operator things aren't going well (cpu spikes mostly).
1. Plan for a future lightweight web dashboard that has more detailed statistics to fill the need I was trying to fill with the terminal version.  This would include creating a new `metrics` package.
1. Keep getting feedback on what values would be useful to expose in different ways.  @graywolf336 mentioned prometheus metrics as an option.  But I figure we can wait until we have our own little dashboard that works for everyone out of the box before we start doing specific implementations for more niche use cases.

PR #111, opened 2020-08-11
Terminal stats and status UI

gabek:
I've been wanting to add a terminal based UI to show server status and stats.  I think I might have gone a little overboard compared to what I was originally thinking, but I think it's kind of nice.

* It's using [termdash](https://github.com/mum4k/termdash/) as a console widget framework.  It actually has mouse support and you can scroll text.  Pretty wild.  But because of that you can't highlight text, and that's pretty annoying.

* This UI is not enabled by default.  You must pass `-enableTerminalUI` when running the service.

* There's probably lots of improvements that can be made as far as layout, appearance, formatting, etc.

* Log messages are intercepted in termui mode and logs them to this new UI instead of stdout.

Screen Shot 2020-08-10 at 8 56 49 PM


gabek:
The X axis on the charts have useless values.  Time would be useful, but those values take up space.

gabek:
@jeyemwey @graywolf336 I'd love to hear your thoughts if you happen to have a chance to give this branch for a spin and run with `-enableTerminalUI`.

Some updates I'm going to make

* "Clients" title should have the count in it
* Memory usage should probably be RAM usage in actual values, probably gigs, instead of % of system memory.
* X values I think should be the point in time that the value was drawn, in seconds since the service has started?

jeyemwey:
I am not sure if I can answer until next week, but from what I see, this looks great!

________________________________

gabek:
Absolutely no rush, thanks!  I'm going to keep tweaking anyway.

jeyemwey:
Hi Gabe, overall I really like it! However, I have some remarks:

1. I don't think that publishing the Stream Key is necassary. Usually, I have set up my streaming box beforehand, and don't need other people to take over the stream.
2. The refresh rate for window size changes is quite low, can this be improved?
3. The video variants and the configuration tabs could use some more space, if my terminal is full screen (which i rarely ever do), it still does not have enough space to show all config items: 
grafik For my production servers, I use other services to monitor CPU and RAM load, so I don't need to see these charts.

gabek:
Thanks for checking it out!  

* That's fair to remove the stream key, it's a convenience but it's also a security issue if somebody were to gain access to that terminal.
* I'll take a look at resizing refresh, but I think it's only refreshing when I update & redraw the screen, and redraws are every few seconds.  It's probably not worth redrawing more often for the rare occasion that somebody is resizing their window.
* I'll have to find a new layout, since it's clear the current one doesn't fit for all users.  But in general I don't see wanting to add configuration around showing and hiding things.  I kind of wanted a basic most important stats so out of the box people can see if performance is hurting.  But for people like you who have other methods, I think it's best just to use your existing tooling.

Ruxton:
I really like this but I feel like it belongs in a client package for a few reasons..

* Why compile all that UI code and carry it around on a server i'll never use it on
* It's way more useful as a client connecting to the already running server/daemon, even if that's only local.
* Exposing an interface in the 'server' that does this, allows others to just develop their own interface to that

gabek:
I don't disagree.

Since there's a very real correlation between performance, hardware capabilities and your specific configuration I wanted people to see these things out of the box.  If somebody loads up a bunch of video variants that they can't handle and sees it isn't performing well, they're likely just to give up on the service instead of looking deeper into it or realizing there's anything they can do about it.  Many likely don't even know how their hardware is performing.  So helping people have a good experience "out of the box" is important to me.

An alternative to this instead of built-in proactive monitoring is to enable reactive alerts.  Start throwing errors when the machine is in the red and make some suggestions on where to look to fix it.  Another logical idea would be to have a web status page with the same info as the this terminal version, but that's a whole other thing.  But that's something to think about.

But I don't want to make people install additional secondary tools and add complexity and confusion.  If they require 3rd party monitoring integrations they might be better off looking for something a bit more piecemeal that they can customize on the operations side.

I think I'll give the performance-based errors a spin.  Continue to monitor the hardware silently, but let people know when things get in the red because that means video segment generation will end up suffering, and they can be pointed to some documentation on how to fix it.

Open to what other people think as well!

graywolf336:
I think if it's not enabled by default but well documented, I am okay with this and would love to have this so I can get the most out of a machine while also not hurting the performance of a live stream.

One alternative I would like to see is maybe prometheus metrics since that is close to an industry standard. I have an infrastructure which is in place and might be moving my owncast server underneath that and we already have prometheus data scrapers for our other services/servers. Would be amazing to see this data in there!

Which leads me to the following train of thought: what if you created a `metrics` package which then other packages pulled information from. So the prometheus package could pull from it and so could the terminal ui?

gabek:
Here's what I put in #78, so I'll just copy it here:

---

After hearing feedback from different people on #111 I've decided to change gears and go with the following game plan:

1. Drop the idea of a full terminal performance/configuration dashboard.
1. Add some simple error events based on hardware utilization to alert the service operator things aren't going well (cpu spikes mostly).
1. Plan for a future lightweight web dashboard that has more detailed statistics to fill the need I was trying to fill with the terminal version.  This would include creating a new `metrics` package.
1. Keep getting feedback on what values would be useful to expose in different ways.  @graywolf336 mentioned prometheus metrics as an option.  But I figure we can wait until we have our own little dashboard that works for everyone out of the box before we start doing specific implementations for more niche use cases.

Issue #114, opened 2020-08-18
Web admin dashboard


gabek:
# Admin dashboard v1: Read-only.

In hopes of having an easy to use place to look at performance and current configurations out of the box, I propose a web dashboard with statistics and configuration details.  See #111 for a previous approach.

The initial use case I'm looking to solve for is somebody setting it up for the first time and doesn't see the direct connection between how you configure the transcoder and how the CPU is used.  It's good to see that if you're maxing out the CPU right next to a list of your video configurations to see that maybe you should remove one of them.  Or on the flip side seeing that you have a lot of breathing room on your hardware and you can experiment with adding more flexibility with video options.

Feedback is highly appreciated since real-world use is what should drive this feature.  What would you like to see on an admin dashboard that would make your life easier when running Owncast?

# Current plan and status

## Home page overview
- [x] List of active video quality variants
- [x] Currently connected client count
- [x] Peak client count for this stream
- [x] Inbound stream details/broadcaster
- [x] Stream duration
- [x] "Offline" / No stream connected state
- [x] Hardware/transcoding/upload alerts?  See "Logs" below.

## Hardware metrics
These can be a single graph or individual for each metric.  Depends what is more useful.
- [x] Current CPU usage
- [x] CPU usage over time graph
- [x] Current RAM usage
- [x] RAM usage over time graph
- [x] Current disk usage
- [x] Disk usage over time graph

## Viewers
- [x] Currently connected client count
- [x] List of currently connected clients with details (duration, username, geo, etc)
- [x] Connected client count over time graph

## Config
- [x] List of high-level config values (name, description, logo, stream key, social links, etc).
- [x] Video configuration with the different video variants configured and other video related details.
- [x] Storage configuration with external storage config if enabled, otherwise a message stating that local storage is used.

## Logs
- [x] Server logs
- [x] Hardware alerts
- [x] Slow upload alerts
- [x] Slow transcoding alerts

## Overall
Once we get all the values on the pages we should attempt to make a cohesive looking UI that people would want to to look at.
- [x] Layout
- [x] Polish

## Metrics package and corresponding API

Collect the above values in a metrics package, and expose them via a new API for the web to consume.

## Auth

HTTP BASIC auth.  Username: `admin`.  Password: is the current stream key

## Serving it

The web server integration is being tracked in https://github.com/owncast/owncast/issues/255.

graywolf336:
[...]

PR #115, opened 2020-08-18
Simple hardware metrics collection + alerting

gabek:
Add some simple hardware utilization alerts in the console if the CPU or RAM usage is high.

While the metrics are just used for alerting right now, the values can be used for #114 as well.

If anybody has any thoughts about this, please let me know!


mattdsteele:
Semi-related: if we wanted to capture usage metrics (like # of concurrent viewers, # of chat messages, etc) into something like Influx/Prometheus/OpenTSDB, what's the way folks implement that? Add code to Owncast that collects these and sends to a hardcoded endpoint per service? I'm not real familiar with anything outside Java metrics and the JMX protocol.

gabek:
External metrics have come up a couple times now, and I honestly don't know the correct answer.  Not that we, or somebody else, couldn't build support for it, but it seems like adding as core functionality such specific integrations doesn't fit the vibe of the project.  Plus once you add one specific integration then it's a slippery slope to all the others.

I've been toying with the idea of some kind of adapter service as a standalone middleman.  Like you'd define the URL to this adapter in Owncast and then it will send off metrics without knowing or caring what the specific integration is.  And then it's up to the adapter to get those off to Influx, Prometheus, etc.  Instead of plugins, it's more of a plug-out.  Then, in theory, people could build whatever they wanted by simply handling that payload.

There's also the `stats.json` file that simply gets written to disk.  I did this on purpose in case somebody wanted an easy way to read in those values from outside of the service as well as being cheap persistence.  This could be expanded upon and other things could poll its contents and do things with it.

gabek:
Opened #128 to discuss metrics in external tooling.  Feel free to discuss!

PR #118, opened 2020-08-20
Remove cache-control header from HLS queries

### LINKED ISSUE #117: CORS preflight requests fail when Cache-Control header is set (DigitalOcean Spaces)
Related to the issue I mentioned at https://github.com/gabek/owncast/issues/52#issuecomment-674559612

When investigating the differences between the Owncast player and the player on my website, I noticed I don't have this code (from player.js):

```
    videojs.Hls.xhr.beforeRequest = function (options) {
      const cachebuster = Math.round(new Date().getTime() / 1000);
      options.uri = options.uri + "?omgwtf=" + cachebuster;
      options.headers = {
        'Cache-Control':'no-cache'
      };
      return options;
    };
```

When I added this in to my website (more speciically, setting `options.headers`), it began to fail.

Looking through Dev Tools, I notice the [CORS preflight request](https://developer.mozilla.org/en-US/docs/Glossary/preflight_request) includes this header:

```
Access-Control-Request-Headers: cache-control
```

Which results in a 403 Forbidden when trying to pull the files from the Spaces bucket. Similarly, if I remove the headers from the Owncast UI, it works fine.

In the Spaces configuration, there's an option to set "Allowed Headers". When I add in "Cache-Control", it works fine:

image

So! I see two approaches we could take:

1. Document that some S3-compatible providers will require you to add this header to their HTTP configuration
2. Remove the `options.headers` from the `beforeRequest` check

I'm cool putting in a PR for either, but I'd lean toward option 2; mainly because I'm not sure setting the headers is even necessary; given the cache-busting query param we're adding. WDYT?

mattdsteele:
Also minor refactor around cachebusting. I considered changing the key to `omgwtfbbq` but thought better of it 🙃 

Fixes #117

### DIFF
--- a/webroot/js/player.js
+++ b/webroot/js/player.js
@@ -48,10 +48,7 @@ class OwncastPlayer {
   init() {
     videojs.Hls.xhr.beforeRequest = function (options) {
       const cachebuster = Math.round(new Date().getTime() / 1000);
-      options.uri = options.uri + "?omgwtf=" + cachebuster;
-      options.headers = {
-        'Cache-Control':'no-cache'
-      };
+      options.uri = `${options.uri}?cachebust=${cachebuster}`;
       return options;
     };
 


gabek:
Awesome, thanks Matt!

PR #120, opened 2020-08-24
frontend refactor with Preact

gingervitis:
Total refactor of the frontend using Preact.

- `index.html` now just loads app.js
- also created `index-standalone-chat.html` that displays all the messages only from the chat.
- created a `video-only.html` starter page for just a potentially embeddable iframe page of just the video. (#110 )
Each of these top level entry points pulls in a top level component which manages things like web socket and external data fetches and saves them to state and passes them down to child components, using familiar React lifecycle patterns.

Other stuff
- created individual components for the chat pieces and some helper components with Preact
- reorganized all the js files into a hopefully easier to parse file structure (for myself)
- cleaned up css styles to use mostly Tailwind styles (since we happen to include it).
- @gabek moved all the favicons to a separate directory

@mattdsteele @graywolf336 let me know what you think.
@gabek please test to see if all the things still work!

I may have a couple outstanding small style bugs that we can tweak later.

Closes #96

PR #121, opened 2020-08-25
Player loads path from Config

Ruxton:
Loads the HLS path from the config file

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -30,14 +30,15 @@ type config struct {
 
 // InstanceDetails defines the user-visible information about this particular instance.
 type InstanceDetails struct {
-	Name          string            `yaml:"name" json:"name"`
-	Title         string            `yaml:"title" json:"title"`
-	Summary       string            `yaml:"summary" json:"summary"`
-	Logo          map[string]string `yaml:"logo" json:"logo"`
-	Tags          []string          `yaml:"tags" json:"tags"`
-	SocialHandles []socialHandle    `yaml:"socialHandles" json:"socialHandles"`
-	ExtraInfoFile string            `yaml:"extraUserInfoFileName" json:"extraUserInfoFileName"`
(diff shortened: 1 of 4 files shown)


gabek:
Hey there!  So there's some bad timing.  The web is going through an overhaul and there's a PR up for that work at #120.  Any changes to the web code will need to move in that direction.  I'm guessing there won't be too many changes to that refactor work from now on, so feel free to apply your updates to it and we can test out your change in the context of the new web code, or feel free to wait until that PR gets merged in.

Ruxton:
@gabek yeah i noticed that and tried to keep the PR minimal for that reason. Holding onto P2P integration atm too, waiting for that rework to go through before sending it in.

gabek:
The "private" path is used for local HLS storage and is not meant to be used for any public access.  If what you're doing is trying to surface the public web accessible path for files then a path in relation to `webroot` would need to be exposed.  We should maybe change `PublicHLSPath` to `PublicWebPath` and then `PublicHLSPath` would be a subdirectory relative to `PublicWebPath`, and that's what you'd expose in the API.

So an example if you customized the values:
PublicWebPath = `/home/user/owncast/public`
PublicHLSPath = `hls`

We could also just hard code the subdirectory to `hls` always, because customizing that doesn't really have much of a point, because what matters is where the webroot lives.
 
This would mean on the filesystem the public HLS files are in `/home/user/owncast/public/hls` and the config API would say that the files are in `hls`, because it's relative to `PublicHLSPath`.

gingervitis:
@Ruxton sorry for the late reply here. I've recently merged in a ton of frontend changes.  A lot of files have also changed and or have been rearranged.
Can you apply your change requests to the updated code instead?

gabek:
We've since simplified and removed the option for custom paths for now.  In the future we hope to circle back around and properly make all these paths customizable.

Issue #140, opened 2020-09-05
Some mobile handling in the UI is wonky


gabek:
- [ ]  When you start typing and keyboard comes up the layout breakpoint changes and breaks the layout on small sizes.
- [x]  Emoji picker is hiding at the bottom of the screen.
- [ ]  On iPad auto-jumping to bottom of chat does not reach the very bottom, it's ~1 message off.

gabek:
> the layout breakpoint changes and breaks the layout on small sizes

I'm not sure how to reproduce this :\

gabek:
@gingervitis I made changes to hopefully address 2 of the 3 here.  I hid the emoji button for now and tied the `jumpToBottom` tighter into `render()` so there isn't the race condition of it jumping to bottom before the DOM update is complete.

gabek:
I originally thought the iPad scroll to bottom issue was related to a race condition with the dom not knowing the correct height, but it seems to be something completely unrelated to that, because as you scroll the left side the right side is sometimes fine and sometimes off.

![image](https://user-images.githubusercontent.com/414923/93525520-08641080-f8eb-11ea-9f2f-dd93025a6cf9.png)

![image](https://user-images.githubusercontent.com/414923/93525685-45300780-f8eb-11ea-8493-0faedb71558f.png)

PR #147, opened 2020-09-13
Misc web updates

gabek:
For #140.

### DIFF
--- a/webroot/js/components/chat/chat.js
+++ b/webroot/js/components/chat/chat.js
@@ -27,6 +27,8 @@ export default class Chat extends Component {
     this.receivedWebsocketMessage = this.receivedWebsocketMessage.bind(this);
     this.websocketDisconnected = this.websocketDisconnected.bind(this);
     this.submitChat = this.submitChat.bind(this);
+
+    this.jumpToBottomPending = false;
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


webroot/js/components/chat/chat.js
```diff
@@ -164,21 +166,39 @@ export default class Chat extends Component {
...
+    );
+
+    // After the render completes (based on requestAnimationFrame) then jump to bottom.
+    // This hopefully fixes the race conditions where jumpTobottom fires before the
+    // DOM element has re-drawn with its new size.
+    if (this.jumpToBottomPending) {
+      this.jumpToBottomPending = false;
+      window.requestAnimationFrame(
+        function () {
+          jumpToBottom(this.scrollableMessagesContainer.current);
```
gingervitis, line 187: we could do `this.jumpToBottom = this.jumpToBottom.bind(this);` in the constructor, for consistency.

webroot/styles/app.css
```diff
@@ -204,3 +204,14 @@ header {
 }
 
 /* ************************************************8 */
+
+
+@media screen and (max-width: 860px) {
```
gingervitis, line 209: you could move all the styles in this group to line 204, and do away with this line.

webroot/styles/chat.css
```diff
@@ -102,6 +102,13 @@
   padding: .25rem;
 }
 
+/* Hide emoji button on small screens */
+@media screen and (max-width: 860px) {
```
gingervitis, line 106: Would prefer if we did this in JS, but I think this should be good enough and easier for now.

PR #152, opened 2020-09-16
Reset the directories when the stream gets disconnected

graywolf336:
Whenever the stream goes offline, reset all the directories. Although, this might be a breaking change and we could potentially have this be configurable?

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path"
 	"strconv"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 
@@ -15,8 +16,9 @@ import (
 )
 
 var (
-	_stats   *models.Stats
-	_storage models.ChunkStorageProvider
+	_stats         *models.Stats
+	_storage       models.ChunkStorageProvider
+	_cleanupTicker *time.Ticker
 )
 
 //Start starts up the core processing
@@ -54,6 +56,25 @@ func createInitialOfflineState() error {
 	ffmpeg.ShowStreamOfflineState()
 
 	return nil
+}
+
+func startCleanupTimer() {
+	_cleanupTicker := time.NewTicker(5 * time.Minute)
+	go func() {
+		for {
+			select {
+			case <-_cleanupTicker.C:
+				resetDirectories()
+			}
+		}
+	}()
+}
+
+// StopCleanupTimer will stop the previous cleanup timer
+func stopCleanupTimer() {
+	if _cleanupTicker != nil {
+		_cleanupTicker.Stop()
+	}
 }
 
 func resetDirectories() {
--- a/core/status.go
+++ b/core/status.go
@@ -27,6 +27,8 @@ func GetStatus() models.Status {
 
 //SetStreamAsConnected sets the stream as connected
 func SetStreamAsConnected() {
+	stopCleanupTimer()
+
 	_stats.StreamConnected = true
 	_stats.LastConnectTime = utils.NullTime{time.Now(), true}
 	_stats.LastDisconnectTime = utils.NullTime{time.Now(), false}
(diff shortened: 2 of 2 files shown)


gabek:
In my (admittedly bloated) HLS refactor I have a change to cleanup after a timeout, so we can do this this immediate cleanup now and then move to that more nuanced change soon.

https://github.com/owncast/owncast/pull/151/files#diff-f75ca533d67f6226e2d6a028714c8c92R118

graywolf336:
Moving to a timeout approach is probably the best approach, as it gives the streamer time to reconnect and also doesn't remove the files prematurely. Is there any way that could be extracted out and applied now?

gabek:
Yeah I think that's what we should do, since thinking about it more if we delete files, and HLS is being served from local files, the clients likely will have not have finished reading them in yet and start getting 404s, missing the last seconds of the stream.  We can just copy the start/stop cleanup timer over and fire that instead of `resetDirectories` directly.   I can do that.

graywolf336:
I just tried this and it's not working 🤔  the folder is being removed on start up but the files are not being removed after a livestream has ended. Wonder if there are errors being thrown that we're not catching.

gabek:
Huh odd.  I just tested it (i'm using local serving of files in this case) and it seems to clean things up:

```
TRAC[2020-09-16T11:57:29-07:00] Resetting file directories to a clean slate.

gabek:video-stream/ (reset-directories-when-offline✗) $ ls hls/0 |wc -l
0
gabek:video-stream/ (reset-directories-when-offline✗) $ ls webroot/hls/0 |wc -l
0
```

gabek:
As an aside, without the delay it's possible the transcoder is still writing files after RTMP disconnect.

1. Disconnect
1. Cleanup
1. Write last couple files

Do you think that might be what you're seeing?

graywolf336:
And yours worked 👏   so going the timed way is the way to go.

graywolf336:
I'd approve it, but I can't since I opened it 😂

core/core.go
```diff
@@ -56,6 +58,25 @@ func createInitialOfflineState() error {
 	return nil
 }
 
+func startCleanupTimer() {
+	_cleanupTicker := time.NewTicker(5 * time.Minute)
+	go func() {
+		for {
+			select {
+			case <-_cleanupTicker.C:
+				resetDirectories()
```
graywolf336, line 67: @gabek something I just realized, this happens no matter whether the stream is still connected or not. 😬 unless I am missing something

gabek: I guess it doesn't help that we just tried to back port a change by copying/pasting from another branch.  If this can wait until 0.0.3 it's working over there.

In the 0.0.3 HLS rework branch here's the actual change:

https://github.com/owncast/owncast/commit/151316131af6526417cbeabaf13c0eec6a903c01#diff-f75ca533d67f6226e2d6a028714c8c92

gabek: But yeah, it only needs to run once, so it can be stopped after it runs after that first 5min delay.  But I'm not sure why you're seeing it continue to run even after a stream starts.  Seems like the ticker never is told to stop?

gabek: I've been testing with the HLS branch, and on goth.land every time I reconnect it's due to restarting the server, so I haven't seen this, but it's totally resetting after a second RTMP connection.

If you run with `--enableVerboseLogging` it'll log out when it runs that reset.

image

gabek: omg!

`_cleanupTicker := time.NewTicker(5 * time.Minute)`
should have been
`_cleanupTicker = time.NewTicker(5 * time.Minute)`

gabek: Hopefully this fixes it.  Testing more... https://github.com/owncast/owncast/commit/cb2794f68c029cd205addd9d5a97fe5b16c92d10

graywolf336: Oh my 🙈  I didn't even realize that! Whoopies! I'll test Sunday :)

PR #162, opened 2020-09-19
Preconnect preact, emoji. Defer some js.

gabek:
I have a hard time telling if this makes a difference or not.

* Defers some initial javascript loading to try to make room for everything else.
* Add preconnect upfront for preact and the emoji picker, hoping that it'll connect early on, and by the time the browser needs to pull those it'll be able to pull things in quicker.

`*shrug*`

### DIFF
--- a/webroot/index.html
+++ b/webroot/index.html
@@ -17,25 +17,34 @@
     <link rel="icon" type="image/png" sizes="96x96" href="/img/favicon/favicon-96x96.png">
     <link rel="icon" type="image/png" sizes="16x16" href="/img/favicon/favicon-16x16.png">
     <link rel="manifest" href="/manifest.json">
+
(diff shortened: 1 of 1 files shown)


mattdsteele:
Basic audit seems like it's a little better. Using Lighthouse with a Mobile setting shows First Contentful Paint goes from 4.7s on `master` vs 3.9s on this branch; and Time To Interactive from 8.7s to 7.5s.

Not giant gains but decent wins!

I tried playing with some component lazy-loading, but was running into problems getting `Suspense` working; not sure what the specific issue is, but I'll submit a separate PR if I get it working.

gabek:
Thanks for looking at it, Matt!  A little better is a little better and a win indeed.

PR #171, opened 2020-09-21
Remove player progress bar.  Closes #159

### LINKED ISSUE #159: People find the video player progress bar distracting
If the player doesn't see you at what it determines is "live" then you get the progress bar jumping around as new segments get downloaded.

This video was sent me from the test stream last night saying it was distracting:
https://www.dropbox.com/s/js4j2b3tittgjhn/jumping.mp4?dl=0

I don't really see a point of showing this progress bar for live.  I know we discussed at one point giving the option to rewind, but in practice it's very uncommon and in many cases, not even possible (due to the the HLS playlist keeping references to very few segments).

If we can't just turn it off via a video.js option, then I think we should hide it via CSS.

gabek:
Looks a lot cleaner, actually.

image

### DIFF
--- a/webroot/js/components/player.js
+++ b/webroot/js/components/player.js
@@ -11,13 +11,18 @@ const VIDEO_SRC = {
 };
 const VIDEO_OPTIONS = {
   autoplay: false,
-  liveui: true, // try this
+  liveui: true,
   preload: 'auto',
+  controlBar: {
+    progressControl: {
+      seekBar: false,
+    },
+  },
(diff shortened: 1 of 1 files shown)


gingervitis:
The lack of a progress bar feels a bit weird to me. I think it's still nice to know how much time has passed. If you pause, it seems hard to tell that it will jump to live when you resume.
I do know the way it worked before was also a bit weird. But I guess we'll see how this goes; see what people say..

PR #184, opened 2020-09-25
Update caching logic/headers

gabek:
This helps with some over-cached assets.

This change:

* Adds ETags for static assets.
* Supports customized max age for different types of assets that we can tweak easily.

With this configuration in theory something like Javascript would have a pretty low expiration time, but then next time it loads it would compare ETags and not have to re-download the unchanged file.

But something that would never change, such as a video segment (`.ts`) has a really long expiration.

And the thumbnail is special-cased to expire at the same rate it's generated. 

The only downside of this that I think is worth keeping in mind is the ETag is being generated based on the `last modified` time of the requested file on the filesystem.  While getting the stats of a file on a filesystem should be fast, there may be a limit.


geekgonecrazy (approved):
This is pretty straight forward.  Looks good to me

PR #188, opened 2020-09-25
Create release docker image without shipping source and go toolchain.

gabek:
This changes the Docker image build process.  The build takes place in one step, and then the result gets copied into the distribution image.

Now we don't have to ship the source code that nobody would use, and the Go toolchain doesn't need to be shipped just for the initial build.

@geekgonecrazy let me know if you have any thoughts on this change.

### DIFF
--- a/scripts/Dockerfile-build
+++ b/scripts/Dockerfile-build
@@ -18,6 +18,11 @@ RUN CGO_ENABLED=1 GOOS=linux go build -a -installsuffix cgo -ldflags "-extldflag
 # Create the image by copying the result of the build into a new alpine image
 FROM alpine
 RUN apk add --no-cache ffmpeg ffmpeg-libs
+
+# Certs
+RUN apk update && apk add ca-certificates && update-ca-certificates
+
+# Copy owncast assets
 WORKDIR /app
 COPY --from=build /build/owncast /app/owncast
 COPY --from=build /build/config.yaml /app/config.yaml
--- a/scripts/Dockerfile-build
+++ b/scripts/Dockerfile-build
@@ -17,7 +17,7 @@ RUN CGO_ENABLED=1 GOOS=linux go build -a -installsuffix cgo -ldflags "-extldflag
 
 # Create the image by copying the result of the build into a new alpine image
 FROM alpine
-RUN apk update && apk add ca-certificates && update-ca-certificates && apk add --no-cache ffmpeg ffmpeg-libs
+RUN apk update && apk add --no-cache ffmpeg ffmpeg-libs ca-certificates && update-ca-certificates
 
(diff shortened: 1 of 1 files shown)


geekgonecrazy:
this is how I typically do my docker images. :grin: will owncast ever need to make any http/https calls out?  if so you might want to include `ca-certificates` in the final image

gabek:
Good idea!

gabek:
I'm getting this error when building the image:

```
WARNING: ca-certificates.crt does not contain exactly one certificate or CRL: skipping
```

Is that something you've seen before?

scripts/Dockerfile-build
```diff
@@ -15,5 +15,18 @@ ENV NAME=${NAME}
 
 RUN CGO_ENABLED=1 GOOS=linux go build -a -installsuffix cgo -ldflags "-extldflags \"-static\" -s -w -X main.GitCommit=$GIT_COMMIT -X main.BuildVersion=$VERSION -X main.BuildType=$NAME" -o owncast .
 
+# Create the image by copying the result of the build into a new alpine image
+FROM alpine
+RUN apk add --no-cache ffmpeg ffmpeg-libs
+
+# Certs
+RUN apk update && apk add ca-certificates && update-ca-certificates
```
geekgonecrazy, line 23: I’ve never needed to run the update.  Also I’d combine this install into previous run so you keep the number of layers small

gabek: Let me know if this is along the lines you were thinking of:

https://github.com/owncast/owncast/pull/188/commits/e41104c0c8d78c35f0d0fcb2e9ec110c629da07c#diff-26db72de85beb4285d47e4cc9f631449R20

scripts/Dockerfile-build
```diff
@@ -15,5 +15,15 @@ ENV NAME=${NAME}
 
 RUN CGO_ENABLED=1 GOOS=linux go build -a -installsuffix cgo -ldflags "-extldflags \"-static\" -s -w -X main.GitCommit=$GIT_COMMIT -X main.BuildVersion=$VERSION -X main.BuildType=$NAME" -o owncast .
 
+# Create the image by copying the result of the build into a new alpine image
+FROM alpine
+RUN apk update && apk add ca-certificates && update-ca-certificates && apk add --no-cache ffmpeg ffmpeg-libs
```
geekgonecrazy, line 20:
```suggestion
RUN apk update && apk add --no-cache ffmpeg ffmpeg-libs ca-certificates && update-ca-certificates
```
Simplifies a bit more.  From my experience I haven’t needed `update-ca-certificates` but I don’t think it hurts any

gabek: Thanks!

PR #204, opened 2020-09-28
Disconnect stream Admin API + HTTP Basic Auth

gabek:
This adds:

* Middleware for HTTP Basic Auth to authenticate Admin requests.
* Simple API to end the inbound RTMP connection to end the current live stream.  `/api/admin/disconnect`

### DIFF
--- a/controllers/admin.go
+++ b/controllers/admin.go
@@ -1,0 +1,13 @@
+package controllers
+
+import (
+	"net/http"
+
+	"github.com/gabek/owncast/core/rtmp"
+)
+
+// DisconnectInboundConnection will force-disconnect an inbound stream
+func DisconnectInboundConnection(w http.ResponseWriter, r *http.Request) {
+	rtmp.Disconnect()
+	w.WriteHeader(http.StatusOK)
+}
--- a/controllers/admin.go
+++ b/controllers/admin.go
@@ -3,11 +3,17 @@ package controllers
 import (
 	"net/http"
 
+	"github.com/gabek/owncast/core"
 	"github.com/gabek/owncast/core/rtmp"
 )
 
 // DisconnectInboundConnection will force-disconnect an inbound stream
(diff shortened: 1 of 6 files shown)


controllers/admin.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"net/http"
+
+	"github.com/gabek/owncast/core/rtmp"
+)
+
+// DisconnectInboundConnection will force-disconnect an inbound stream
+func DisconnectInboundConnection(w http.ResponseWriter, r *http.Request) {
+	rtmp.Disconnect()
+	w.WriteHeader(http.StatusOK)
```
graywolf336, line 12: Wonder if we should go ahead and write a body back, such as `{ success: true }` so that clients at least get a response back and specifically for the future web client. I'm just thinking we would do something like

```js
const result = await fetch();
await result.json();
```

Or something to that 🤔

graywolf336:
Does this work and actually kick a stream off? For example, does OBS try to reconnect and restart it?

controllers/admin.go
gabek: Yeah there's no reason not to return some standardized success.  I'll add that.

gabek:
> Does this work and actually kick a stream off? For example, does OBS try to reconnect and restart it?

I tested it with OBS and ffmpeg as broadcasting clients, and it does indeed kick it off.   OBS reconnects:

```
INFO[2020-09-28T18:21:25-07:00] Incoming RTMP connected.
INFO[2020-09-28T18:23:07-07:00] Inbound stream disconnect requested.
INFO[2020-09-28T18:23:07-07:00] RTMP disconnected.
INFO[2020-09-28T18:23:13-07:00] Incoming RTMP connected.
```

We still need to figure out an automated stream key rotation, but that, along with this, is a good start to the broadcasting demo server.  Plus we'll want this disconnect functionality in the admin anyway.

PR #209, opened 2020-09-30
First pass at YP registration/configuration

gabek:
Because we're starting work on the admin, I'd like to be able to display the status of YP integration in the dashboard.  So this is the first pass at the "Ping Handshake".  If enabled it'll operate silently unless verbose logging is enabled, and it's disabled by default.

The new optional config file block is the following:

```
yp:
  enabled: true
  instanceURL: http://watch.owncast.online
``` 
The two properties are for enabling, and what the publicly accessible URL to the instance is.  That's what the YP service will use in the handshake and for the link to this instance.

A `ypServiceURL` property is also available to point at a different YP service (like localhost) for development purposes.

1. Owncast instance makes a `ping` request to the YP API service with the instance's URL and key.
1. If instance has previously been registered with the service and the key doesn't match an error is thrown.
1. YP service makes a request back to the instance's `/api/yp` API to get the details of the instance (name, logo, viewer count, etc).
1. New key is returned.
1. Cycle repeats every 4 minutes. If the instance doesn't re-ping in 5minutes then it's seen as offline.

### Owncast updates to support this
These are all quite small and doesn't touch much outside of its own package.  But it does add a new route, and keeps an instance of `yp` in `core` to tell to start when a stream begins, and to stop when it ends.

* Adds optional NSFW config property to `InstanceDetails`.
* Changes `Logo` from a map of strings to an actual typed `Large` and `Small` version.
* Enhances the Thumbnail generation to also add a Preview animated gif if YP is enabled.
* To keep things together both the outbound request `yp.go` and the inbound API endpoint `api.go` are a part of a standalone `yp` package.
* The rotating key is written to `.yp.key`.


mattdsteele:
I've not looked at the code yet but as an RFC, I like everything proposed.

If we wanted to be slightly more "standardized" with the message validation we could send the messages as a [JWT](https://jwt.io/) with an [HMAC](https://en.wikipedia.org/wiki/HMAC) signature. I don't actually know much about how this is implemented beyond the description and Go library described by @jmhobbs [in his blog post](https://www.velvetcache.org/2020/01/29/netlify-cloudflare-crazy-delicious).

That said, I don't know there's a huge benefit to using JWT here (beyond verifying a message payload, which doesn't seem like a big concern for us), and I would definitely stumble trying to implement it, so maybe this is a v2 concern :)

config-example.yaml
```diff
@@ -25,3 +25,9 @@ instanceDetails:
 videoSettings:
   # Change this value and keep it secure.  Treat it like a password to your live stream.
   streamingKey: abc123
+
+  # Enable YP to be listed in the Owncast directory and let people discover your instance.
+yp:
+  enabled: true
```
mattdsteele, line 31: Given my workflow of starting a new instance with `cp config-example.yaml config.yaml`, this effectively defaults to true. That's probably OK, we should just be mindful of it.

gabek: Good point!  Changed.

core/ffmpeg/thumbnailGenerator.go
```diff
@@ -83,12 +84,32 @@ func fireThumbnailGenerator(chunkPath string, variantIndex int) error {
...
 
+	// If YP support is enabled also create an animated GIF preview
+	if config.Config.YP.Enabled {
+		makeAnimatedGifPreview(mostRecentFile, previewGifFile)
+	}
+
 	return nil
 }
+
+func makeAnimatedGifPreview(sourceFile string, outputFile string) {
```
mattdsteele, line 99: This is neat! Had no idea you could do this with ffmpeg!

yp/yp.go
```diff
@@ -0,0 +1,139 @@
...
+		return ""
+	}
+
+	return string(fileBytes)
+}
+
+// DisplayInstructions will let the user know they are not in the directory by default and
+// how they can enable the feature.
+func DisplayInstructions() {
+	text := "Your instance can be listed on the Owncast directory at http://something.something by enabling YP in your config.  Learn more at http://something.something."
```
mattdsteele, line 137: Should these URLs be https://yp.owncast.online? Or are you thinking the actual web-based directory might be something different?

gabek: Yeah, as it stands I've been building the frontend at https://directory.owncast.online and the API at https://yp.owncast.online since they're built and deployed as separate projects.  But this might change.  And I figured we'd point them at some documentation on our docs site discussing how to configure it and how it works, but I'm not sure what the URL of that doc will be.

Realistically I'll probably completely pull this message, since it's only if you're logging verbosely, and it makes sense just to put this stuff in the admin once we have it.

gabek:
> I've not looked at the code yet but as an RFC, I like everything proposed.
> 
> If we wanted to be slightly more "standardized" with the message validation we could send the messages as a [JWT](https://jwt.io/) with an [HMAC](https://en.wikipedia.org/wiki/HMAC) signature. I don't actually know much about how this is implemented beyond the description and Go library described by @jmhobbs [in his blog post](https://www.velvetcache.org/2020/01/29/netlify-cloudflare-crazy-delicious).
> 
> That said, I don't know there's a huge benefit to using JWT here (beyond verifying a message payload, which doesn't seem like a big concern for us), and I would definitely stumble trying to implement it, so maybe this is a v2 concern :)

I do think we might want to tighten things up by validating payloads and make sure that the server that made the `ping` request is the same server that is responding to the `/yp` API call.  After reading John's post maybe JWT is a nice solution for that, but it might also be overkill.

I'm going to merge this in so we can develop against it, but I'm going to do some experimenting and possibly add another layer of verification in order to plug any holes that might exist in this.

PR #213, opened 2020-10-01
Add timestamp to title chat

### LINKED ISSUE #177: Add timestamp of message as title attribute to chat messages
While I don't want to see the timestamp directly in the UI, sometimes I'm curious how long ago somebody sent a message.  Hovering over this message to get a timestamp string would be a compromise and stay out of the way.

ahmadkarlam:
Fix #177

### DIFF
--- a/webroot/index.html
+++ b/webroot/index.html
@@ -35,9 +35,6 @@
     <script src="//unpkg.com/showdown/dist/showdown.min.js" defer></script>
     <script type="module" src="https://cdn.jsdelivr.net/npm/@justinribeiro/lite-youtube@0.6.2/lite-youtube.js" defer></script>
 
-    <!-- moment js -->
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.29.0/moment.min.js" integrity="sha512-Izh34nqeeR7/nwthfeE0SI3c8uhFSnqxV0sI9TvTcXiFJkMd6fB644O64BRq2P/LA/+7eRvCw4GmLsXksyTHBg==" crossorigin="anonymous"></script>
-
     <link href="./styles/video.css" rel="stylesheet" />
     <link href="./styles/chat.css" rel="stylesheet" />
     <link href="./styles/user-content.css" rel="stylesheet" />
--- a/webroot/index.html
+++ b/webroot/index.html
@@ -35,6 +35,8 @@
     <script src="//unpkg.com/showdown/dist/showdown.min.js" defer></script>
     <script type="module" src="https://cdn.jsdelivr.net/npm/@justinribeiro/lite-youtube@0.6.2/lite-youtube.js" defer></script>
 
+    <!-- moment js -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.29.0/moment.min.js" integrity="sha512-Izh34nqeeR7/nwthfeE0SI3c8uhFSnqxV0sI9TvTcXiFJkMd6fB644O64BRq2P/LA/+7eRvCw4GmLsXksyTHBg==" crossorigin="anonymous"></script>
 
     <link href="./styles/video.css" rel="stylesheet" />
     <link href="./styles/chat.css" rel="stylesheet" />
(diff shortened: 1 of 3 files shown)


gingervitis (changes requested):
Could you make this timestamp title a bit more human-readable? Even as admins looking at it, a title like "2020-10-01T20:02:46.36728084Z" isn't particularly useful or easy to read.

Perhaps you can prefix it with "Sent at.. [readable time string]".

ahmadkarlam:
> Could you make this timestamp title a bit more human-readable? Even as admins looking at it, a title like "2020-10-01T20:02:46.36728084Z" isn't particularly useful or easy to read.
> 
> Perhaps you can prefix it with "Sent at.. [readable time string]".

how about a format like this `Sat Oct 03 2020 07:20:59 GMT+0700 (Western Indonesia Time)` ?

gingervitis:
Let's do this:
a. Just display the time like HH:MM:SS (`.toLocaleTimeString()`)
b. However, if the message happened more than 1 day ago, display, print it out like `Oct 2 at HH:MM:SS`

Hope that helps!

ahmadkarlam:
@gingervitis I added moment.js for diff and date format. I hope it's okay.

webroot/index.html
```diff
@@ -35,6 +35,8 @@
     <script src="//unpkg.com/showdown/dist/showdown.min.js" defer></script>
     <script type="module" src="https://cdn.jsdelivr.net/npm/@justinribeiro/lite-youtube@0.6.2/lite-youtube.js" defer></script>
 
+    <!-- moment js -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.29.0/moment.min.js" integrity="sha512-Izh34nqeeR7/nwthfeE0SI3c8uhFSnqxV0sI9TvTcXiFJkMd6fB644O64BRq2P/LA/+7eRvCw4GmLsXksyTHBg==" crossorigin="anonymous"></script>
```
graywolf336, line 39: If we are going to use any library for this, let's please go with a lightweight one and not the behemoth library that is momentjs. An alternative would be something like https://github.com/iamkun/dayjs

graywolf336 (changes requested):

webroot/index.html
gingervitis: Oh nice find! @graywolf336

Though for this particular task it doesn't seem necessary to pull in another external dependency, since we're not doing another date math anywhere else in the app. But you can totally borrow code from it to see how they do date calculations and put it in a util file for the messages @ahmadkarlam!

gingervitis: @ahmadkarlam have a look at the Javascript Date API, there are some pretty good functions in there you can use to do some calculations in probably the same amount of code you'd need with an external library.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

You can some calculations with just new Date() which gets now, and new Date(messageTimestamp), and do some math on the getHour() of their results.

Also, no need to worry about converting timezones or anything like that. Just using your computer's local date for this is good enough.  `toLocaleString()` and `toLocaleTimeString()` seem like a pretty good options.

webroot/js/components/chat/message.js
```diff
@@ -35,6 +35,7 @@ export default class Message extends Component {
               </div>
               <div
                 class="message-text text-gray-300 font-normal overflow-y-hidden"
+                title=${`Sent at ${formatTimestamp(timestamp)}`}
```
gingervitis, line 38: Looks great @ahmadkarlam ! Could you declare this string up around line 19 with the other declaration, just for code style consistency?

webroot/js/utils/chat.js
```diff
@@ -278,3 +278,15 @@ export function convertOnPaste( event = { preventDefault() {} }) {
     document.execCommand('insertText', false, value);
   }
 }
+
+export function formatTimestamp(sentAt) {
+  sentAt = new Date(sentAt);
```
gingervitis, line 283: In case `sentAt` is falsey, let's return an empty string '' to start off with;

ahmadkarlam: Done

webroot/js/components/chat/message.js
ahmadkarlam: Done

gingervitis:
This shall do. Thank you for your time on this @ahmadkarlam !

PR #214, opened 2020-10-01
Make video segment filename unique

### LINKED ISSUE #187: Make video segment filename unique
Right now if you were to run multiple copies of Owncast and point to the same external storage bucket there's a non-zero chance of having the same file names.

We should append some random identifier into `hls_segment_filename` in the Transcoder as a simple way to create unique segment names.

The solution right now would be to have different buckets for different instances, but it would be nice to not have to do that.

anoopmsivadas:
Adds a unique segment identifier to make the segment file names unique

Fixes #187

### DIFF
--- a/core/ffmpeg/transcoder.go
+++ b/core/ffmpeg/transcoder.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/teris-io/shortid"
 
 	"github.com/gabek/owncast/config"
 	"github.com/gabek/owncast/utils"
@@ -25,6 +26,7 @@ type Transcoder struct {
 	segmentLengthSeconds int
 	appendToStream       bool
 	ffmpegPath           string
+	segmentIdentifier    string
 }
 
 // HLSVariant is a combination of settings that results in a single HLS stream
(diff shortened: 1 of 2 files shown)


core/ffmpeg/transcoder.go
```diff
@@ -125,7 +131,7 @@ func (t *Transcoder) getString() string {
 		// Filenames
 		"-master_pl_name", "stream.m3u8",
 		"-strftime 1",                                                               // Support the use of strftime in filenames
-		"-hls_segment_filename", path.Join(t.segmentOutputPath, "/%v/stream-%s.ts"), // Each segment's filename
+		"-hls_segment_filename", path.Join(t.segmentOutputPath, "/%v/stream-%s-"+string(t.segmentIdentifier)+".ts"), // Each segment's filename
```
graywolf336, line 134: Why is `string(t.segmentIdentifier)` required if `t.segmentIdentifier` is already a string?

anoopmsivadas: oops! initially used it with a direct function call. will remove that.

gabek:
Thanks @anoopmsivadas, just tested it out and it looks good!  Thanks for making that string update.

PR #217, opened 2020-10-02
Connected clients admin API

gabek:
* Up until now Client IDs have been the user's IP address and their web browser's user-agent concatenated.  Now it's still that, but a MD5 hash of it to make it a little less unwieldy when displaying them.
* We consider a new client being added when **either** 1) Somebody connects to the websocket or 2) Starts downloading video.  Some people will be doing both, some people will only be watching video (not connected to the websocket and instead using an external player such as a web embed, VLC, QuickTime, etc).  If they do both they will have the same client id and they won't show up as a duplicate.
* This is always how it has worked (using chat and video to count viewers), but instead of just the client ID strings it's actual client objects now.
* Websocket (chat) clients are different than this new, more generic, client type we use for storing basic client details.  The websocket deals with the actual socket, callbacks, etc.  But the websocket client does know things like username and message count.  So when a list of clients is asked for in `GetClients()` we ask the chat package about client ids and if it knows about a websocket client with that id it'll return a client object with the additionally populated data of username and message count.  This is done with `GetViewerClientFromChatClient()`, a new method that takes a websocket client and returns a generic viewer client from it.

### Geo IP implementation 
I added support for GeoIP using https://freegeoip.app.  I think this is a cool addition from the admin side, to be able to see (approximately) where your viewers are tuning in from, but obviously it relies on a 3rd party API and that's not great.  I'm going to keep looking to see if I can find a GeoIP database we can use ourselves that is free, and allowed to be re-distributed.

* Once an IP address is known it tells the GeoIP package to fetch the geo data.  This is an asynchronous "fire and forget" operation and the result is not returned.
* When actual geo data is needed it asks the GeoIP package for data, and if it has previously fetched it, it'll return it.
* This is designed this way so fetching geo from the external API is considered a background action that does not block or slow anything down.

image


gabek:
Was also thinking of the possibility of people plugging in their own geoip service since most of them work the same, but the response's keys are different from one to the other, but the data is the same.  So I'm not sure how we'd match up the keys without having an overly complex mapping somewhere.

gabek:
@graywolf336 @gingervitis @geekgonecrazy @mattdsteele @jeyemwey 

I think it'd be really nice to have this broad viewer geo data in the admin, but I'm not sure it's worth using an external API.  Do any of you have an idea to find a middle ground?  Otherwise I think we'll just have to kill this feature, but I do think it'd be a nice thing to have in the admin, since admittedly we can't offer much about viewers right now, especially compared to other services that are able to give very detailed demographic information (not that we'd want that much detail).

mattdsteele:
What are your concerns re: using an external API?

Generally I get worried when integrations are a potential point of failure, but it looks like the way you've built this out behind goroutines and can always fall back to just showing the raw IP, it seems unlikely to cause any issues.

I'll check out the code, but conceptually I'm 👍 for merging.

geoip/geoip.go
```diff
@@ -0,0 +1,72 @@
...
+
+		var apiResponse geoRequestResponse
+		err := json.Unmarshal(body, &apiResponse)
+		if err != nil {
+			log.Errorln(err)
+			return
+		}
+
+		response := GeoDetails(apiResponse)
+		geoIPCache[ip] = response
```
mattdsteele, line 69: I don't write enough Go to have a good feel for if it's idiomatic, but I tend to have all the results of my goroutines run through a channel, rather than updating a map directly like this. Is there a heuristic you use for when channels are useful?

gabek: I don't write enough Go to know either, but I'm trying to resolve that! :). But it's a good question, and I don't have a good answer.  I guess I'm thinking that I'm updating something still in scope, but it probably is more correct to always return everything back via channels.

gabek:
> What are your concerns re: using an external API?
> 
> Generally I get worried when integrations are a potential point of failure, but it looks like the way you've built this out behind goroutines and can always fall back to just showing the raw IP, it seems unlikely to cause any issues.
> 
> I'll check out the code, but conceptually I'm 👍 for merging.

I'm only thinking about it more than usual because once a release goes out then it's out of our hands.  I usually wouldn't be concerned if this was a hosted web product or something that we can update ourselves if the API were to go offline, or start doing something funny.  But you're right, unless the API starts returning completely incorrect values then it wouldn't impact much.  Even then it wouldn't break anything except give bad output.

mattdsteele:
Ostensibly we could host a facade service (hosted at geoip.owncast.online or the like) and point clients toward it, rather than an external service directly? That would let us swap out implementations, or just shut it off if it got too bothersome.

I'm not sure I've thought through all the privacy implications of us taking this on, though.

geoip/geoip.go
mattdsteele: Yeah maybe the determination is as simple as if it's in the same file? I don't think anything needs changed here.

gabek:
> Ostensibly we could host a facade service (hosted at geoip.owncast.online or the like) and point clients toward it

I was thinking the same thing!  But figured we probably don't want to get in the business of API hosting haha.

gabek:
I have a solution!

Our GeoIP util can have [native support](https://github.com/oschwald/geoip2-golang) for the free [MaxMind GeoLite2](https://dev.maxmind.com/geoip/geoip2/geolite2/) database (that's what it seems most of these free geoip API services are using anyway), and we allow operators of Owncast instances to drop a copy of the database in there if they want it.  We can [link directly](https://www.maxmind.com/en/geolite2/signup) to where they can get a copy.  Then we're not bundling a proprietary database  (and wasting space doing so), and people have choice.

In the future this would be another good example of a plugin.

gabek:
Made an update that uses the GeoLite2 database if it's in `/data`.  I threw a copy of the database for testing in the Rocket.Chat room if anybody wants to test with it.  I'll write up some documentation on how to enable it.  We can also surface this in the admin to show that it's enabled/disabled.

gabek:
Documentation added: https://owncast.online/docs/geoip/

gabek:
Merging since I'm happy with this resolution.  But if anybody else has any thoughts, let me know!

PR #219, opened 2020-10-02
Fix: Optional socialHandles and tags

### LINKED ISSUE #202: Make socialHandles not required when rendering page
Getting the following error in 0.0.2 (in docker container, running on K3s)
```
Uncaught (in promise) TypeError: socialHandles is null
    render https://stream.spamasaurus.com/js/app.js:403
    Preact 3
```
Page rendering does not continue/complete properly.

My config.yaml has all socialHandles-related settings commented:
```
# See documentation for full list of supported social links.  All optional.
#  socialHandles:
#    - platform: twitter
#      url: http://twitter.com/owncast
#    - platform: instagram
#      url: http://instagram.biz/owncast
#    - platform: facebook
#      url: http://facebook.gov/owncast
```

As soon as I uncomment at least one entry, the page renders correctly.

jeyemwey:
Fixes #202 and makes `socialHandles` optional. Since the code looked similiar for the `tags` list above, I have changed the behaviour, too.

The code looks a little bulky since I had to move the `<ul>` with the "Follow me" label up into function, so it wouldn't show if there are no socialHandles.


webroot/js/app.js
```diff
@@ -398,17 +397,33 @@ export default class App extends Component {
...
-          (item, index) => html`
-            <li key="social${index}">
-              <${SocialIcon} platform=${item.platform} url=${item.url} />
-            </li>
-          `
-        );
+        )
+      : null;
+
+    const socialIconsList = (function () {
```
gingervitis, line 403: Since it's getting clunky, why don't we extract this into its own component like a <SocialList> (or a socialist? jk), that we can pass the `socialHandles` into, and let the component do the rest of the work.

webroot/js/app.js
```diff
@@ -387,9 +387,8 @@ export default class App extends Component {
     const bgLogo = { backgroundImage: `url(${smallLogo})` };
     const bgLogoLarge = { backgroundImage: `url(${largeLogo})` };
 
-    const tagList = !tags.length
-      ? null
-      : tags.map(
+    const tagList = (tags !== null && tags.length > 0)
```
gingervitis, line 390: Now that I think about it, we could probably just do a check for `(!tags) ?..` and not worry about `tags.length` as the mapping will just stop if length=0.

webroot/js/app.js
jeyemwey: I did not feel like inventing new components for this, but now that you say it, yess.

webroot/js/app.js
jeyemwey: With this, it's explicit 🗡️

graywolf336: While it is explicit, it will simplify the logic and make it easier to read. 🤔

webroot/js/components/socialIconsList.js
```diff
@@ -40,3 +41,24 @@ export default function SocialIcon(props) {
       </a>
   `);
 }
+
+export default function (props) {
+  const { handles } = props;
+  console.log(props);
```
gabek, line 47: Debugging log got left in.

webroot/js/app.js
jeyemwey: I have moved the icons list to its own component now.

webroot/js/components/socialIconsList.js
jeyemwey: Oops!

gabek:
Doh sorry for the merge conflict.

webroot/js/app.js
gingervitis: Sorry, just saw this update! 
Can I be picky and request that the file name be in kebab-case, instead of camelCase, to be consistent with the other files in /webroot? 
Otherwise looks good! :shipit:

webroot/js/app.js
jeyemwey: Did as you pleased, _now_ it *should* be done.

gingervitis:
sweet, thank you!

Issue #164, opened 2020-09-19
Serving web assets as gzip'ed


gabek:
Not sure if the tradeoff of supporting compressed assets and using the CPU to do so is worth the gain of squeezing the files down smaller.  Also it doesn't look like there's any out of the box support for this in Go's `http`, but we could use some middleware.

gabek:
Since in general we're CPU bound, adding another task to the CPU is probably not a great idea.  Closing.

PR #220, opened 2020-10-04
Move all remote dependencies to be local

gabek:
This change moves all of our remote dependencies to be local modules.  Hopefully there's no fallout from this.

@gingervitis @mattdsteele Let me know what you think!

## How to add a new Javascript/npm dependency with this setup

1. Go to `build/javascript`
1. Add your thing `npm install bloat.js`
1. Run `npm run build`

It'll then:

1. Download the javascript dependency
2. Use snowpack to do its modular magic
3. Copy the local files to `webroot/js/web_modules`

Note: These are not bundled, they're still standalone files.  If we want to bundle them for an additional performance boost we can address that with Snowpack as well.


gabek:
Apparently this change to make things better makes things way worse.  So **shrug**.

Previously:

image

Now:

image

mattdsteele:
Ah weird! I'll take a look at what might be going on.

mattdsteele:
I think the big driver here is just that we aren't serving gzipped assets, and some of the files (especially tailwind.min.css) is pretty massive.

Here's what it looks like if I put Caddy in front and enable gzip:

image

By comparison, here's what's on `master`, which is pretty similar:

image

So #164 might be more important if we end up serving more files locally; or just expect folks to use a reverse proxy that does these optimizations.

mattdsteele:
It's probably also worth looking at optimizing Tailwind specifically; even minified it's still hundreds of KB of render-blocking CSS.

Once this lands and we have a local JS toolchain in place, it should be possible to create a custom Tailwind build which only contains the styles we use. https://tailwindcss.com/docs/controlling-file-size

webroot/index-standalone-chat.html
```diff
@@ -6,17 +6,16 @@
     <link href="https://unpkg.com/tailwindcss@^1.0/dist/tailwind.min.css" rel="stylesheet" />
```
gabek, line 6: Missed changing to local copy of tailwind.

gabek:
Thanks for looking at this Matt!  That all makes sense, and I feel much better now!  We can certainly look at fixing that bulky CSS.  I've wondered if we should be gzipping natively within our own web server, but ultimately decided against it as to not throw more CPU load at the whole system.  But maybe it's worth looking at.

PR #221, opened 2020-10-04
Fixes #211: Update mod paths

### LINKED ISSUE #211: Update import and mod path
https://github.com/owncast/owncast/blob/6a0e8deae3b9a9b550a8bb4c0d193bdb6e126b96/go.mod#L1

The import path and go mod path still points to the old repo.  I think github probably handles redirect when it tries to fetch.  But might be good to update it to prevent any problems

leowmjw:


go.sum
```diff
@@ -1,26 +1,24 @@
-github.com/PuerkitoBio/goquery v1.5.1/go.mod h1:GsLWisAFVj4WgDibEWF4pvYnkVQBpKBKeU+7zCJoLcc=
 github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d h1:G0m3OIz70MZUWq3EgK3CesDbo8upS2Vm9/P3FtgI+Jk=
 github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
 github.com/amalfra/etag v0.0.0-20190921100247-cafc8de96bc5 h1:+ty4KYpIDUhjzsVsV+HiTeYEfufBc/4FLNiqIGU1A1U=
 github.com/amalfra/etag v0.0.0-20190921100247-cafc8de96bc5/go.mod h1:Qk51jPgvIaO549MR+IvLP/uMZbZGs05QJSzEhDVZ1jc=
-github.com/andybalholm/cascadia v1.1.0/go.mod h1:GsXiBklL0woXo1j/WYWtSYYC4ouU9PqHO0sqidkEA4Y=
-github.com/aws/aws-sdk-go v1.34.0 h1:brux2dRrlwCF5JhTL7MUT3WUwo9zfDHZZp3+g3Mvlmo=
-github.com/aws/aws-sdk-go v1.34.0/go.mod h1:5zCpMtNQVjRREroY7sYe8lOMRSxkhG6MZveU8YkpAk0=
+github.com/aws/aws-sdk-go v1.35.2 h1:qK+noh6b9KW+5CP1NmmWsQCUbnzucSGrjHEs69MEl6A=
```
gabek, line 5: I see this change is bumping the version of a handful of the libraries we're using.  While this is probably harmless, just so we don't have to worry about the ramifications of this, and to keep changes in scope to the package name only, would you mind reverting these unrelated changes?

geekgonecrazy: Agreed all looks good except for the bumping of versions.  Such a change typically want to do explicitly in another PR just because of nature of upgrading dependencies

leowmjw:
@geekgonecrazy @gabek Pls see if my revert is to your request.  Thanks!

geekgonecrazy (approved):
LGTM

gabek:
Seems to be no side effects from this change, thanks @leowmjw!

PR #226, opened 2020-10-05
fix: prevents messages ending with spaces to exceed the msg limit

Pranav2612000:
This PR removes the trim() so that the spaces at the end of the message are counted while calculating message length. It also replaces .innerText by textContent, as innerText has some unexpected behaviour with backspace.


gingervitis:
Cool, this works for me. Thank you for the fix and the `textContent` tip!

PR #228, opened 2020-10-06
ADD: Save volume settings in localstorage

### LINKED ISSUE #175: Save volume settings
We should cookie or localStorage the last video volume setting that the user had.
It's pretty jarring to reload the page and have the video play at full volume every time.

SoldierCorp:
Changes
- Volume is being saved in localstorage and set on reload

closes #175


webroot/js/components/player.js
```diff
@@ -95,6 +100,10 @@ class OwncastPlayer {
     }
   }
 
+  handleVolume(e) {
+    localStorage.setItem('owncastVolume', this.vjsPlayer.volume());
```
gabek, line 104: There should be a `getLocalStorage`/`setLocalStorage` in `helpers.js`, could we use those instead so these getters and setters are centralized?

SoldierCorp: Working on it!

SoldierCorp: Done!

webroot/js/components/player.js
```diff
@@ -79,6 +82,9 @@ class OwncastPlayer {
   startPlayer() {
     this.log('Start playing');
     const source = { ...VIDEO_SRC };
+
+    if (getLocalStorage(PLAYER_VOLUME) !== null)
```
gingervitis, line 86: Hey @SoldierCorp , Thanks for doing this! Looks great. 
Just a small nit - can we put `{}` around this condition, just to keep the code style consistent across the app? :)

thanks!

SoldierCorp: I added a cleaner version

webroot/js/components/player.js
```diff
@@ -83,8 +83,7 @@ class OwncastPlayer {
     this.log('Start playing');
     const source = { ...VIDEO_SRC };
 
-    if (getLocalStorage(PLAYER_VOLUME) !== null)
-      this.vjsPlayer.volume(getLocalStorage(PLAYER_VOLUME));
+    this.vjsPlayer.volume(getLocalStorage(PLAYER_VOLUME) || 1);
```
gingervitis, line 86: I think if the localStorage value = 0/mute, this will also end up being 1.

SoldierCorp: I tested the functionality and it's working as expected.

Peek 2020-10-07 00-09

gingervitis: TIL localStorage keeps everything as strings, such as "0", which is truthy, so that's why this works. I guess this will do.  Thanks @SoldierCorp!

PR #231, opened 2020-10-08
Remove additional viewer counts from status bar.

gabek:
These values will be displayed in the admin, and there's no real need for regular users to know these counts.  They were confusing to people anyway.


gingervitis:
yay, cleanups!

PR #233, opened 2020-10-09
Add system chat message type

gabek:
I split out regular chat messages and the new "system" message.  I added some arbitrary styling to the system one, just to make it look a little different, but we can iterate on that later since we're removing avatars and stuff anyway.

image

### DIFF
--- a/webroot/js/components/chat/chat-message-view.js
+++ b/webroot/js/components/chat/chat-message-view.js
@@ -1,0 +1,46 @@
+import { h, Component } from '/js/web_modules/preact.js';
+import htm from '/js/web_modules/htm.js';
+const html = htm.bind(h);
+
+import { messageBubbleColorForString } from '../../utils/user-colors.js';
+import { formatMessageText, formatTimestamp } from '../../utils/chat.js';
+import { generateAvatar } from '../../utils/helpers.js';
+
+export default class ChatMessageView extends Component {
+  render() {
+    const { message, username } = this.props;
+    const { image, author, body, timestamp } = message;
+
+    const formattedMessage = formatMessageText(body, username);
+    const avatar = image || generateAvatar(author);
+    const formattedTimestamp = formatTimestamp(timestamp);
+
+    const authorColor = messageBubbleColorForString(author);
+    const avatarBgColor = { backgroundColor: authorColor };
+    const authorTextColor = { color: authorColor };
+
+    return html`
+      <div class="message flex flex-row items-start p-3">
+        <div
+          class="message-avatar rounded-full flex items-center justify-center mr-3"
+          style=${avatarBgColor}
+        >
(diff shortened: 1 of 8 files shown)


gingervitis:
Unless we're expecting system messages to be super crazy looking in the future, we don't need to maintain 2 different files that look almost the same. ;)

It's great that you broke it out into its own component, (which ideally should be a functional component since it doesn't hold any state or lifecycle 😉).  We can just make your current `chat-message-view`  read `message.type` from props , and conditionally render the small area that is different from a regular message.

gabek:
I do think they should look different, but I have no idea what that look is.  I don't really care too much as long as they're different.  I didn't spend any time thinking about the design here, I just wanted the new type to be available.  We can redesign the system message later.

webroot/js/components/chat/chat-message-view.js
```diff
@@ -0,0 +1,46 @@
...
+    const { image, author, body, timestamp } = message;
+
+    const formattedMessage = formatMessageText(body, username);
+    const avatar = image || generateAvatar(author);
+    const formattedTimestamp = formatTimestamp(timestamp);
+
+    const authorColor = messageBubbleColorForString(author);
+    const avatarBgColor = { backgroundColor: authorColor };
+    const authorTextColor = { color: authorColor };
+
```
gingervitis, line 21: Since it seems that only a few tailwind classes are different for the different types of message, you could make some consts here like 

```
const messageContentStyle = classNames({
 'message-content': true,
' text-sm': message.type === SYSTEM,
 'break-words': true,
  'w-full': true,
});
```
..then pass this to the message-content div, `div class=${messageContentStyle}..`
you may need to import classNames from utils.

gingervitis:
fwiw, I think the new message style you made looks great! 😁

webroot/js/components/chat/chat-message-view.js
gabek: Ok I made the system message more different so they don't have to be combined.  I just don't want to limit what we do with system messages because we combine the style with non-system messages.  I want to be able to iterate on one without touching the other.

gabek:
> fwiw, I think the new message style you made looks great! 😁

Too late already made it worse so I didn't have to combine the views.

webroot/js/components/chat/chat-message-view.js
gingervitis: you can probably still do it in one file.
you could also break out system message in to a sub-component of chat-message, so you don't have to repeat 70% of the code in this file.

gabek: That's assuming the final look of this message will have 70% of the same code.  I think this change may be premature because I don't know what it should look like, and I don't want to make decisions now that limit us later, so I'm closing this until there's a real plan behind it.

PR #238, opened 2020-10-12
Add Docker-compose

### LINKED ISSUE #186: Docker-Compose for easy setup
A docker compose file would be a great addition.  I would envision for now it having:
* owncast
* traefik or caddy.  

You should be able to quickly configure it by adjusting a couple of environment variables in the docker-compose.

Then `docker-compose up -d` has owncast sitting there with reverse proxy ready to go on hosts port 80 and 443.  The rtmp port should also be bound to the host port: 1935.

https://github.com/RocketChat/Rocket.Chat/blob/develop/docker-compose.yml - its commented out here.  But might be a good starting point

geekgonecrazy:
closes #186

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,43 @@
+version: '2'
+
+services:
+  owncast:
+    image: gabekangas/owncast:0.0.2
+    restart: unless-stopped
+    volumes:
+      - ${PWD}/config-example.yaml:/app/config.yaml # Adjust once you create your own config.yaml
+      - owncast_db:/db # Remove if you don't want chat persistant
+    command: /app/owncast --configFile=config.yaml --chatDatabase=/db/chat.db
+    ports:
+    - 8080:8080
+    - 1935:1935
+    labels:
+      - "traefik.enable=false"
+      - "traefik.http.routers.owncast.rule=Host(`live.your.org`)"
+      - "traefik.http.routers.owncast.entrypoints=web,websecure"
+      - "traefik.http.routers.owncast.tls.certresolver=mytlschallenge"
+  traefik:
+    image: traefik:latest
+    restart: unless-stopped
+    command: 
+      - --providers.docker=true
+      - --providers.docker.exposedbydefault=false
+      - --entrypoints.web.address=:80
+      - --entrypoints.websecure.address=:443
+      - --certificatesresolvers.mytlschallenge.acme.tlschallenge=true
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,43 @@
...
+    restart: unless-stopped
+    volumes:
+      - ${PWD}/config-example.yaml:/app/config.yaml # Adjust once you create your own config.yaml
+      - owncast_db:/db # Remove if you don't want chat persistant
+    command: /app/owncast --configFile=config.yaml --chatDatabase=/db/chat.db
+    ports:
+    - 8080:8080
+    - 1935:1935
+    labels:
+      - "traefik.enable=false"
```
geekgonecrazy, line 15: Could probably turn this to true since using staging letsencrypt below

docker-compose.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - "traefik.http.routers.owncast.rule=Host(`live.your.org`)"
```
gabek, line 16: Do you think it might be worthwhile to support environment variables for your hostname and email address so they don't have to edit the docker-config file itself?

gabek:
Feel free to merge whenever you're ready!

muesli:
I'd recommend not adding traefik as a service to the default docker-compose. While you can nicely reverse-proxy owncast with traefik (doing so myself!), people most likely already got their reverse proxy setup running. In that case the extra traefik layer is just cumbersome.

muesli:
How about something really simple:

```yaml
version: "2"

services:
  owncast:
    image: gabekangas/owncast:0.0.2
    restart: unless-stopped
    volumes:
      - config.yaml:/app/config.yaml # Adjust once you create your own config.yaml
      - content.md:/app/data/content.md
      - owncast_db:/db # Remove if you don't want chat persistant
    ports:
      - 1935:1935
      - 8080:8080

volumes:
  owncast_db:
```

gabek:
I think @geekgonecrazy's original thought was to have something that allowed people to go from nothing to a fully working implementation.  Maybe the better bet is have a simple compose file like you suggested, and then put a "full example" in documentation somewhere as a way to show how you can get it all working together.

gabek:
How about we throw this into the `examples` directory?

geekgonecrazy:
Putting in examples would work.  It's too bad it isn't easier to use variables like helm charts.  So could use to enable a service in the docker-compose

PR #256, opened 2020-10-15
Variable collisions; Possible use of nil value

bitspill:
Renamed a few variable which had a collision shadowing a package or interface name and added/moved a few error checks to ensure `nil` values do not get used resulting in a nil pointer panic

### DIFF
--- a/controllers/admin/hardware.go
+++ b/controllers/admin/hardware.go
@@ -9,8 +9,8 @@ import (
 
 // GetHardwareStats will return hardware utilization over time
 func GetHardwareStats(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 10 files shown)


gabek:
This is awesome, thank you!  I'll give it a spin within the next day or so and we can get it merged in.

gabek:
Looks great, just took it for a sanity check run.  Thanks for tightening this up for us!  Let us know if you happen to have a live stream of yours that we can check out, I'd love to see what you're building!

gabek:
Also, @bitspill, if you're at all interested, feel free to join our Rocket.Chat if you'd like to join us or following along!
https://owncast.rocket.chat/

PR #275, opened 2020-10-20
Add Javascript update Github Actions workflow

gabek:
@mattdsteele 

I was having issues last night, but I recreated this from scratch with minor tweaks and things seem to be ok today.

Last night I saw it erroring out and committing all of `node_modules`, but I think the later was a problem with the actual workflow, and even after I fixed it pending PRs were still using the old workflow, so when they merged in they brought all those bogus commits with them.

So I _think_ this is what we want?  Let me know if this looks good to you.

Test commits: https://github.com/owncast/owncast/commits/gek/js-workflow

### DIFF
--- a/.github/workflows/javascript-packages.yaml
+++ b/.github/workflows/javascript-packages.yaml
@@ -1,0 +1,28 @@
+name: javascript-packages
+on:
+  push:
+    paths:
+      - build/javascript/package.json
+
+jobs:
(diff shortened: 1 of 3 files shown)


mattdsteele (approved):
LGTM!

gabek:
Thanks for looking, Matt!

PR #281, opened 2020-10-22
Custom thumbnail poster component

gabek:
There's probably bugs and cases I'm not catching.  But if somebody wants to take a look at this and take it for a spin, please do!  It's to have a nicer looking poster that refreshes without flickering.  It uses the default global transition duration, but I'd probably slow it down if I could figure out how to override that.  But when I try to override it with something like `transition: opacity 2.0s;` it just ignores the animation all together and snaps.  No biggie.  I'm sure there's improvements that can be made here.


gingervitis:
I made some mild refactors. and updated the video-only component.

PR #289, opened 2020-10-24
First pass at centralized database reference. Closes #282

### LINKED ISSUE #282: Refactor database reference so it can be used in other packages
Right now the database is opened and accessed directly within the `chat` package.  We should refactor this so the database lives on its own and can be injected into packages who need it, such as chat, config, etc.  This is especially needed with our file-based sqlite database, as you can't open it multiple times from different places.

gabek:
This moves the database reference out of the `chat` package and into a new `data` package.  `GetDatabase()` is a new public method that other packages can use to grab a reference to the database.

The location of the database is changed from `./chat.db` to `data/owncast.db` as it will no longer just be used for chat history (config values, recordings, events, etc).  It can still be overwritten via the command line flag of `database`.

These changes will need to be called out in a changelog for 0.0.3.

### DIFF
--- a/core/chat/persistence.go
+++ b/core/chat/persistence.go
@@ -14,10 +14,10 @@ var _db *sql.DB
 
 func setupPersistence() {
 	_db = data.GetDatabase()
-	createTable(_db)
+	createTable()
 }
 
-func createTable(db *sql.DB) {
+func createTable() {
 	createTableSQL := `CREATE TABLE IF NOT EXISTS messages (
 		"id" string NOT NULL PRIMARY KEY,
 		"author" TEXT,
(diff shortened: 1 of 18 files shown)


core/chat/persistence.go
```diff
@@ -2,33 +2,19 @@ package chat
...
-		if err != nil {
-			log.Fatal(err.Error())
-		}
-	}
-
-	sqliteDatabase, _ := sql.Open("sqlite3", file)
-	_db = sqliteDatabase
-	createTable(sqliteDatabase)
+	_db = data.GetDatabase()
+	createTable(_db)
```
mattdsteele, line 17: nit: I don't think you need both the `_db` param and to store it in the `_db` field

mattdsteele (approved):

Issue #198, opened 2020-09-28
When transcoder fails and RTMP is dropped we don't reset state


gabek:
We still think somebody is broadcasting after a failed broadcasting session, making it impossible for somebody to re-connect to the service to broadcast in the future.

The case I saw this:

* I connected and started streaming.
* The transcoder failed due to an unrelated issue.
* I got disconnected.
* Instance still things I'm connected and rejects any future inbound RTMP.

gabek:
Once #151 lands I'll look at this separately.

gabek:
This is actually a bit of a mess at the moment since the `rtmp` package imports `core` in order to reach out and let it know when a RTMP connection starts/stops and to set broadcaster details.  But because of that core can't import `rtmp` in order to fire `Disconnect()` on it due to cyclical imports.  I'll start thinking about how this can be refactored.

PR #290, opened 2020-10-24
Stop rtmp from reaching back into core

gabek:
This change addresses the issue where the `rtmp` package reaches back into `core`, therefore making it impossible, due to cyclical imports, to do things like reach into `rtmp` to fire `Disconnect` from core.

To work around this I'm doing the following:

* When starting `rtmp` pass along the two methods `rtmp` would need (`SetStreamAsConnected`, `SetBroadcaster`) as callbacks, so it can call them without reaching into `core`.
* `rtmp` keeps references to these functions and calls them just like it used to.

This is probably not the most Go'ish way of doing this, but admittedly I prefer it to packages randomly reaching into other packages and breaking the separations of concerns.  In this scenario now rtmp only has access what what is specifically passed to it as dependancies and nothing more.  And as a benefit these functions can be un-exported.

The core issue behind this one is #198, and this PR fixes that.

Let me know what you think.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -11,6 +11,7 @@ import (
 	"github.com/owncast/owncast/config"
 	"github.com/owncast/owncast/core/chat"
 	"github.com/owncast/owncast/core/ffmpeg"
+	"github.com/owncast/owncast/core/rtmp"
 	"github.com/owncast/owncast/models"
 	"github.com/owncast/owncast/utils"
 	"github.com/owncast/owncast/yp"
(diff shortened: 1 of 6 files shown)


mattdsteele:
You might be able to make this more idiomatic by passing in a few channels and send a message back, but I don't think it's that much different from a callback. And both are better than circular package dependencies!

gabek:
That would certainly fit in better, but yeah, the idea is pretty much the same.  @geekgonecrazy @graywolf336 do you have any thoughts?

gabek:
I never looked much into it, but I did a little searching to see if callbacks are as foreign in Go as I thought, and it seems like maybe not.  There's plenty of people who talk about using them, so I guess I'll just keep it as is, since it's a simple solution.  At some point I'd like to re-architect a bit and actually draw out who needs to talk to what so it's not as much wild west.

Issue #291, opened 2020-10-25
Refactor + speed up build script


gabek:
Not the end of the world since we don't perform builds that often, but it's a bit of a quality life improvement.  Plus it's probably time to add some structure to this script, prepare it for the future, and make it more readable and maintainable.  It's doing more than it started out doing.

Right now we iterate over all the different binaries we need to build and pass each, individually, to [`xgo`](https://github.com/karalabe/xgo), and then package up each resulting build into a zip file, with all the webroot and extra stuff. 

What we should do is pass these binary builds to `xgo`, let it build all of them at once, and then use the resulting binaries to package up the release.

It's similar, but different, because each time you run `xgo` it spins up a new docker container, so right now we spin up three containers, but we should only do this once.

gabek:
Closing this as it's not important to prioritize.

PR #292, opened 2020-10-25
Create a production build of Tailwind at release time

gabek:
It makes a production build of Tailwind, pipes it through `postcss` to use `cssnano`, and saves it to the temp directory.  When each release is packaged up it copies in this file.

I also filed #291 to spend some time updating the build script since it's time to take it a little more seriously.

### DIFF
--- a/build/javascript/package-lock.json
+++ b/build/javascript/package-lock.json
@@ -4,6 +4,32 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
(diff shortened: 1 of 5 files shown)


mattdsteele:
Yeah I think this looks good!

Similarly another optimization we should do for a production build is to minify/bundle the vendor web assets, using some of Snowpack's [optimization features](https://www.snowpack.dev/#optimized-builds).

We could potentially do the same for first-party JS code, though that starts to degrade the hackability of a release by end-users.

PR #317, opened 2020-11-03
Bundle and serve admin

gabek:
This is a WIP for building the admin, bundling it into the owncast project and then serving it from `/admin`.  Obviously the admin isn't complete yet, but there's something there enough to test bundling it.

1. Clone the admin repo to a temp directory from Github.
1. Build + export the admin (npm and next stuff).
1. Copy the exported admin site to the `admin` dir in the owncast project.
1. Run `pkger`.
1. Remove `admin` directory as it's now bundled into `pkged.go`.

Couple notes:
1. The admin source is still hard coded pointing to `localhost`, so only test this locally.
1. `pkger` needs to be available on your machine: https://github.com/markbates/pkger#installation
1. `pkger` is being run from a hardcoded location.  Not sure how we should point to it with Go modules, since we no longer use the $GOPATH to point to go binaries.

Pinging @jmhobbs too but it wouldn't let me set you as a reviewer :P


gabek:
@jeyemwey Were you ever able to run the admin?  I know you had issues at one point with links not going anywhere, but that's not anything @gingervitis or I have seen.

jeyemwey:
I have to check again and come back to you. I hope that I can check that tomorrow.

jeyemwey:
I have started the admin page again now and it looks like Next is just taking some time to render the pages initially which was weird to me. Should be fine once Next built and exported the pages.

I have no issues releasing this now :)

gabek:
Ahhh yeah.  It's fast when it's rendered.  Thanks for looking at it!

admin

gabek:
The admin obviously isn't done, but I'm going to merge this in so we can start to work with it and on it more closely tied to the main app.

PR #329, opened 2020-11-09
[316] HLS playlist and segment desync monitor.

### LINKED ISSUE #316: Monitor HLS content and warn if it's not updating
If HLS playlists are not being updated with new segments after `(segmentLengthSeconds * 3)` seconds then log out a warning.

We should be able to keep track of this and warn from the `HLSHandler`.

jmhobbs:
Uses the HLS handler to monitor for segment writes that do not update the playlist.

Closes #316

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -25,7 +25,7 @@ var (
 	_broadcaster *models.Broadcaster
 )
 
-var handler ffmpeg.HLSHandler
+var handler *ffmpeg.HLSHandler = &ffmpeg.HLSHandler{}
 var fileWriter = ffmpeg.FileWriterReceiverService{}
 
 //Start starts up the core processing
@@ -45,9 +45,8 @@ func Start() error {
 	// The HLS handler takes the written HLS playlists and segments
 	// and makes storage decisions.  It's rather simple right now
 	// but will play more useful when recordings come into play.
-	handler = ffmpeg.HLSHandler{}
-	handler.Storage = _storage
-	fileWriter.SetupFileWriterReceiverService(&handler)
+	handler = ffmpeg.NewHLSHandler(_storage, config.Config.GetVideoSegmentSecondsLength())
+	fileWriter.SetupFileWriterReceiverService(handler)
 
 	if err := createInitialOfflineState(); err != nil {
 		log.Error("failed to create the initial offline state")
--- a/core/ffmpeg/hlsHandler.go
+++ b/core/ffmpeg/hlsHandler.go
@@ -1,22 +1,44 @@
 package ffmpeg
 
 import (
+	"context"
+	"path/filepath"
+	"time"
+
 	"github.com/owncast/owncast/models"
+	log "github.com/sirupsen/logrus"
 )
 
 // HLSHandler gets told about available HLS playlists and segments
 type HLSHandler struct {
 	Storage models.StorageProvider
+	Monitor *hlsVariantWriteMonitor
+}
+
+// NewHLSHandler returns an initialized HLSHandler with monitor running.
+func NewHLSHandler(storage models.StorageProvider, segmentLength int) *HLSHandler {
+	monitor := newHlsVariantWriteMonitor(
+		context.TODO(),
+		log.StandardLogger(),
+		time.Duration(segmentLength*3)*time.Second,
+	)
+
+	return &HLSHandler{
+		Storage: storage,
+		Monitor: monitor,
+	}
 }
 
 // SegmentWritten is fired when a HLS segment is written to disk
 func (h *HLSHandler) SegmentWritten(localFilePath string) {
 	h.Storage.SegmentWritten(localFilePath)
+	h.Monitor.SegmentWritten(filepath.Dir(localFilePath), time.Now())
 }
 
 // VariantPlaylistWritten is fired when a HLS variant playlist is written to disk
 func (h *HLSHandler) VariantPlaylistWritten(localFilePath string) {
 	h.Storage.VariantPlaylistWritten(localFilePath)
+	h.Monitor.VariantPlaylistWritten(filepath.Dir(localFilePath), time.Now())
 }
 
 // MasterPlaylistWritten is fired when a HLS master playlist is written to disk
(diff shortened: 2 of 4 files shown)


core/core.go
```diff
@@ -25,7 +25,7 @@ var (
 	_broadcaster *models.Broadcaster
 )
 
-var handler ffmpeg.HLSHandler
+var handler *ffmpeg.HLSHandler = &ffmpeg.HLSHandler{}
```
jmhobbs, line 28: I had to initialize the pointer here because in [setupStorage](https://github.com/owncast/owncast/blob/master/core/storage.go#L12) it assigns to `handler.Storage`.  I can rework that bit, as it's essentially a no-op as far as I can tell, but I wanted to minimize changes.

core/ffmpeg/hlsHandler.go
```diff
@@ -1,22 +1,44 @@
...
 type HLSHandler struct {
 	Storage models.StorageProvider
+	Monitor *hlsVariantWriteMonitor
+}
+
+// NewHLSHandler returns an initialized HLSHandler with monitor running.
+func NewHLSHandler(storage models.StorageProvider, segmentLength int) *HLSHandler {
+	monitor := newHlsVariantWriteMonitor(
+		context.TODO(),
+		log.StandardLogger(),
```
jmhobbs, line 22: This is the same logger called with the default logrus methods.

core/ffmpeg/hlsVariantWriteMonitor.go
```diff
@@ -0,0 +1,95 @@
...
+	segmentEvents  chan hlsWriteEvent
+	logger         *log.Logger
+}
+
+type hlsWriteEvent struct {
+	Dir  string
+	Time time.Time
+}
+
+func newHlsVariantWriteMonitor(ctx context.Context, logger *log.Logger, notificationThreshold time.Duration) *hlsVariantWriteMonitor {
```
jmhobbs, line 21: I had to accept a logger here so I could write tests.  It's a bit out of the norm since it doesn't use the logrus default methods.

Additionally I added a context in case this needs clean shutdown/reload later on.

core/ffmpeg/hlsVariantWriteMonitor_test.go
```diff
@@ -0,0 +1,141 @@
...
+		t.Errorf("incorrect log error level output: %v", hook.LastEntry().Level)
+	}
+
+	expected := `HLS playlist at "hls/0" not updated with recent segments (8s drift)`
+	if hook.LastEntry().Message != expected {
+		t.Errorf("log message does not match\nexpected: %q\n  actual: %q", expected, hook.LastEntry().Message)
+	}
+}
+
+func Test_hlsVariantWriteMonitor_segmentDelay(t *testing.T) {
```
jmhobbs, line 45: If the time between segments has been delayed long enough, I don't count that as a playlist write miss, since it could be from a stream disconnect and then reconnect.

core/ffmpeg/hlsVariantWriteMonitor.go
```diff
@@ -0,0 +1,95 @@
...
+			continue
+
+		case event = <-h.segmentEvents:
+			variant, ok = variants[event.Dir]
+			if !ok {
+				variant = &hlsVariantWriteHistory{event.Time, event.Time}
+				drift = 0
+			} else {
+				drift = event.Time.Sub(variant.lastPlaylistWrite)
+				viable = event.Time.Sub(variant.lastSegmentWrite) < notificationThreshold
```
jmhobbs, line 84: If the time between segments has been delayed long enough, I don't count that as a playlist write miss, since it could be from a stream disconnect and then reconnect.

gabek:
In case you noticed it, I was pulling down your PR on a machine with a new OS install, and I realized we didn't die if ffmpeg wasn't found (just threw an error), so I fixed that and accidentally pushed it to your branch :-1:.  But I fixed it.  Just thought I'd clarify :)  _Now_ I can play with your sweet updates.

core/ffmpeg/hlsVariantWriteMonitor.go
gabek: I'm probably missing something, but it seems like `viable` is always ending up as `false` as it's using the same `notificationThreshold` that it's using for notifying against.  Sorry if I missed the whole point of something :)

gabek:
I mentioned that basic performance monitor I had previously added, but with your change it adds structure and the concept of a standalone (and testable!) monitor.  This is really great, because with that we can use it as a base for any other HLS monitoring, too.  Given you're keeping a true `hlsVariantWriteHistory` I think I can move the old alert to follow the same pattern since they're using similar data (the old one is simply the speed of segment generation).

core/ffmpeg/hlsVariantWriteMonitor.go
jmhobbs: 🤔  Not positive I understand.  `notificationThreshold` is constant in this scope (it's assuming all variants write HLS segments at the same rate). So I'm taking the current segment write time - last segment write and seeing if it's greater than the allowed drift of playlist/segment writes.  

So a segment written > notificationThreshold from playlist triggers the log, but if no segments have been written in that same window, I assume that is due to a disconnect in streaming.  `Test_hlsVariantWriteMonitor_segmentDelay` covers this bit.

I think the reset on disconnect is probably the better option, I'll see if I can work that in.

core/ffmpeg/hlsVariantWriteMonitor_test.go
```diff
@@ -0,0 +1,141 @@
...
+	monitor := newHlsVariantWriteMonitor(ctx, logger, time.Second*5)
+
+	monitor.VariantPlaylistWritten("hls/0", moment)
+	monitor.SegmentWritten("hls/0", moment)
+	monitor.SegmentWritten("hls/0", moment.Add(time.Second*4))
+	monitor.SegmentWritten("hls/0", moment.Add(time.Second*8))
+
+	time.Sleep(50 * time.Millisecond)
+
+	if len(hook.Entries) != 1 {
```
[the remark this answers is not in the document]
jmhobbs: Yeah!  It's nice logrus provides it, so I decided to take advantage of it.

gabek:
Going to close this as time as passed and I dropped the ball on it, but if the need arises I'll come back to this and port it forward.

PR #345, opened 2020-11-12
Log potential errors returned by http.ListenAndServe

muesli:
We'd otherwise ignore any errors, for example when the listen fails due to the port already being used.


gabek:
Nice!  This certainly gives a clearer failure state in those cases.  Thank you!

PR #348, opened 2020-11-12
Add GoReleaser config and GitHub workflow

muesli:
This automatically builds new packages / binaries for:

- Linux (amd64 & x86)
- Debian packages (amd64 & x86)
- RPM packages (amd64 & x86)
- Alpine/APK packages (amd64 & x86)
- Brew formula

It also adds a GitHub workflow that automatically tries to build a new snapshot on push/PR, verifying that the current state of git would be releaseable.

You can try building a new snapshot release locally yourself, just run:

```bash
goreleaser release --snapshot --rm-dist --skip-publish
```

To do a proper release, you would typically first tag the release, then run GoReleaser like this:

```bash
git tag -a -s -m "v1.2.3" "v1.2.3"
goreleaser release --rm-dist
```

This will publish the tag on git, build all the packages / binaries, generate a ChangeLog, and eventually create a new GitHub release that all those assets get uploaded to.

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -1,0 +1,26 @@
+name: goreleaser
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Deps
+        run: sudo apt-get update -qq && sudo apt-get install gcc-multilib libc6-dev-i386-cross
+
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          args: release --snapshot --skip-publish --skip-sign --rm-dist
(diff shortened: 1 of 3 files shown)


leuc:
I can work on this. ~~What is needed to complete this? Is goreleaser a good fit or are there other options too?~~

A proper apk/deb/rpm would simplify Docker installs and it's a good way to configure systemd as well, to close #923 

After trying out this PR, goreleaser seems to be a pretty good and portable choice. Give that OwnCast is already targeted at a broad spectrum of platforms, it makes sense to describes the build and packing in YAML.

I would propose a _gradual migration_, to move away from the custom `build/*/*.sh` scripts.

First have goreleaser create the current release archives without signing, then see about the apk/dep/rpm. OwnCast needs to be able to use [custom config, log and state directories](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#id-1.14.4.3.6.2), so it's not quite there yet. Would be good to do it right from the start, instead of `/opt/owncast`.

Given that the GitHub Actions Workflow can not (or rather should not) create signed releases, i am not entirely sure about the (long term) use?

I am gonna have a look and see if i can get the packing of the archives working, so they are identical to what comes out of `build/release/build.sh`. (This PR currently only builds the binary, no admin UI etc)

Issue #357, opened 2020-11-12
Clean up linter warnings


gabek:
@muesli pointed us to `golangci-lint` and it surfaced a bunch of things we should address.  Most importantly the errors that are being ignored.  Bulk pasting them here for reference so we can start to look at them.

```
core/ffmpeg/thumbnailGenerator.go:39: core/ffmpeg/thumbnailGenerator.go:39: Line contains TODO/BUG/FIXME: "TODO: evaluate if this is ever stopped" (godox)
				//TODO: evaluate if this is ever stopped
core/ffmpeg/fileWriterReceiverService.go:107:52: `returnError` - `r` is unused (unparam)
func returnError(err error, w http.ResponseWriter, r *http.Request) {
                                                   ^
core/ffmpeg/fileWriterReceiverService.go:85: unnecessary trailing newline (whitespace)

	} else if strings.HasSuffix(path, ".ts") {
core/ffmpeg/fileWriterReceiverService.go:101: unnecessary trailing newline (whitespace)

	} else if strings.HasSuffix(path, ".m3u8") {
core/storageproviders/local.go:38:8: Error return value of `s.Save` is not checked (errcheck)
	s.Save(localFilePath, 0)
	      ^
core/storageproviders/local.go:52:8: Error return value of `s.Save` is not checked (errcheck)
	s.Save(localFilePath, 0)
	      ^
core/storageproviders/local.go:66:12: Error return value of `utils.Copy` is not checked (errcheck)
	utils.Copy(filePath, newPath)
	          ^
core/storageproviders/s3Storage.go:121:25: Error return value of `s.rewriteRemotePlaylist` is not checked (errcheck)
	s.rewriteRemotePlaylist(localFilePath)
	                       ^
core/storageproviders/s3Storage.go:194:2: ineffectual assignment to `err` (ineffassign)
	err = p.DecodeFrom(bufio.NewReader(f), false)
	^
core/storageproviders/local.go:26:3: S1000: should use for range instead of for { select {} } (gosimple)
		for {
		^
core/storageproviders/s3Storage.go:25:54: S1019: should use make(map[string]string) instead (gosimple)
var _queuedPlaylistUpdates = make(map[string]string, 0)
                                                     ^
metrics/alerting.go:54:12: unnecessary conversion (unconvert)
	return int((values[len(values)-1].Value + values[len(values)-2].Value) / 2)
	          ^
[...]

PR #358, opened 2020-11-13
Start cleaning up linter errors.

gabek:
This cleans up #357, a whole bunch of linter warnings.  Mostly around returning errors from things.

The downside of going through and fixing bulk warnings is I could have made a typo, or reversed a bool, or something else stupid.  So if a couple of you wouldn't mind just taking this branch for a test run, and taking a look over the changes, I'd feel better about it.

### DIFF
--- a/controllers/admin/index.go
+++ b/controllers/admin/index.go
@@ -39,8 +39,7 @@ func ServeAdmin(w http.ResponseWriter, r *http.Request) {
 
 	mimeType := mime.TypeByExtension(filepath.Ext(path))
 	w.Header().Set("Content-Type", mimeType)
-	_, err = w.Write(b)
-	if err != nil {
+	if _, err = w.Write(b); err != nil {
 		log.Errorln(err)
 	}
 }
--- a/controllers/admin/index.go
+++ b/controllers/admin/index.go
@@ -27,7 +27,7 @@ func ServeAdmin(w http.ResponseWriter, r *http.Request) {
 	f, err := pkger.Open(path)
 	if err != nil {
 		log.Warnln(err, path)
-		errorHandler(w, r, http.StatusNotFound)
+		errorHandler(w, http.StatusNotFound)
 		return
 	}
 
@@ -39,9 +39,12 @@ func ServeAdmin(w http.ResponseWriter, r *http.Request) {
 
 	mimeType := mime.TypeByExtension(filepath.Ext(path))
 	w.Header().Set("Content-Type", mimeType)
-	w.Write(b)
+	_, err = w.Write(b)
+	if err != nil {
+		log.Errorln(err)
+	}
 }
 
-func errorHandler(w http.ResponseWriter, r *http.Request, status int) {
+func errorHandler(w http.ResponseWriter, status int) {
 	w.WriteHeader(status)
 }
--- a/core/chat/client.go
+++ b/core/chat/client.go
@@ -94,13 +94,22 @@ func (c *Client) listenWrite() {
 		select {
 		// Send a PING keepalive
 		case msg := <-c.pingch:
-			websocket.JSON.Send(c.ws, msg)
+			err := websocket.JSON.Send(c.ws, msg)
+			if err != nil {
+				log.Errorln(err)
+			}
 		// send message to the client
 		case msg := <-c.ch:
 			// log.Println("Send:", msg)
-			websocket.JSON.Send(c.ws, msg)
+			err := websocket.JSON.Send(c.ws, msg)
+			if err != nil {
+				log.Errorln(err)
+			}
 		case msg := <-c.usernameChangeChannel:
-			websocket.JSON.Send(c.ws, msg)
+			err := websocket.JSON.Send(c.ws, msg)
+			if err != nil {
+				log.Errorln(err)
+			}
 		// receive done request
 		case <-c.doneCh:
 			_server.remove(c)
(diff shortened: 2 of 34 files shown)


core/chat/client.go
```diff
@@ -94,13 +94,22 @@ func (c *Client) listenWrite() {
...
 		// Send a PING keepalive
 		case msg := <-c.pingch:
-			websocket.JSON.Send(c.ws, msg)
+			err := websocket.JSON.Send(c.ws, msg)
+			if err != nil {
+				log.Errorln(err)
+			}
 		// send message to the client
 		case msg := <-c.ch:
 			// log.Println("Send:", msg)
```
jmhobbs, line 103: nitpick, could delete this line

utils/clientId.go
```diff
@@ -18,8 +18,11 @@ func GenerateClientIDFromRequest(req *http.Request) string {
...
 
 	// Create a MD5 hash of this ip + useragent
-	hasher := md5.New()
-	hasher.Write([]byte(clientID))
+	hasher := md5.New() //nolint
+	_, err := hasher.Write([]byte(clientID))
+	if err != nil {
+		log.Fatalln(err)
+	}
 	return hex.EncodeToString(hasher.Sum(nil))
```
jmhobbs, line 26: Not a bug, but you could trim this section down to:
```go
	clientID := strings.Join(ipAddressComponents, ":") + req.UserAgent()

	return hex.EncodeToString(md5.Sum([]byte(clientID))) //nolint
```

core/chat/persistence.go
```diff
@@ -48,7 +51,10 @@ func addMessage(message models.ChatMessage) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	tx.Commit()
+	err = tx.Commit()
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	defer stmt.Close()
```
muesli, line 59: I think this defer should be moved up. Granted, currently we'd always abort if an error happens between the `stmt` declaration and this defer, but this could change and would potentially leak resources.

jmhobbs (commented):
Not seeing any obvious typos or anything, just a couple suggestions.

controllers/admin/index.go
```diff
@@ -39,9 +39,12 @@ func ServeAdmin(w http.ResponseWriter, r *http.Request) {
 
 	mimeType := mime.TypeByExtension(filepath.Ext(path))
 	w.Header().Set("Content-Type", mimeType)
-	w.Write(b)
+	_, err = w.Write(b)
+	if err != nil {
```
graywolf336, line 43: I don't know how others feel, but I personally prefer to write these type of two lines as one. So:

```go
if _, err = w.Write(b); err != nil {
```

core/chat/persistence.go
```diff
@@ -61,6 +67,9 @@ func getChatHistory() []models.ChatMessage {
 	if err != nil {
 		log.Fatal(err)
 	}
+	if rows.Err() != nil {
```
graywolf336, line 70: Is this this different than the `err`'s value? If so, might change it to be:

`if err := rows.Err(); err != nil {`

utils/clientId.go
graywolf336: This way looks a lot cleaner 👏

core/chat/persistence.go
muesli: It is different indeed: it will return the last error that occurred during iteration over the result set. It should be called after the `for rows.Next()` loop to ensure that we really handled all rows in the result.

core/rtmp/rtmp.go
```diff
@@ -90,7 +90,10 @@ func HandleConn(c *rtmp.Conn, nc net.Conn) {
 	log.Infoln("Incoming RTMP connected.")
 
 	pipePath := utils.GetTemporaryPipePath()
-	syscall.Mkfifo(pipePath, 0666)
+	err := syscall.Mkfifo(pipePath, 0666)
+	if err != nil && err.Error() != "file exists" {
```
muesli, line 94: We should probably use `os.IsNotExist` here. I assume it would also be returned by `syscall.Mkfifo`.

core/chat/persistence.go
gabek: Nice catch!

utils/clientId.go
gabek: I'm actually not able to get that work because `md5.Sum()` returns back a statically sized byte slice, `[16]byte`, and `hex.EncodeToString` doesn't like that.  I'm not sure why?  How is a certain sized slice still not a slice of that type?

muesli: Just create a copy of it, like:

```go
b := md5.Sum([]byte(clientID))
hex.EncodeToString(b[:])
```

gabek: Nice!  Thanks for the tip.

controllers/admin/index.go
gabek: Agreed.  Too bad there wasn't a linter rule for that!  I won't go back the code base and change them all but we should standardize on that going forward.

gabek:
Thank you everyone for your feedback!  Hopefully these updates addresses them and then we can get this in as to not cause conflicts.

muesli:
LGTM!

gabek:
Merging, thanks everyone!

PR #362, opened 2020-11-15
Use a multistage Docker build to slim down the image a bit

muesli:
This uses a separate stage for building, so we don't drag in gcc, build-base, and linux-headers into the final Owncast image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,21 @@
-FROM golang:alpine
-EXPOSE 8080 1935
-RUN mkdir /app 
-ADD . /app
+# Builder
+FROM golang:alpine AS builder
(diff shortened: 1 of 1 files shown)


gabek:
The final image we distribute is actually built with https://github.com/owncast/owncast/blob/master/build/release/Dockerfile-build and the root `Dockerfile` is just a simple one for developers, but it'd be nice to slim down that build too.  I wonder if these builds shouldn't be two different Dockerfiles.

muesli:
Oh, I see! Yeah, I think we should unify that approach, especially as we point users to this Dockerfile from the docs, too. Also means less maintenance.

PR #363, opened 2020-11-15
Extended config example

muesli:
I imagine this may help people setting up Owncast for the first time, especially the `streamQualities` settings.

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -14,6 +14,8 @@ instanceDetails:
     - software
     - streaming
 
+  nsfw: false
+
   # https://owncast.online/docs/configuration/#external-links
   # for full list of supported social links.  All optional.
   socialHandles:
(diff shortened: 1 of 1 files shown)


config-example.yaml
```diff
@@ -26,5 +28,24 @@ videoSettings:
...
+      encoderPreset: veryfast
+
+    - high:
+      videoBitrate: 1600
+      encoderPreset: veryfast
+
 # Set to true if you don't want the service checking for future releases.
 disableUpgradeChecks: false
+
+# Enable if you want your stream to be listed in the Owncast directory.
```
gabek, line 48: Since we're not ready to publicly release the directory, I've left these out of the config examples.  It's probably best not to encourage people to actively register with the directory yet since it hasn't been tested much.  But if people dig through the config struct and figure it out themselves, then they can go for it :)

muesli: Fair enough, updating the PR.

gabek:
And just as a side note, this is the default config we use when creating releases.  So I do worry a bit about having three video configurations being enabled by default, out of the box.  I think we might need to split up `config-example.yaml` and `config-default.yaml`.  The default config should probably only have one video variant as to not clobber people's hardware right out of the box.  But we can just copy example to default, just keep the middle quality and update the build scripts to use that instead.  I'll file a todo.

PR #366, opened 2020-11-16
Do not highlight usernames inside links

### LINKED ISSUE #156: Do not perform username highlighting inside urls
In this case I typed "http://gabekangas.com"

image

mattdsteele:
Fixes #156

This switches out manual username highlighting with the [markjs](https://markjs.io/) library. Basic testing it fixes the weird link issue:

image

I'm only marginally happy with the library here; it's [not designed for vdom/strings](https://github.com/julmot/mark.js/issues/115) and required constructing some throwaway nodes, which is a little ugly. It's also a little bulkier than I remember (14k). This would probably be something to worth keeping an eye on replacing in the future.


gabek:
Sorry that become more of a hassle than anticipated, but thank you so much for looking into it!  :confetti_ball: 

It's certainly a net positive, even given the the library is a bit bulky.  At least it's not buggy :)

gabek:
Just tested and works great!  I added @gingervitis as a reviewer so she's aware of the changes.

gingervitis:
Thanks for looking into this! I figured we'd have to look into using some markdown library for this soon.. 

Anyway, I did notice this happening with the update. If you have a space in your user name, it will treat each word in the name as a token. 
image

mattdsteele:
Ah interesting - I'll see if there are any settings to configure that
behavior!

gingervitis:
Oh cool I see markjs has `markRegExp()` in addition to `mark()` that we could use. Though my regex skills currently aren't crafty enough to make sure it doesn't find the user name  if it's part of a longer word.

mattdsteele:
@gingervitis The matching on individual words should be fixed with the latest push. And yeah, there's quite a few cool options we could mess with! https://markjs.io/#mark

gingervitis:
Oh sweet! Can you also add this option?
```
accuracy: {
  value: 'exactly',
  limiters: [",", ".", "'", '?', '@'],
},
```
These will also help cover these situations:
image


I think after that we're good to go :)

PR #369, opened 2020-11-17
a fix for message list not scrolling to bottom when new messages come in

gingervitis:
We're telling the message list to jumpToBottom when new messages come in but it hasn't seemed all that responsive lately..  @gabek  suspects that it might be because of the extra parsing and markup that we're adding to the message before rendering it. 
So instead of setting an arbitrary timeout before scrolling, I'm trying out this DOM mutation observer. We'll scroll to the bottom when we detect in the DOM that a new child has been added to the message container div.


gabek:
It works great for me both with and without the changes in #366.  Awesome!  So much nicer.

gabek (approved):

gabek:
Curious if @mattdsteele has any thoughts about this approach, if not, merge away!

mattdsteele:
I really like mutation observers. They're a little brute force, but I think it'd work well here. Let's go! :raised_hands:

PR #373, opened 2020-11-17
Simplify Logo requirement from users.

gingervitis:
- Only require 1 logo file, instead of a `small` and `large` one.  Just require `logo`.
- Update frontend so that primary header logo will ALWAYS be Owncast logo, which would help with design consistency and Owncast branding. 
- User's logo will still display below in "user content" area, as the large profile circle.
- Make recommendation in notes that logo be a square of at least 150x150px with some padding so that it won't be cropped off when we add a circle border radius around it.


gabek:
Thank you!

gabek:
Heads up to everyone: @geekgonecrazy @mattdsteele @jmhobbs @jeyemwey @mattdsteele @muesli @bitspill  @graywolf336 

Because this change uses the existing `logo` key, your old config will be providing a map instead of a string and the config will complain.  I've added an explicit error message telling people to look at the changelog if they updated Owncast.

PR #382, opened 2020-11-18
Add database schema versioning & migration skeleton

muesli:
This also adds a bit of error handling to the database init.

### DIFF
--- a/core/data/data.go
+++ b/core/data/data.go
@@ -6,11 +6,16 @@ package data
 
 import (
 	"database/sql"
+	"fmt"
 	"os"
 
 	"github.com/owncast/owncast/config"
 	"github.com/owncast/owncast/utils"
 	log "github.com/sirupsen/logrus"
+)
+
+const (
+	schemaVersion = 0
 )
 
 var _db *sql.DB
@@ -19,7 +24,7 @@ func GetDatabase() *sql.DB {
 	return _db
 }
 
-func SetupPersistence() {
+func SetupPersistence() error {
 	file := config.Config.DatabaseFilePath
 
 	// Create empty DB file if it doesn't exist.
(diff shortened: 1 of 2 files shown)


core/data/data.go
```diff
@@ -32,6 +37,66 @@ func SetupPersistence() {
...
+		_, err := db.Exec("INSERT INTO config(key, value) VALUES(?, ?)", "version", schemaVersion)
+		if err != nil {
+			return err
+		}
+		version = schemaVersion
+	}
+
+	// is database from a newer Owncast version?
+	if version > schemaVersion {
+		return fmt.Errorf("incompatible database version %d (versions up to %d are supported)",
```
gabek, line 70: For consistency could you use `log` instead of `fmt`?

core/data/data.go
```diff
@@ -32,6 +37,66 @@ func SetupPersistence() {
...
+
+	_db = db
+	return nil
+}
+
+func migrateDatabase(db *sql.DB, from, to int) error {
+	fmt.Printf("Migrating database from version %d to %d\n", from, to)
+	for v := from; v < to; v++ {
+		switch v {
+		case 0:
```
gabek, line 89: I wonder for the future, if instead of continuing to grow this switch statement for every migration, if we could create a directory of migrations and run them, by name, dynamically.  Something like a `runMigration0(), runMigration1()`, etc.  I know Go isn't a big fan of dynamic programming, but could be a nice way to organize this code.

Not that we need this now, but we could always think about adding it later if we think might be a worthwhile approach.

gabek (commented):
This is a great addition, thank you!  With 0.0.3 we figured there's so few installs, and the data we're storing is not critical, so we wouldn't yet jump into migrations, but I think this is the last time we can argue that.  Next time we do things the right way!

core/data/data.go
muesli: This isn't printing to screen though, this is merely generating an `error`. The error gets caught by `main()` and will use `log` to print it to screen.

core/data/data.go
muesli: That's one way, even though the code isn't too nice with the use of `reflect` then. Another alternative I've used before is a slice of `[]string`s, containing the SQL commands for each migration. You can then simply range over that `[][]string` and access each version's migration steps by index:

```go
for _, v := range dbMigrations[from:to] {
    for _, s := range v {
        db.Exec(s)
        ...
    }
}
```

core/data/data.go
gabek: Sorry, I commented on the wrong line:

86: fmt.Printf("Migrating database from version %d to %d\n", from, to)
90: fmt.Printf("Migration step from %d to %d\n", v, v+1)

core/data/data.go
gabek: Oh mixing the two ideas makes a good solution:
A slice of references to functions, so the migration logic still doesn't have to live inline.

```
for _, f := range v {
  f()
```

Just ideas, we can come back to that some other time haha.

core/data/data.go
muesli: Done!

gabek:
Thank you! :confetti_ball:

PR #384, opened 2020-11-19
End to end automated tests run on Github workflows

gabek:
This will:

- Build and run an instance of Owncast
- Download a copy of ffmpeg
- Stream a static test file to Owncast
- Hit the APIs and test against them.

If you go to `test/automated` and run `run.sh` it'll run locally as well.

It's a little hacky dealing with the copy of ffmpeg, by downloading a copy and adding the directory of that download into the script's path so it can be found.

Feedback appreciated!  @jmhobbs you might have some thoughts around this since you're better at writing shell scripts than I am :)


geekgonecrazy:
This actually looks pretty solid.  Doing those tests in javascript actually is super easy to follow.  The amount of bash here while having to work around with ffmpeg is small and easy to manage.  This looks pretty good to me

jeyemwey:
Looks good! I had some problems since my `config.yaml` was old, but after that it worked like a charm. Maybe we can add a backup-restore-mechanism since some tests rely on very precise configs?

gabek:
A safer way than to be messing with local config files is just to specify a different config file when running the owncast binary, so I'll update it to do that and skip the copying of default.

mattdsteele:
Yeah I like all of this!

Doesn't have to be part of this go-round, but we may want to extend this with some browser-based testing; i.e. verify no console errors, the page renders and particular selectors work, screenshots don't look super wonky, etc.

I've used [Puppeteer](https://pptr.dev/) to do this in the past and it's worked really well; it has a nice JavaScript API, and downloads a headless Chromium so it's pretty self-contained and not nearly as flaky as Selenium or the like.

gabek:
I think that's a great idea!  Filed this with your suggestions of some high level items to test for.

PR #431, opened 2020-12-06
chat fixes and optimizations

### LINKED ISSUE #416: Some chat message was able to grind the web page to a halt for everyone
1. The web page froze for everyone.
1. I restarted the server.
1. Reloading the web page caused the same issue, still froze.
1. I guessed it was something in the chat database, so I removed the chat database and restarted the server again.
1. And this fixed it.

I have a copy of the database from this time, so I'll try and get it to be reproduced, but nothing in the chat database stands out as a culprit. But the steps above point to it being chat related.

gingervitis:
- format messages on didMount instead of didUpdate. will also prevent bad setState loops when message is blank
- convert message.js to functional comp
- prevent extra rerenders in messages and chat with shouldComponentUpdate checks

Fixes #416

### DIFF
--- a/webroot/js/components/chat/chat-message-view.js
+++ b/webroot/js/components/chat/chat-message-view.js
@@ -11,26 +11,39 @@ import { convertToText } from '../../utils/chat.js';
 import { SOCKET_MESSAGE_TYPES } from '../../utils/websocket.js';
 
 export default class ChatMessageView extends Component {
-  async componentDidUpdate(prevProps) {
-    const { message, username } = this.props;
-    if (prevProps.message === message && this.state.formattedMessage) {
-      return;
-    }
-    const { body } = message;
-
-    const formattedMessage = await formatMessageText(body, username);
-    this.setState({
-      formattedMessage
-    });
-
+  constructor(props) {
+    super(props);
+    this.state = {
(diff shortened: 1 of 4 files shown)


webroot/js/components/chat/chat-message-view.js
```diff
@@ -11,26 +11,39 @@ import { convertToText } from '../../utils/chat.js';
...
+  shouldComponentUpdate(nextProps, nextState) {
+    const { formattedMessage } = this.state;
+    const { formattedMessage: nextFormattedMessage } = nextState;
+
+    return (formattedMessage !== nextFormattedMessage);
+  }
+
+  async componentDidMount() {
     const { message, username } = this.props;
-    if (prevProps.message === message && this.state.formattedMessage) {
```
gingervitis, line 16: @mattdsteele I converted this to `componentDidMount` instead!  We also found that rare edge case of formattedMessage coming up empty, which ended up failing out of here, and then calling setState in a crazy loop!

mattdsteele:
Neat; this looks good to me!

I don't have a ton of experience with functional components (or hooks, for that matter), but when they make for cleaner code I'm up for the refactors.

mattdsteele (approved):

PR #476, opened 2020-12-19
Update docker-compose.yaml example

Marzal:
With this change works on 0.3 release.

chatDatabase -> database

### DIFF
--- a/examples/docker-compose.yaml
+++ b/examples/docker-compose.yaml
@@ -2,7 +2,7 @@ version: '2'
 
 services:
   owncast:
-    image: gabekangas/owncast:0.0.3
+    image: gabekangas/owncast:latest
     restart: unless-stopped
     volumes:
       - ${PWD}/config-example.yaml:/app/config.yaml # Adjust once you create your own config.yaml
--- a/examples/docker-compose.yaml
+++ b/examples/docker-compose.yaml
@@ -2,12 +2,12 @@ version: '2'
 
 services:
   owncast:
-    image: gabekangas/owncast:0.0.2
+    image: gabekangas/owncast:0.0.3
     restart: unless-stopped
     volumes:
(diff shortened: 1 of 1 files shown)


examples/docker-compose.yaml
```diff
@@ -2,12 +2,12 @@ version: '2'
 
 services:
   owncast:
-    image: gabekangas/owncast:0.0.2
+    image: gabekangas/owncast:0.0.3
```
muesli, line 5: I think I would change this to `owncast:latest` to avoid having to update this file with every release.

examples/docker-compose.yaml
```diff
@@ -2,12 +2,12 @@ version: '2'
...
     restart: unless-stopped
     volumes:
       - ${PWD}/config-example.yaml:/app/config.yaml # Adjust once you create your own config.yaml
       - owncast_db:/db # Remove if you don't want chat persistant
-    command: /app/owncast --configFile=config.yaml --chatDatabase=/db/chat.db
+    command: /app/owncast -configFile=config.yaml -database=/db/chat.db
```
muesli, line 10: We should add the `config.yaml` and `chat.db` as volumes to this docker-compose, so people understand how they can configure and maintain state for a dockerized version of Owncast.

examples/docker-compose.yaml
Marzal: Agree, done

examples/docker-compose.yaml
Marzal: chat.db is already on owncast_db volume, isn't it?

geekgonecrazy (approved):

geekgonecrazy:
This looks good to me 👍

gabek:
Thank you so much for this! :1st_place_medal:

PR #495, opened 2020-12-22
Auto play when stream is live

sytranvn:
Currently video is not autoplaying when stream is live. This is workaround to play the video without sound. User must unmute by herself.

### DIFF
--- a/webroot/js/components/player.js
+++ b/webroot/js/components/player.js
@@ -85,7 +85,11 @@ class OwncastPlayer {
 
     this.vjsPlayer.volume(getLocalStorage(PLAYER_VOLUME) || 1);
     this.vjsPlayer.src(source);
-    // this.vjsPlayer.play();
+    this.vjsPlayer.play().catch(() => {
+      // Autoplay was prevented. Start playback without sound.
+      this.vjsPlayer.muted(true);
+      this.vjsPlayer.play();
+    });
   }
 
   handleReady() {


sytranvn:
https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#mei
Once user engaging score with the site is enough the sound will be unmuted automatically. 

chrome://media-engagement
Origin | Sessions | Sessions with playback | Last Playback | Is High | Score
-- | -- | -- | -- | -- | --
http://owncast.local:8080 | 7 | 6 | 2020-12-22T11:16:35.555Z | Yes | 0.30
-- | -- | -- | -- | -- | --

gingervitis:
I believe we had actually implemented it this way in an initial iteration of the player (autoplay, muted), but we ended up deciding against it because of some UX reasons that I can't recall. @gabek do you remember?

sytranvn:
>  * don't autoplay. just show play button when stream is online so that it's easier to start playign without looking for the unmute button

I found this on #56 . Maybe we should still try to play if the page already qualified MEI. Otherwise let user press the play button.

sytranvn:
I will look more into that PR.

gabek:
I believe the logic was it was found to be less annoying to have to manually press play than it was to manually press "unmute".

gingervitis:
Since it was really just a matter of un-muting being too annoying to click, we could make a large unmute button during initial autoplay, kinda like what YouTube does.
image

If there are problems autoplaying, we could just show the large play button.

gabek:
I'm open to it if it's generally preferred.  I kind of like how we have it, with the stills updating, and then the user can click play whenever they're ready.  A big unmute button feels like it would be kind of weird, and it would still force the user to press a button, just a different one.  Let's talk about it next week: https://github.com/owncast/owncast/discussions/167#discussioncomment-228568

muesli:
I think this could be a server-side setting, too. As an admin of an instance I might not even want people to auto-play unless they're really watching, just to save some bandwidth.

sytranvn:
> I'm open to it if it's generally preferred. 

> Since it was really just a matter of un-muting being too annoying to click, we could make a large unmute button during initial autoplay, kinda like what YouTube does.

From my perspective, it's reasonable to see the video of streaming website auto plays if content is available. Compare UX of
- User goes to the website
- User clicks play button to watch live stream 
- User clicks pause/mute if she wants
Vs
- User go to the website to watch live stream
- User clicks pause/mute if she wants
___
> I think this could be a server-side setting, too. As an admin of an instance I might not even want people to auto-play unless they're really watching, just to save some bandwidth.

Yes, controlling  bandwidth budget is also important. Configurable autoplay is nice.

gabek:
Hi @sytranvn, we decided not to move forward on autoplay right now.  If there's enough demand we might look at it again in the future.  Sorry about that!

sytranvn:
@gabek No problem.

fenarinarsa:
> Hi @sytranvn, we decided not to move forward on autoplay right now. If there's enough demand we might look at it again in the future. Sorry about that!

Would it be possible to reconsider?

Literally every live stream service do autoplay when the video goes live or the user just lands on the player's page, even Peertube does it.

It's actually counter-intuitive to have to press the play button for users. Especially when the stream stops and starts again (like, when you have technical issues).

Also I don't understand the "mute" thing, other services don't have the issue?

gabek:
> Also I don't understand the "mute" thing, other services don't have the issue?

Other services do have the same issue, but it's likely been a while since you've seen it occur because the media engagement index of the sites you see autoplay working on have met the requirements.

https://developer.chrome.com/blog/autoplay/#media-engagement-index

If you open Twitch, for example, for the first time on a browser it falls under the same rules. It'll start autoplaying but you'll have to unmute it.


<blockquote><div> Chrome Developers</div><div><strong><a href="https://developer.chrome.com/blog/autoplay/">Autoplay policy in Chrome - Chrome Developers</a></strong></div><div>Learn best practices for good user experiences with the new autoplay policies in Chrome.
</div></blockquote>

fenarinarsa:
> If you open Twitch, for example, for the first time on a browser it falls under the same rules. It'll start autoplaying but you'll have to unmute it.

Oh, okay, thanks for the info!

In this case I think enabling autoplay in mute mode would be a far better option, since when a user comes to a streaming website, they're expecting the stream to play automatically if it's live.

And actually my first test was to stop the stream and make it start again to simulate a stream incident, and my viewers complained that the stream didn't restart automatically.

gabek:
> In this case I think enabling autoplay in mute mode would be a far better option

That brings it back to the original discussion where having a tiny mute button that auto-hides in the player control bar is much harder to find, or even know it's there, compared to a big, obvious, play button. It would create a much worse user experience than what it is right now.

If we were to go this route it would require redesigning the player to support a "Big unmute" like mentioned [here](https://github.com/owncast/owncast/pull/495#issuecomment-749686871). And unless somebody wanted to spearhead doing autoplay right, that's not something that's in the cards.

gabek:
Another option is to not auto-play at all until the media engagement index meets the requirements, then that bad user experience around muting, or player redesign requirements wouldn't be an issue.

It would cause confusion, however, since autoplay would work for some people, and not for others. Plus it might surprise people to see it autoplaying when visiting the page since the first X times they visited it autoplaying didn't occur. So it's just more things to think about.

koraysels:
if autoplay is out of the question (which i find very sad  because it is easier to make users unmute than press play) Can we adjust the frequency of the stills being updated?

gabek:
Thumbnails get generated at 20 seconds, currently. It takes additional system resources to create them, so it's a balance. This hasn't been something that has been an issue previously, but if it's something many people want to change, it can be discussed. Please don't hesitate to file an issue for the request.

gabek:
> if autoplay is out of the question (which i find very sad because it is easier to make users unmute than press play)

I do believe it is way easier to press the big play button than to unmute. However, I think we could prototype a flow of:

If unmuted autoplay is available (MEI is met), then play it. If it is not met, then leave it how it is now

 I don't know if there's an API to check MEI status, but this could be a compromise if it's possible and worth prototyping.

koraysels:
I understand, maybe make it a config option so people can decide for themselves?  

so 
- 1 config for frequency of thumbnail generation,  

- configure toggle autoplay:
    - if enabled:
      that means autoplay AND muted) 
    - if disabled: behaviour is like before. 

No need to check MEI.. i don't understand what the fuss is about. autoplay is allowed by browsers as long as the content is muted by default....

gabek:
> autoplay is allowed by browsers as long as the content is muted by default

And it's a really lousy experience. The player controls disappear when you move the mouse away, you won't know why audio isn't working. When you hover over the player, the mute button is small because it should be. When paused, the player gives you one big action: Play. That makes way more sense.

I was just giving a suggestion that could explore the problem. No big deal.

koraysels:
> > autoplay is allowed by browsers as long as the content is muted by default
> 
> And it's a really lousy experience. The player controls disappear when you move the mouse away, you won't know why audio isn't working. When you hover over the player, the mute button is small because it should be. When paused, the player gives you one big action: Play. That makes way more sense.
> 
> I was just giving a suggestion that could explore the problem. No big deal.

The approach would depend on how the player is implemented. Personally, I would have the player display a prominent "unmute" button if the audio is muted. The reason being, in my case(s) it's for a radio station that also streams video, so having that visual prompt, I agree, is important. Flexibility and offering choices are key in this situation—rather than imposing a preference, it’s better to let the user of owncast decide. I’m more than happy to help contribute to the frontend code if needed.

koraysels:
could we pick this up again? i see more people wanting a muted autoplay for the video. this is a really really easy fix. i dont understand why this owuld not be possible?

gabek:
I made a suggestion above that didn't seem to be sufficient so it never moved forward.

koraysels:
I don't really see that suggestion anywhere apart from an opinion about useabilty. What I’m proposing is simply switching the buttons: have an autoplaying muted video with a prominent unmute button (similar to the current large play button). ( I am biased because I am looking at this from an online radio perspective, not a streaming service that streams viedeogames or other stuff)

We could implement this either via a separate embed route or by using query parameters.

gabek:
https://github.com/owncast/owncast/pull/495#issuecomment-1435812583

koraysels:
I dont think you understand. when muted, autoplay does work for everyone. Anyway, i made the chagne myself I implemented a seperate route `/embed/video/autoplay` 
this does exactly what I meant. here : https://github.com/koraysels/owncast

gabek:
I understand fully. I'll reiterate what I've said above: Unmuting is a worse experience than pressing play. That's why I proposed what I did. It's the best of both options.

Issue #426, opened 2020-12-05
browser notification when new messages come in (and when stream starts)


gingervitis:
If your owncast browser tab is hidden, it'd be nice to have some sort of notification that tells you that there are new messages, or that the stream has started. Perhaps a subtle indicator in the <title>.

gingervitis:
closed with #496

PR #496, opened 2020-12-22
misc issue fixes

### LINKED ISSUE #428: Hide viewer count on the player if the stream isn't live

### LINKED ISSUE #491: ublock origin breaks javascript due to filename
Somebody posted the following on Reddt:

> One thing I noticed is that the webpage didn't load at all for me at first, which is because uBlock Origin blocks https://watch.owncast.online/js/components/social-icons-list.js (using Firefox 84, but that shouldn't matter).
>
> There's a rule that's part of the "Fanboy’s Social" blocklist that seems to block any filename that contains social-icons-. I'm not entirely sure, but I think that list is part of the default uBlock Origin installation, so for any of those users the site probably also doesn't load.

gingervitis:
- rename social-icons to prevent adblockers from blocking scripts, fix #491
- hide viewer count when stream is off, fixes #428 
- show a notice in document title of num messages if window is blurred, #426


gabek:
This all works great!  THANK YOU!

PR #501, opened 2020-12-22
wrap tags so they don't extend off the view

### LINKED ISSUE #500: Too many tags mess with the mobile layout
This streamer has added so many tags that the container gets too wide for the layout on mobile. It is currently not possible to scroll from left to right.

I can only suspect that this will also be an issue on smaller desktop screens or if a steamer adds even more tags.

As a user, I want to see all stream information without having to scroll.

As a streamer, I want to add as many tags as I want without thinking of destroying the web interface.

gingervitis:
fixes #500

### DIFF
--- a/webroot/js/app.js
+++ b/webroot/js/app.js
@@ -372,7 +372,7 @@ export default class App extends Component {
           (tag, index) => html`
             <li
               key="tag${index}"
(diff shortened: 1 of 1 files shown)


jeyemwey:
Looks good to me!

PR #517, opened 2020-12-23
Added new endpoints to modify settings in-memory

SeaLife:


SeaLife:
Those endpoints are just a "quick-win" to update stream details while owncast is running.

They need to be reworked for https://github.com/owncast/owncast/milestone/4 :)

Issue #314, opened 2020-11-01
Add message removal feature in admin


gabek:
At some point post-0.0.4 we should think about adding chat moderating to the admin.

----
Edit: 12/16

- Add new chat page to admin that lists the chat messages.
- Add API for toggling the existing `visible` flag of messages.
- Create new websocket `UPDATE` event and send updated messages.
- Client should update its existing copy of these messages with the updated ones and re-render, filtering on `visible`.

aaronpk:
frankly having a chat that's not behind any sort of authentication needs chat moderation yesterday. even just the ability to delete individual messages would be better than nothing.

Gargron:
I'm assuming no moderation can be done until #321. I do wonder what's possible with no user accounts. I'm guessing any kind of kick, ban, or timeout would have to be IP-based.

gabek:
Yeah, the wheels are turning, including adding a more robust identity for chat so we have both the IP address and a pseudo-user we can manage: https://github.com/owncast/owncast/discussions/440.  It's still being fleshed out, though.

0.0.3 is the first version with a web admin of any sort, so now we can start building management tools.

aaronpk:
Fancier moderation and account stuff is great and all but please consider shipping a "delete message" feature ASAP before worrying about account/identity/etc stuff.

shirt-dev:
There should also be an option to disable the chat completely.

gingervitis:
I'll start looking into creating a UI to delete messages in the Admin sometime this week.

gabek:
First pass at moderation is in master for the next release.

PR #524, opened 2020-12-26
Update chat message visibility for moderation

gingervitis:
Currently this branch creates an API for the Admin side to post an update to change the visibility of specific message IDs. 

TODO: 
- [x] update websockets across all connected clients to notify the message visibility update
- [x] update client UI to show/remove messages based on visibility state as well as websocket updates

Addresses parts of:
https://github.com/owncast/owncast/issues/314


router/router.go
```diff
@@ -82,6 +82,9 @@ func Start() error {
 	// Get warning/error logs
 	http.HandleFunc("/api/admin/logs/warnings", middleware.RequireAdminAuth(admin.GetWarnings))
 
+	// Update chat message visibilty
+	http.HandleFunc("/api/admin/chat/updatemessagevisibility", middleware.RequireAdminAuth(admin.UpdateMessageVisibility))
```
SeaLife, line 86: I guess, we sould talk about the API Endpoints in general to give them better names...
This should probably use path variables instead of content.
It would probably also easier to just delete the message completely from the database.

The endpoint should probably look like something: `/api/admin/chat/messages/:messageId` and use the Method `DELETE` instead of post (in case we'r deleting the message).
Even if we'r not deleting the message, the endpoint would then accept a body to modify the message, not the message id itself, because this should be in the URL.

In REST-API's unique identifiers should be in the URL... :)

gabek: We've certainly never set out to be a strict REST API.  We're more of a "the request is the payload" style.  In this case any number of IDs can be sent as an array.

SeaLife: Ouh, okay..

Issue #525, opened 2020-12-27
Add Ko-fi platform


CosmicToast:
https://ko-fi.com/
It's a patreon-like but with a polar-opposite culture (tip jar by default, as opposed to exchange of regular money for privileges).

I would add it myself, but looking at the code it looks like the primary thing that needs changing is just adding image data to a gif file (and maybe some js/css? I couldn't tell).

gabek:
Yeah!  Do you happen to have a nice quality ko-fi icon we can use?  Otherwise I'm sure we can dig one up.

CosmicToast:
They've got a whole package of assets for this exact purpose!
https://more.ko-fi.com/brand-assets (under "Logos")
they ask that it not be modified, but yeah, as long as it's used verbatim (same as twitter iirc?) it should be fine and as high a quality as you can get

gabek:
Awesome, thank you!

Issue #515, opened 2020-12-23
Improve quality of social handle icons


SeaLife:
Hey,

i would suggest improving the overall quality of the social handle brand icons.

I recommend switching to a Font or to SVG's. :)

Fonts i can recommend: 

- https://materialdesignicons.com/
- https://fontawesome.com/icons?s=brands 

Both fonts contains all major brand icons as font icon. :)

Regards
SeaLife

PR #528, opened 2020-12-27
#515 switched to SVG icons, #525 added Ko-Fi platform

SeaLife:


webroot/js/components/platform-logos-list.js
```diff
@@ -1,10 +1,52 @@
...
+  const platformInfo = SOCIAL_PLATFORMS[platform.toLowerCase()];
+  const inList = !!platformInfo;
+  const iconSupplied = !!icon
+  const name = inList ? platformInfo.name : platform;
+  const colorSupplied = !!color
+
+  const finalIcon = iconSupplied ? icon : (inList ? platformInfo.icon : '/img/socialhandles/default.svg')
+  const finalColor = colorSupplied ? color : (inList ? platformInfo?.color ? platformInfo.color : '#000000' : '#000000');
+
+  const style = `-webkit-mask-image: ${finalIcon}; mask-image: url(${finalIcon}); background-color: ${finalColor};`
```
gingervitis, line 19: Hey @SeaLife I tried checking out your SVG icons, but I couldn't get them to show up. For some reason the `mask-image` styles aren't getting rendered for me.

image

gingervitis:
Thanks for working on converting some logos to SVG! 
I'm currently working on building out administrating the config through the Admin site, and I was thinking instead of keeping a current JS object of icons I was considering having the admin just read all the available icon files from a folder (them being svg format or gif or png or whatever). Could we just define the icon's color directly in svg code? Like just add the fill color or stroke color as attributes to the xml. like `<path fill="#1DA1F2" ...>` for Twitter.

SeaLife:
Hey, this is probably possible @gingervitis, im not that deep into SVGs, so i got my SVG files from https://simpleicons.org/ and they are all color-less, but the required color is stated there.

You can ofc. modify/change this. (or i could do this, probably)

SeaLife:
@gingervitis i've added the suggested change regarding the colors. Its still working for me :)
I need to remove some left-overs in the go-code and you should be ready to go :)

gabek:
Looks ok to me when I just tested it.  Curious if it works for ginger, as she had issues with them displaying properly before.

SeaLife:
In my last commit i changed the classes, because i set the color for the icons in the SVG files now, so @gingervitis should re-test this :)

gabek:
Tested and looks good on my end.

gingervitis:
Hey sorry finally got around to looking at it! Seems to be working for me now. Looks good.

PR #570, opened 2021-01-02
Manually bust m3u8 caches

### LINKED ISSUE #567: Custom S3 metadata for playlist files (CDN caching)
Now that we're serving playlist files from S3, they're subject to similar storage and caching rules as anything else we upload into a bucket. But since we frequently upload new playlists (while the `.ts` segments are essentially immutable), we need ways to optimize the cache TTL, so CDNs know how frequently they need to refresh files from the bucket.

The goal should be to always request fresh playlist files from the bucket, while the segments can live on the edge for long periods.

In particular I'm running into this with DigitalOcean's Spaces, along with their built-in CDN. It's possible to [set a global cache TTL on the bucket](https://www.digitalocean.com/docs/spaces/how-to/manage-cdn-cache/#ttl), but none of the values are particularly useful for playlists (the minimum TTL is 1 minute; which isn't great either for playlists or media files IMO)

It's possible to [set a custom TTL](https://www.digitalocean.com/docs/spaces/how-to/set-file-metadata/) on a file by setting a metadata value when uploading to S3 (the `x-amz-meta-max-age` key, with a value in seconds).

One way to resolve this might be to add a new configuration property in the s3 space that lets the user add arbitrary S3 metadata elements. Maybe something like:

```yaml
s3:
  playlistAttributes:
    - x-amz-meta-max-age: 0
```

Before starting to put a fix together I wanted to get another set of eyes on the problem. How are others managing CDN-fronted playlist files? Is there a more general approach I should look at?

mattdsteele:
This reimplements the `beforeRequest` videojs hook to add query params to make each playlist request unique. I'm not sure when it stopped working, but the API appears to have changed [with this commit](https://github.com/videojs/http-streaming/commit/22af0b29e56e8a25d55737b0148da0ab5e4d461b).

This fixes #567 and lets us avoid taking more specific implementations to cache-bust on various CDNs.

### DIFF
--- a/webroot/js/components/player.js
+++ b/webroot/js/components/player.js
@@ -57,13 +57,15 @@ class OwncastPlayer {
   }
 
   init() {
-    this.vjsPlayer = videojs(VIDEO_ID, VIDEO_OPTIONS);
-
-    this.vjsPlayer.beforeRequest = function (options) {
-      const cachebuster = Math.round(new Date().getTime() / 1000);
-      options.uri = `${options.uri}?cachebust=${cachebuster}`;
+    videojs.Vhs.xhr.beforeRequest = options => {
+      if (options.uri.match('m3u8')) {
+        const cachebuster = Math.round(new Date().getTime() / 1000);
+        options.uri = `${options.uri}?cachebust=${cachebuster}`;
+      }
       return options;
     };
+
+    this.vjsPlayer = videojs(VIDEO_ID, VIDEO_OPTIONS);
 
     this.addAirplay();
     this.vjsPlayer.ready(this.handleReady);


gabek:
Nice!  I didn't notice it stopped working, either.  I guess for most cases our cache headers are doing their job, but it makes sense with a CDN that this extra nudge makes a difference.  Thank you!

geekgonecrazy:
In the future would be nice to some how optimize this a bit and make more efficiently bust cache

gabek:
We do have server-side caching logic, but when you put external storage + CDNs in the mix then our own server's caching logic is out of play.  Take a look at https://github.com/owncast/owncast/blob/master/controllers/index.go#L54-L59 to see how we're doing it locally.  It'd be nice if we could do it on 3rd party services, too, though.

geekgonecrazy:
Right I guess that's what I mean is just cdn cache busting.  Something in future to let us handle like twitch does. Seems to chain m3u8 playlists with different names.  I guess the only one they don't cache is the initial one and they do a redirect instead to the latest m3u8.

But future stuff.  No need to dive into this now

PR #574, opened 2021-01-02
Implement webhook events for external integrations

geekgonecrazy:
- [x] Stream started
- [x] Stream stopped
- [x] Chat message sent
- [x] Message visibility changed (from moderation)
- [x] Username changed

Yaml format for this is:
```
webhooks:
  - url: {something that will receive all events}
  - url: {something else}
    events: 
    - streamStarted
    - streamStopped
  - url : {something else}
    events: 
    - messageSent
```

Event types:

* messageSent
* usernameChanged
* visibilityToggle
* streamStarted
* streamStopped

The payloads you'll get on the webhooks:

For streamStarted 
```
{
    "type": "streamStarted"
}
```

For streamStopped
```
{
    "type": "streamStopped"
}
```

For messageSent:
```
{
    "type": "messageSent",
    "eventData": {
      "author": "geekgonecrazy3",
      "body": "<p>test again</p>",
      "rawBody": "test again",
      "id": "f2FltnaGR",
      "visible": true,
      "timestamp": "2021-01-02T17:05:15.476060425-06:00"
    }
}
```

For visibilityToggle:
```
{
    "type": "visibilityToggle",
    "eventData": {
      "author": "geekgonecrazy3",
      "body": "<p>test again</p>",
      "id": "f2FltnaGR",
      "visible": false,
      "timestamp": "2021-01-02T17:05:15.476060425-06:00"
    }
}
```

For usernameChanged:
```
{
    "type": "usernameChanged",
    "eventData": {
      "oldName": "geekgonecrazy3",
      "newName": "geekgonecrazy4",
      "image": "",
      "type": "NAME_CHANGE",
      "id": "4uSut7aGg"
    }
}
```


core/webhooks/chat.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// TODO: relocate to shared location that we and chat package can use
+const (
+	CHAT             = "CHAT"
+	NAMECHANGE       = "NAME_CHANGE"
+	PING             = "PING"
+	PONG             = "PONG"
+	VISIBILITYUPDATE = "VISIBILITY-UPDATE"
+)
```
geekgonecrazy, line 14: Trying to figure out where I can put this so it can be used by both the chat package and the webhook package.  Can refer to the one in the chat package or it ends up in a circular dependency issue

gabek: Should we move it into `models` as an exported `ChatEvent` or something, and let all the other packages just pull it from models?

gabek:
This is super awesome!  It's exciting to start adding official support for 3rd parties.

So here's what I see being the next steps:

1. The work to deprecate the config file is already in motion, so while it totally made sense to develop this feature using the config file like you did, it probably should be launched with the new configuration work.  So that means migrating your config over to the new data persistence branch that's been on going.  I can totally do that along with the rest of the config migration I'm doing.
2. And then we'd need to build a new Admin page for "Integrations" or something where you can add and remove webhooks.
3. We build one (or more!) proof of concepts as example code for people.  I was thinking of finding some thing that's popular on Twitch and build it for owncast using the webhooks in node or something, to show how it doesn't need to be built in go.  Maybe even build it on [Glitch](https://glitch.com/) or something, though I'm not sure if you can host APIs on there.

This means we can't really release this until 0.0.6, but it would come along with all of the new stuff in that release.  In the mean time we'd be able to do that example projects and documentation for people and launch 3rd party support with a nice splash.

core/webhooks/chat.go
geekgonecrazy: I could.  Maybe a type similar to the Webhook Event Type?

geekgonecrazy:
1. Absolutely agree.  Though I wonder will it be flexible enough for this?  Cause I can see adding a last successful field along side this as well.
2. Yes for sure.  Being able to add / remove easily I think will be essential.  Maybe an integration section.  With webhooks in it.  As well as api keys as the other section?

Like:
- Integrations
  - Hooks
  - API keys

3. Yes I actually want to take this to completion and run this on my server and finish integrating twitch and owncast.
4. I thought about a set of tests.  But wasn't sure if should do that now or wait until we have api endpoints to add / remove.

core/webhooks/chat.go
gabek: Yeah!  That should be an easy win against the circular dependency.

core/chat/client.go
```diff
@@ -157,15 +149,15 @@ func (c *Client) listenRead() {
 				log.Errorln(err)
 			}
 
-			messageType := messageTypeCheck["type"]
+			messageType := messageTypeCheck["type"].(string)
```
geekgonecrazy, line 152: not 100% sure about this...

gabek:
Pointed this PR at the new `external-integrations` branch.

geekgonecrazy:
Ah awesome!  😁

PR #585, opened 2021-01-05
Access tokens + Send system message external API

gabek:


controllers/admin/accessToken.go
```diff
@@ -0,0 +1,88 @@
...
+		controllers.InternalErrorHandler(w, err)
+		return
+	}
+
+	if err := data.InsertToken(token, request.Name, request.Scopes); err != nil {
+		controllers.InternalErrorHandler(w, err)
+		return
+	}
+
+	controllers.WriteSimpleResponse(w, true, "created "+request.Name)
```
geekgonecrazy, line 49: I wonder if returning name and token in json  isn’t more useful long term here?   That way if someone wanted to programmatically or via api create a token

gabek: Done.

controllers/admin/chat.go
```diff
@@ -58,3 +55,57 @@ func GetChatMessages(w http.ResponseWriter, r *http.Request) {
...
+func SendSystemMessage(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	var message models.ChatEvent
+	if err := json.NewDecoder(r.Body).Decode(&message); err != nil {
+		controllers.InternalErrorHandler(w, err)
+		return
+	}
+
+	message.MessageType = "SYSTEM"
```
geekgonecrazy, line 70: It's in my PR, but just calling this out as one we will need to adjust when merging my PR down since this is MessageType now

controllers/admin/chat.go
```diff
@@ -58,3 +55,57 @@ func GetChatMessages(w http.ResponseWriter, r *http.Request) {
...
+func SendUserMessage(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	var message models.ChatEvent
+	if err := json.NewDecoder(r.Body).Decode(&message); err != nil {
+		controllers.InternalErrorHandler(w, err)
+		return
+	}
+
+	message.MessageType = "CHAT"
```
geekgonecrazy, line 97: Same with this one

geekgonecrazy:
```suggestion
	message.MessageType = models.MessageSent
```

router/router.go
```diff
@@ -105,6 +106,20 @@ func Start() error {
...
+	http.HandleFunc("/api/admin/accesstokens", middleware.RequireAdminAuth(admin.GetAccessTokens))
+
+	// Delete a single access token
+	http.HandleFunc("/api/admin/deleteaccesstoken", middleware.RequireAdminAuth(admin.DeleteAccessToken))
+
+	// Create a single access token
+	http.HandleFunc("/api/admin/createaccesstoken", middleware.RequireAdminAuth(admin.CreateAccessToken))
+
+	// Send a system message to chat
+	http.HandleFunc("/api/admin/sendsystemmessage", middleware.RequireAccessToken(models.ScopeCanSendSystemMessages, admin.SendSystemMessage))
```
Gargron, line 119: Would it perhaps make sense to differentiate user/system message through a request param instead of separate controllers? (Although that would require moving the scope checking logic so maybe not)

gabek:
Merging this into the `external-integrations` feature branch to continue iterating.

PR #589, opened 2021-01-05
Pass at migrating webhooks to datastore + management apis

gabek:
This PR diff's against @geekgonecrazy's PR branch #574.

This creates APIs around getting, creating and deleting webhooks, storing them on disk, migrating away from the config file.

* Some of this is duplicated from the access token branch (badRequestHandler -> BadRequestHandler, etc)
* Centralized all event types to `models.EventType`

@geekgonecrazy: feel free to change anything you like.

Examples:
```
curl -X POST -H "Content-Type: application/json" -d '{"url": "http://my.webhook.biz/hook", "events": ["CHAT"]}' http://localhost:8080/api/admin/webhooks/create
```
```
curl -X POST -H "Content-Type: application/json" -d '{"id": 1}' http://localhost:8080/api/admin/webhooks/delete
```


core/webhooks/webhooks.go
```diff
@@ -43,12 +28,14 @@ type WebhookChatMessage struct {
...
 
 func SendEventToWebhooks(payload WebhookEvent) {
-	for _, webhook := range config.Config.Webhooks {
+	webhooks := data.GetWebhooksForEvent(payload.Type)
+
+	for _, webhook := range webhooks {
 		log.Debugf("Checking Webhook %s to send event: %s", webhook.Url, payload.Type)
 
-		eventsAccepted := strings.Join(webhook.Events, ",")
+		eventsAccepted := strings.Join(interface{}(webhook.Events).([]string), ",")
```
geekgonecrazy, line 36: Is this and the if below even needed any more with the new query you added?

geekgonecrazy:
Nice only the one question there.  But other then that it looks good.  I need to give it a whirl.  I like the change to more consistent event naming.  I’ll need to update integration.  But I think probably better having consistent with internal style

core/webhooks/chat.go
```diff
@@ -10,11 +10,11 @@ func SendChatEvent(chatEvent models.ChatEvent) {
...
+		webhookEvent.Type = models.MessageSent
 	case models.UserNameChanged:
-		webhookEvent.Type = UserNameChanged
+		webhookEvent.Type = models.UserNameChanged
 	case models.VisibiltyToggled:
-		webhookEvent.Type = VisibiltyToggled
+		webhookEvent.Type = models.VisibiltyToggled
 	}
 
 	webhookEvent.EventData = &WebhookChatMessage{
```
geekgonecrazy, line 20: I guess now with the types being the same we could just create webhookEvent directly and just pass MessageType directly over?

core/webhooks/webhooks.go
gabek: No longer needed, thanks for catching!

core/webhooks/chat.go
gabek: Yeah I left it just to keep the structure of what you already did, but I'm just going to assign it directly.

geekgonecrazy (approved):
This looks good to me :+1:

Issue #572, opened 2021-01-02
Chat height calculation fails on iPads


jeyemwey:
When watching a stream with an iPad, I cannot type chat messages. The keyboard lays over the chat input.

The issue gets worse if you have multiple tabs open and exists on both landscape and portrait orientation. I'm using iOS 14 and the Safari Browser.

thilobillerbeck:
I just investigated a little bit and found out, that using `vh` in mobile safari seems to be a common problem. I found a solution by placing the relative to `right` and `bottom` instead of `top` and `right`. Fix is coming soon.

PR #593, opened 2021-01-06
align the chat-views position to the bottom of the page to overcome view-height bugs

thilobillerbeck:
This is a possible fix for the problem described in #572. Since `vh` problems seem to be common in Safari on mobile devices, aligning the chat-views position to the bottom is a possible solution which seems to work on other devices too.

### DIFF
--- a/webroot/styles/chat.css
+++ b/webroot/styles/chat.css
@@ -3,7 +3,7 @@
 #chat-container {
   position: fixed;
   z-index: 9;
-  top: var(--header-height);
+  bottom: 0;
   right: 0;
   width: var(--right-col-width);
 
--- a/webroot/styles/chat.css
+++ b/webroot/styles/chat.css
@@ -10,6 +10,12 @@
   height: calc(100vh - var(--header-height));
 }
 
+@media screen and (max-width: 729px) {
+  #chat-container {
+    top: var(--header-height);
+  }
+}
+
 #message-input-container {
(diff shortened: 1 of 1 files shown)


jeyemwey:
Works f0r me, thanks for your help! image

gingervitis:
Oh cool. I currently don't have access to any devices to test on, but looks worth a shot.

gabek:
I'll give it a spin too, and merge it in if all looks good.

gingervitis:
Oh wait, are you able to scroll the messages view? (when you have lots of messages)?

gingervitis:
Ok it looks like I'm unable to scroll the the messages view in "Single column" mode (narrow screen), on both mobile and desktop browsers.

thilobillerbeck:
I introduced a media query, which should re-enable chat scrolling on mobile devices (single column). However, I am not able to test if this works on iPhones.

gabek:
Cool, I'll check it out.

gabek:
I think it's functioning as expected?  Attached is me testing it.

chat-test

gabek:
Are we good to merge this in?  Not sure if there are other edge cases we should be looking at.

gingervitis:
Ok just played around with the update in iPhone and iPad simulator as well as desktop Chrome, Safari and FF and they all look good.  Thanks!

PR #608, opened 2021-01-08
example with TLS wrapper (RTMPS) and traefik

felix-engelmann:
While the example captures the setup where the website and outgoing stream is secured, it is also important to secure the incoming RTMP stream with a TLS wrapper.
Traefik is able to wrap bare TCP streams and this pull request demonstrates an example configuration.

### DIFF
--- a/examples/docker-compose.tls-rtmp.yaml
+++ b/examples/docker-compose.tls-rtmp.yaml
@@ -10,20 +10,26 @@ services:
     command: /app/owncast -configFile=config.yaml -database=/db/chat.db
     ports:
     - 8080:8080
-    - 1935:1935
     labels:
       - "traefik.enable=false"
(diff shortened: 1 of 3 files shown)


gabek:
Would you mind creating a new example file with just the features you're trying to show an example of?  If I were to look at this docker-compose file I wouldn't be sure what it was an example of since it has traefik, minio, and the standard Owncast SSL.  Maybe a `docker-compose-rtmp-tls.yaml` or something?

felix-engelmann:
Thanks for the suggestions. It's split into 3 separate examples building on top of each other. I included the RTMP TLS in the MinIO example, because examples should provide security by default.

gabek:
Awesome, thank you @felix-engelmann!

Issue #612, opened 2021-01-08
Do not auto-close emoji picker after selecting an emoji


gabek:
So people can go all twitch-style and pick a bunch of crazy emojis to show how they think the stream is poggers, or whatever.

PR #614, opened 2021-01-10
emoji picker updates #612

gingervitis:
- don't close picker after selecting emoji,  so we can pick multiple emojis in a row 
- move up picker div to see emojis inputted into textfield
- restore focus into textfield after picker closes


gingervitis:
This may need some testing across various browser types and sizes,  regarding the positioning of the picker

gingervitis:
Just tested this in various iOS simulators and other browsers and seems good.

PR #626, opened 2021-01-13
Use a dynamic system port for internal HLS writes. Closes #577

### LINKED ISSUE #577: Support dynamic port for HLS file writes
https://github.com/owncast/owncast/blob/176861645b32534328048156e9ab79b1e615df1c/core/ffmpeg/fileWriterReceiverService.go#L37

A friend asked me what we use the 8081 port for and I didn't even know that we do use it.

We should document the usage of this port and maybe allow the user to change the port if necassary.

gabek:
Small extra hoop needs to be jumped through, you have to create a separate tcp listener and then create the web server with it in order to gain access to the port, but it works.

Ignore the test failure, that's due to the WIP datastore branch this is being merged into.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,3 +17,5 @@ var WebServerPortOverride = 8080
 var RTMPServerPortOverride = 1935
 
 var HighestQualityStreamIndex = 0
+
+var InternalHLSListenerPort = "8927"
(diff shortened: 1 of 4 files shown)


geekgonecrazy (approved):
Looks good to me.  :+1:

PR #628, opened 2021-01-14
Add user joined chat event

gabek:
If a user has previously set a custom chat username, then this event will get sent.  It's mostly useful for webhooks, but it's also displaying in the chat.

Side note: The different chat events on the server-side are getting a bit out of hand with all the per-event go channels and needs to be refactored.  But that's a different thing for a different time.


gabek:
This is now just living in the `external-integrations` branch.

PR #631, opened 2021-01-14
Add support for running frontend on URL paths other than root

### LINKED ISSUE #505: URL base/root path
Be able to set a "base" or "root" path that is prefixed to all paths to the site, in routing and link generation.

This is a standard feature of web apps, so that you can host multiple apps on the same domain using a proxy server like nginx to direct to different web apps based on the URL path.

Ex. if I set this to `basePath: "/stream"`, then `localhost:8080/stream` will be the home page, `localhost:8080/stream/admin` will be the admin, freeing up other paths on `localhost` to host other content.

rdtq:
This resolves https://github.com/owncast/owncast/issues/505

I tested this and it works great when running on `/` and also on different path, for example with following Caddy config:
```
my-owncast-instance.xyz

handle_path /my_secret_path/* {
        uri strip_prefix /my_secret_path/
        reverse_proxy 127.0.0.1:8080
}

```

That way you can have your owncast instance run privately for only those who have secret link, e.g. https://my-owncast-instance.xyz/my_secret_path/

### DIFF
--- a/controllers/emoji.go
+++ b/controllers/emoji.go
@@ -15,7 +15,7 @@ import (
 
 // Make this path configurable if somebody has a valid reason
 // to need it to be.  The config is getting a bit bloated.
-const emojiDir = "/img/emoji" // Relative to webroot
+const emojiDir = "./img/emoji" // Relative to webroot
 
 // GetCustomEmoji returns a list of custom emoji via the API.
 func GetCustomEmoji(w http.ResponseWriter, r *http.Request) {
--- a/static/metadata.html
+++ b/static/metadata.html
@@ -42,7 +42,7 @@
   <link rel="icon" type="image/png" sizes="32x32" href="/img/favicon/favicon-32x32.png">
   <link rel="icon" type="image/png" sizes="96x96" href="/img/favicon/favicon-96x96.png">
   <link rel="icon" type="image/png" sizes="16x16" href="/img/favicon/favicon-16x16.png">
-  <link rel="manifest" href="/manifest.json">
+  <link rel="manifest" href="./manifest.json">
 
   <meta name="msapplication-TileColor" content="#ffffff">
   <meta name="msapplication-TileImage" content="/img/favicon/ms-icon-144x144.png">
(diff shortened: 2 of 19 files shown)


jeyemwey:
Thanks for your contributions! I think, the admin interface might have some issues if it is not delivered from host.org/admin/ directly.

rdtq:
> Thanks for your contributions! I think, the admin interface might have some issues if it is not delivered from host.org/admin/ directly.

Indeed, admin dashboard won't work on https://my-owncast-instance.xyz/my_secret_path/admin/
However, since we are already using reverse proxy, we can work around it with the following Caddy (or similar Nginx) config:
```
my-owncast-instance.xyz

handle /admin/* {
        reverse_proxy 127.0.0.1:8080
}

handle_path /my_secret_path/* {
        uri strip_prefix /my_secret_path/
        reverse_proxy 127.0.0.1:8080
}
```

So we get working stream on https://my-owncast-instance.xyz/my_secret_path/admin/ , and working admin on https://my-owncast-instance.xyz/admin/ with one simple config.

gabek:
Like @graywolf336 said https://github.com/owncast/owncast/issues/505#issuecomment-749904447, while this might make things work right now, I do fear it'll just break for you immediately as we continue to add things.  With our very limited resources I can't say we'll be able to test additional scenarios like yours as we build new features.  We can only do our best for Owncast to work out of the box, and any custom configurations unfortunately can't be tested.  If it were just making sure javascript was loaded using relative paths then that's enforceable, but when it comes to things like rewriting how emoji URLs are generated that's likely not going to be something people will think about.

rdtq:
Ah, that makes sense! 

What if we simply return relative paths for emojis from API? Then your future code will implicitly work on both root and non-root base without ever thinking about it.

You can take a look at the latest commit, API returns relative paths for emojis now (and slight modification of emoji code to support it).

rdtq:
Cleaned up a little bit so changes look less intrusive to existing code

gingervitis:
@rdtq instead of changing paths in a bunch of files, in your own `index.html` file, can you just try adding a base href header?
For example, after the `<title>` tag add this:
`<base href="/my-secret/path/" />`
or
`<base href="https://www.mysite.com/my-secret/path/" />` 
(though you may run in to CORS issues depending on where you're serving things).

rdtq:
@gingervitis I've just tried `base` tag approach, unfortunately it didn't work. 

Looking at MDN docs at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base `base` seems to work only when paths are relative already, i.e. don't start with `/`.

PR #650, opened 2021-01-21
Only trigger responsive behavior if zoom does not happen on touch screen devices

### LINKED ISSUE #594: [Bug] Pinch zooming on mobile Safari seems to affect the responsive site behavior
As this screen recording shows, pinch-zooming on devices like an iPad seem to trigger some responsive chat-view behavior, which makes it impossible to zoom into the stream description for example.

https://user-images.githubusercontent.com/7442383/103780133-e080f000-5034-11eb-9d09-85b3eb138d6b.mp4

thilobillerbeck:
This resolves #594 by only reacting responsively to zoom actions if the device does not have a touch screen. Since pinch zooming allows users to focus on content relatively straight forward, there is no need to behave responsively (in my opinion).

### DIFF
--- a/webroot/js/app.js
+++ b/webroot/js/app.js
@@ -104,7 +104,9 @@ export default class App extends Component {
 
   componentDidMount() {
     this.getConfig();
-    window.addEventListener('resize', this.handleWindowResize);
+    if (!this.hasTouchScreen) {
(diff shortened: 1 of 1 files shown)


jeyemwey:
Does the job for me. Thanks!

gingervitis:
This works for me too! Thank you!

PR #651, opened 2021-01-21
Log hardware utilization messages as warnings instead of errors

petersveter108:
changed logging from Error to Warn for hdw utilization messages


gabek:
Thank you @petersveter108 for your first Owncast contribution!

Issue #675, opened 2021-01-31
buggy `backtick` rendering in chat messages


gingervitis:
- Messages sent with words wrapped in backtick quote characters (&grave;code&grave;) should render with monospaced text like `code`. Instead they come back with no formatting.
- When sending multiple backticks in a row (&grave;&grave;&grave;), the word `undefined` gets printed out instead.

gabek:
My guess is three backticks gets interpreted by the backend as invalid markdown and then returns back an empty string, and then our chat message formatting logic doesn't like the emptiness.

PR #676, opened 2021-01-31
Fix codeblocks in chat

craftamap:
I wrote a quick fix for #675 .

This fix consists of two parts.

Part 1:
When the chat message looks like this `  ```a ` without actual content in the code block (`a` does not count as it will set the language for the block). This will be converted into the following html by goldmark:

```html
<pre><code></code></pre>
```
This will be trimmed into an empty string by bluemonday, as both tags are forbitten. The empty string will than omited, as its empty. Therefore, undefined will be displayed in the chat. I added a quick check if the message is empty. If a message is empty, it will not be sent to the frontend/to the chat.

Part 2:

I added pre and code as allowed tags.

### DIFF
--- a/core/chat/server.go
+++ b/core/chat/server.go
@@ -117,11 +117,13 @@ func (s *server) Listen() {
 			// and standardize this message into something safe we can send everyone else.
 			msg.RenderAndSanitizeMessageBody()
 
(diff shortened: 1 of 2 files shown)


gabek:
Awesome, thank you for looking into that!

PR #745, opened 2021-02-19
<script type="module" ...> tags to reduce round trips under HTTP/2

ForestJohnson:
I did some 1st impression load time tests with owncast frontend app, testing  pre-loading the JS modules via <script type="module" src="..."> tags. The 1st test with HTTP/1.1 took about 10 seconds regardless of whether the script tags were there or not:

https://picopublish.sequentialread.com/files/owncast-first-impression-http-1.1.mkv

But watch what happens when you turn on HTTP/2:

https://picopublish.sequentialread.com/files/owncast-first-impression-http-2.mkv

Under terrible latency conditions (1 second latency) these script tags will reduce load time by ~75%.

according to caniuse this browser feature is on 92% of all clients today: https://caniuse.com/es6-module

### DIFF
--- a/webroot/index.html
+++ b/webroot/index.html
@@ -33,6 +33,44 @@
     <link href="./styles/chat.css" rel="stylesheet" />
     <link href="./styles/user-content.css" rel="stylesheet" />
     <link href="./styles/app.css" rel="stylesheet" />
+
+    <!-- The following script tags are not required for the app to run, 
+         however they will make it load a lot faster (fewer round trips) when HTTP/2 is used. 
+        
+         If you wish to re-generate this list, run the following shell command 
+         (assuming a linux or unix-ish system):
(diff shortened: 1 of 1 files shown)


gabek:
This is awesome, thank you for looking into this, Forest!  

For anybody who comes upon this, it might look counterproductive because in theory ES6 modules load files when you need them.  However, the Owncast web UI is a single page that is guaranteed to require using all of the modules.  So I'm guessing that forcing the browser to preload the modules up front end up ultimately being faster because it can do it all in a single HTTP/2 pass instead of traversing all the modules and lazy loading as needed (and they're all needed).

mattdsteele:
Yeah, this is a nice perf win! 🎉 

There's a slightly more "canonical" way to do this these days, using `<link ref="preload">`: https://web.dev/uses-rel-preload/

This might be a little safer (as it just downloads and parses, but doesn't try to execute the files), so it might be worth experimenting with?

gabek:
Nice, thanks for the sanity check!

PR #766, opened 2021-02-26
Fix the http port used not in sync with what is logged

graywolf336:
This was driving me bonkers as the port was different but it didn't display it in the logs 🙈

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -72,10 +72,8 @@ func Start() error {
 	// start the rtmp server
 	go rtmp.Start(setStreamAsConnected, setBroadcaster)
 
-	port := config.WebServerPort
 	rtmpPort := data.GetRTMPPortNumber()
-	log.Infof("Web server is listening on port %d, RTMP is accepting inbound streams on port %d.", port, rtmpPort)
-	log.Infoln("The web admin interface is available at /admin.")
+	log.Infof("RTMP is accepting inbound streams on port %d.", rtmpPort)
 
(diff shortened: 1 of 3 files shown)


core/core.go
```diff
@@ -72,7 +72,7 @@ func Start() error {
 	// start the rtmp server
 	go rtmp.Start(setStreamAsConnected, setBroadcaster)
 
-	port := config.WebServerPort
+	port := data.GetHTTPPortNumber()
```
gabek, line 75: This isn't correct if somebody overrides the port on the command line.  This is necessary if somebody is already running something on port 8080 and can't get to the admin to change it.

It should be getting set in https://github.com/owncast/owncast/blob/master/main.go#L112

graywolf336: Heh, you're right but there's a race condition which is causing the logging to be incorrect.

https://github.com/owncast/owncast/blob/1d9774e4533512ce708bdd7abeb10fa7e6a93806/main.go#L107-L109
Is called **before** the `config.WebServerPort` is set:
https://github.com/owncast/owncast/blob/1d9774e4533512ce708bdd7abeb10fa7e6a93806/main.go#L111-L122

gabek: ah yeah good catch!  `core.Start()` shouldn't be logging the web server port, that should be in `router.Start()`.

graywolf336: Logging moved to the appropriate package. :)

PR #771, opened 2021-02-26
Remove peak viewer counts from public status feed

### LINKED ISSUE #765: Remove overallMaxViewerCount and sessionMaxViewerCount from /api/status
These values used to be used in the web frontend, but they no longer are.  Instead it should only be used in `/api/admin/serverstatus`.

jeyemwey:
Fixes #765.

### DIFF
--- a/controllers/status.go
+++ b/controllers/status.go
@@ -15,6 +15,10 @@ func GetStatus(w http.ResponseWriter, r *http.Request) {
 	status := core.GetStatus()
 	w.Header().Set("Content-Type", "application/json")
 
+	// Omit these values from public
+	status.OverallMaxViewerCount = 0
+	status.SessionMaxViewerCount = 0
+
 	if err := json.NewEncoder(w).Encode(status); err != nil {
 		InternalErrorHandler(w, err)
 	}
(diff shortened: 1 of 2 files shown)


gabek:
We should return a a new api response object that doesn't include those two values instead of just setting them to zero.  We have a `adminStatusResponse` so similarly maybe a `webStatusResponse`.

jeyemwey:
`controllers/admin/status.go` uses core.GetStatus to get most of its information, so separating the status models would require to move some functionality to the admin controller instead of just removing it from the core controller.

I can do it later though.

gabek:
There's no need to touch the admin endpoint.  Use `controllers/admin/status.go` as an example.  It gets the status via `core.GetStatus()` and then packages it up in `adminStatusResponse`.  So `controllers/status.go` should get status via `core.GetStatus()` and then package it up in a new `webStatusResponse`, neither of them directly returning `core.GetStatus()` verbatim.

jeyemwey:
Updated, please check if it works out now.

gabek:
Nice, this is great!  Thank you!

Issue #716, opened 2021-02-12
Add ability to upload logo via the admin


gabek:
1. Add upload Logo to the General settings page in the Admin.
2. Create a new endpoint that accepts the upload.
2. Endpoint saves the file in the `data` directory.
3. Endpoint runs `SetLogo` with the new filename.

I've also toyed with the idea of the admin having a dropdown of all the images available in the `data` directory so you can simply select one, instead of having to type it in if you ever want to change it.  Right now you have to know/remember the name of the image that you saved.

gabek:
@nebunez I thought of you with this since you haven't had a chance to play with the backend much yet and it ties into the admin stuff you're already familiar with.

gabek:
No rush on getting this in.  It's not slated for any particular release and it'll go in when somebody wants to work on it.

PR #773, opened 2021-02-28
update logo POST route

nebunez:
The logo POST route now accepts a base64 encoded image instead of a path to an
image in the data folder. The route now saves the posted image to the data
folder, with the correct file type extension, and updates logo path in the
database appropriately.

Closes https://github.com/owncast/owncast/issues/716

### DIFF
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -1,11 +1,14 @@
 package admin
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


gabek:
Once https://github.com/owncast/owncast-admin/pull/54 is merged in this can also get merged in, since they go hand in hand.  Thanks for working on this @nebunez, looking forward for it to be in 0.0.7!

gabek (approved):

PR #783, opened 2021-03-03
Defensively handle API values

### LINKED ISSUE #781: Removing scaled height and/or scaled width from stream variant throws error

gabek:
@graywolf336 reported an issue where he tried to clear the width/height values, and the web UI sent empty strings.  Obviously the backend isn't expecting strings for a number value, so errors were thrown.  I could just fix this on the frontend, but given the API is open it's best to defensively handle this so anybody else who might send an empty string or a nil value won't get an error.

However, this feels like overkill.  @graywolf336 @muesli @geekgonecrazy do you have any other solutions to unmarshalling unexpected types that's better than manually handling the cases?

Closes #781

### DIFF
--- a/models/streamOutputVariant.go
+++ b/models/streamOutputVariant.go
@@ -1,6 +1,8 @@
 package models
 
-import "encoding/json"
+import (
+	"encoding/json"
+)
 
 // StreamOutputVariant defines the output specifics of a single HLS stream variant.
 type StreamOutputVariant struct {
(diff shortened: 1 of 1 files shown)


gabek:
You know what, if anybody sends the wrong type they should get an error and they should fix it.  This approach doesn't fly.  Closing.

muesli:
You could make them pointers, so these values are actually nullable. That said, such things typically make validating requests rather hard and the code fairly unreadable.

graywolf336:
@gabek in my opinion you made the correct call there. If they make a bad request to the server, then it’s a bad request and they should fix it. There’s a good reason for the 403 status code 😊

PR #799, opened 2021-03-09
Add support for disabling chat.

gabek:
If chat is disabled:

1. Will auto-hide the chat UI.
2. Will disable the input in case somebody gets to the chat input anyway.
3. Ignores any chat messages on the server.

(I noticed prettier formatted a lot of code that isn't related to this change, so it makes it look bigger than it actually is)


nebunez (approved):
Looks good, works in my tests. Obviously, Go lint errors should be addressed.

Issue #670, opened 2021-01-27
Extend keyboard controls


thilobillerbeck:
Since the user now has the ability to pause and play the video (#579) by pressing the "Space" key, I think it's a good idea to expand on that and introduce more keyboard controls. Some examples would be:

- Mute (M)
- Fullscreen (F)
- Volume (unsure, but maybe use arrow keys or numbers / numblock)

In addition, media keys could be included, which are less common on normal keyboards, but could aid the accessibility entertainment devices like home theatre PCs.

thilobillerbeck:
Since I already worked on the "Space" key part, I think I'll work on this issue as well.

gabek:
Awesome, thank you!

PR #804, opened 2021-03-09
Added some basic keyboard controls to the owncast frontend

thilobillerbeck:
As described in #670, I thought about some new keyboard controls for the owncast frontend. I ended up with implementing the following:

- Play / Pause (Space, P, Media-PlayPause)
- (Un)Mute (M)
- Fullscreen (F)
- Toggle chat (C)
- Volume increase (0) / decrease (9) by 10%

Maybe you have some more suggestions, which controls could be useful, when navigating the owncast frontend.


gabek:
Awesome, thank you for adding this!  I do a lot of mute toggling myself, so this will come in handy :)

webroot/js/app.js
```diff
@@ -383,9 +383,45 @@ export default class App extends Component {
     }
   }
 
+  handleMuteKeyPressed() {
+    this.player.vjsPlayer.muted(!this.player.vjsPlayer.muted());
```
gingervitis, line 387: This doesn't seem to work if you start the player already muted.  Interesting, in these cases `this.player.vjsPlayer.muted()` returns false, even if it's muted.

webroot/js/app.js
```diff
@@ -383,9 +383,45 @@ export default class App extends Component {
...
+  }
+
+  handleVolumeSet(factor) {
+    this.player.vjsPlayer.volume(this.player.vjsPlayer.volume() + factor);
+  }
+
   handleKeyPressed(e) {
-    if (e.code === 'Space' && e.target === document.body && this.state.streamOnline) {
-      this.handleSpaceBarPressed(e);
+    if (e.target === document.body && this.state.streamOnline) {
```
gingervitis, line 403: So I noticed that if you initially click to play from the play button, the focus stays on the player, and then the keyboard commands won't work. However, they will work if you take out the e.target check. I don't quite recall why the check is there. Can we remove it?

```suggestion
    if (this.state.streamOnline) {
```

thilobillerbeck: This check exists to ensure that keyboard controls are not tiggered as long as a chat input is focussed. Otherwise users would constantly start and stop their video when typing chat messages.

thilobillerbeck:
I replaced the e.target check with a check if input elements are not focussed. Sadly, I had to use each element id, since the inputs don't share properties like the tag (input) or a class. The first of which can btw. hurt accessibility.

gabek:
Same here, just tested the recent updates and wasn't able to unmute if it's muted by default.  If you're curious why it would be muted by default, I'll often view a stream, mute it, and then leave.  So when I come back it'll be saved, and muted by default.

thilobillerbeck:
Ok, I'll try to reproduce the issue.

thilobillerbeck:
It turns out, that a mute on load in vjs means, that the volume is set to 0% by default. Since mute and volume are two different things in vjs, changing the muted state wont help. I came up with the solution to set the volume to 50%, if the mute key is pressed, when the volume is set to 0%.

gabek:
Just tested, and all works for me!  Once you resolve that conflict that seems to be there for `app.js` I think we're good to go!  I'll have to make sure to document these shortcuts in the docs, super handy.  Thank you!

thilobillerbeck:
just to be sure, should I leave these two "key pressed" methods separated or should I merge them?

gabek:
Good question!  I'd say probably merge them, so there's one place for people to look at this logic.  Thanks for asking!

gabek:
Thank you! 🥇

PR #1032, opened 2021-05-21
Fix #981 Use -webserverip to set http listen address

### LINKED ISSUE #981: Allow binding the web server port to a specific address
The best example is when using a reverse proxy, which means the server admin could allow connections to the actual server port (8080) only from localhost (while everyone else connects through the proxy itself), but includes people with multiple network interfaces who may want to have the web port listening to only one of the addresses they have.

leuc:
Followed the HTTP port code with an extra IP address validation in the router.

It works via command line flag, but not sure what is needed for configuration from the admin web UI.

Literally my first Go code :)


config/config.go
```diff
@@ -18,6 +18,9 @@ var VersionNumber = StaticVersionNumber
 // WebServerPort is the port for Owncast's webserver that is used for this execution of the service.
 var WebServerPort = 8080
 
+// Bind WebServer to this IP address. Be secure by default.
+var WebServerIP = "127.0.0.1"
```
gabek: While I totally understand why you'd want to default to the loopback address, most people running Owncast won't know about this flag, not to mention all the existing users. They'll hit their public interface and nothing will come back, and they won't know why.  So I suggest we listen on `0.0.0.0` by default so it will continue to work for everyone.

config/defaults.go
```diff
@@ -46,6 +47,7 @@ func GetDefaults() Defaults {
 		YPServer:  "https://directory.owncast.online",
 
 		WebServerPort:  8080,
+		WebServerIP:    "127.0.0.1",
```
gabek: Ditto.  I should probably de-duplicate these two sets default values, sorry about that.

gabek (commented):
Thank you!  Just one comment about the default.

leuc:
default listen address changed. 

Added admin REST API endpoint, with type validation. Only tested with curl, not against the actual admin UI.

This should validate the IP, so people can't brick the config. Not sure about the UI interaction in case of an error, so left out for now.

controllers/admin/config.go
```diff
@@ -293,6 +293,12 @@ func SetWebServerPort(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// don't panic when input is not a float
+	_, is_float := configValue.Value.(float64)
+	if !is_float {
+		return
+	}
```
gabek: Thanks for adding this check!  Instead of asserting the type once here to check it, and then another time to actually use it, could this be swaped around a bit  (excuse the bad code inside a comment box):

```
if port, ok := configValue.Value.(float64); ok {
...setHttpPort(port)
} else {
BadRequestHandler(w, "bad request or some message")
}
```

And the same can probably be done with the `SetWebServerIP` check/use as well.

gabek:
@leuc Would you mind commenting on the issue at https://github.com/owncast/owncast/issues/981 so I can make sure you get credit for closing it? :)  It won't let me assign somebody who hasn't been involved in the discussion haha.

controllers/admin/config.go
leuc: Gotcha, will do.

leuc: 71aca7e should have a better logic and also checks port range and validates given IP

gabek:
Thanks for working on this!  I'm not sure yet if we want this configurable in the admin or not (I'm kind of leaning to it being a cli flag only), so I'll probably pull out the APIs if that's the case, but I'll leave it for now so we have the option.  Thanks again!

PR #1038, opened 2021-05-22
Fix #1001 set RTMP port from cli

### LINKED ISSUE #1001: Provide command line option for setting the RTMP listening port
I am trying to install Owncast on a new server which is already set up to have nginx listen on port 1935, so I can't start the app in order to get into the config to change the RTMP port. I'd like to see a command line option to set the RTMP port that way I can start the service without shutting down nginx first.

leuc:
Using flag.Int instead of manual string conversion, seems cleaner? flag does validate the type and complains.


gabek:
Agreed!  Thanks for doing that.

PR #1039, opened 2021-05-22
Add -transcoderlog flag

### LINKED ISSUE #1040: Allow the customization of the logging path
Per conversation: https://owncast.rocket.chat/channel/general?msg=E9RuDzmBqDL7Qkcu3

tomleb:
As discussed, I am adding this flag to be able to specify where to save the transcoder logs.
I have kept the default to be the same as before (eg: `transcoder.log` relative to the pwd). 
This is my first dive into this project so feel free to comment on the approach, etc. I was not too sure if I was supposed to use the config.* as global variables, but that seemed to be how it is done.

EDIT: I left the `transcoder_*` file as is, not sure if you want me to use `config.TranscoderLogFilePath` there as well.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"path/filepath"
 	"fmt"
 )
 
@@ -8,6 +9,9 @@ import (
 
 // DatabaseFilePath is the path to the file ot be used as the global database for this run of the application.
 var DatabaseFilePath = "data/owncast.db"
+
+// LogDirectory is the path to various log files
+var LogDirectory = "."
 
 // EnableDebugFeatures will print additional data to help in debugging.
 var EnableDebugFeatures = false
@@ -37,4 +41,8 @@ func GetReleaseString() string {
 	var gitCommit = GitCommit
 
 	return fmt.Sprintf("Owncast v%s-%s (%s)", versionNumber, buildPlatform, gitCommit)
+}
+
+func GetTranscoderLogFilePath() string {
+	return filepath.Join(LogDirectory, "transcoder.log")
 }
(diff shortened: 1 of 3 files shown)


gabek:
Awesome, thank you!

Slight change, would you mind changing it so instead of specifying the full path of the transcoder log, we can specify the directory of _all_ logging?  Then a `./owncast --logingpath /opt/things/stuff` would end up with a transcoder.log of `/opt/things/stuff/transcoder.log`, and any future logging we can also build paths from that directory.

config/config.go
```diff
@@ -38,3 +42,7 @@ func GetReleaseString() string {
 
 	return fmt.Sprintf("Owncast v%s-%s (%s)", versionNumber, buildPlatform, gitCommit)
 }
+
+func GetTranscoderLogFilePath() string {
```
gabek, line 46: Sorry, one nit!  Could you remove Transcoder from this since it can be used for all logs, not just a single log?  `GetLogPath` or `GetLogDirectory` or something that makes sense.

tomleb: My idea was that LogDirectory would be shared for all logs, and then we have a function for each of the other logs but those use the LogDirectory variable.

eg:
If you wanted to add another type of log, you'd add the function like so:

```go
func GetMyOtherLogFilePath() string {
    return filepath.Join(LogDirectory, "my_other_log.log")
}
```

and then you can use that for the other logs. This way we avoid having a bunch of filepath.Join(LogDirectory, "my_other_log.log") scattered everywhere in the code base, you just need to edit that one function.

What do you think?

gabek: I get it, that works for me!  Thanks for clarifying!

gabek:
Thanks for working on this!

PR #1042, opened 2021-05-22
GitHub GO Lint Action: fix git reference

leuc:
the lint action didn't specify the correct git reference on pull requests.

simply removing the given reference **should** fix this. the underlying `actions/checkout@v2` should do the right thing for both push and pull request events. try, could not fully test it.


gabek:
Happy to give it a shot!  The action works fine generally, except with forks/pulls.  Thanks!

PR #1048, opened 2021-05-23
fix: avoid duplicate cache-control header

### LINKED ISSUE #1047: Cache-Control header is wrong when using S3
I just quickly set up this amazing project for some tests and noticed that the `cache-control` header is set wrong.

*Actual Result*
image

*Expected Result*
`cache-control: max-age=31557600`

nodomain:
Fixes #1047


gabek:
Awesome!  Thanks for catching this and fixing it!

PR #1054, opened 2021-05-25
Separate logo and offline stream image

### LINKED ISSUE #840: Being able to have custom Offline image
It would be nice to be able to have a custom png/jpg for a offline image placeholder.  If it has to be svg, then allow that separate selection to be made.

Thank  you,
Awesome product!

ForestJohnson:
Allow the user to specify a different image for their logo vs. the image that is displayed in lieu of the stream when the stream is offline.

The offline stream image will default to the logo image if its not set.

I changed probably more than the bare minimum but I just wanted to be sure it would work. owncast-admin PR is:  https://github.com/owncast/owncast-admin/pull/193/files

I did test this manually, as well as the changes I made to the owncast-admin and it seemed to work as intended as far as I can tell. But my test was not very rigorous


webroot/js/components/player.js
```diff
@@ -34,7 +34,7 @@ const VIDEO_OPTIONS = {
   sources: [VIDEO_SRC],
 };
 
-export const POSTER_DEFAULT = `/img/logo.png`;
```
ForestJohnson, line 37: I have no idea if this is right or what this is supposed to do

gabek:
So here's the problem.  If we add this as is we'll have two identical image upload components, right next to each other in the admin, with no real rhyme or reason why you'd use one over the other.  **I** know that it's optional, but it's still there, and people are going to wonder what the deal is.

One thing Owncast has going for it is people find it simple, so I think we really have to make sure that things that go into the most basic settings don't have more junk in it than we need, it already has more than I'd like.

So far only you and one other person has asked for this very specific feature.  And while I don't think it's _bad_ I do think it would add cognitive overhead for every person who visits the admin.  Right now there's "the logo" and that's it, they don't have to think too much about it.  Having "the logo" and "the other logo that's used sometimes" would need to be clearer.

So if we were going to put this in I'd want a redesign of how we manage images, instead of another upload component right next to the existing one.  And admittedly I'm not sure what that would be offhand.  Or maybe we could shove it into some kind of new "advanced settings" location somewhere.

ForestJohnson:
>  no real rhyme or reason why you'd use one over the other

fair enough, expecting people to be willing to read in order to use the app is truly is a bridge too far. I might sound sarcastic here but I make the same argument you are making about usability all the time. Its a slippery slope towards looking like Blender's UI or something.  Or Jenkins CI.

gabek:
I do think there's a place we can find to put it, I just haven't thought much about it.  I'll chat with @ginger about somewhere where optional/secondary things can go.

PR #1056, opened 2021-05-26
added emoji focus check to prevent triggering shortcuts

### LINKED ISSUE #1025: Typing in the emoji search box triggers player keyboard shortcuts
For example, if you open the Emoji search box and you type something like "money" the "m" will mute the player.

thilobillerbeck:
This should fix #1025. Just if you are curious, the shortcut method checks each element by its id, since not all inputs share the input tag. Maybe we could shorten this statement and prevent issues with shortcuts by using the input tag for every input in the frontend. This should also help improving accessibility.

### DIFF
--- a/webroot/js/app.js
+++ b/webroot/js/app.js
@@ -430,6 +430,7 @@ export default class App extends Component {
     } else if (
       e.target !== document.getElementById('message-input') &&
       e.target !== document.getElementById('username-change-input') &&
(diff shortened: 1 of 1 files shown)


gabek:
Thanks for fixing this!  Next time the the UI gets an update (and likely swap out that emoji picker) we'll have to keep all that in mind.

PR #1070, opened 2021-05-31
Fix rtmp secret validation to allow `/` (#1069)

### LINKED ISSUE #1069: "/" in stream key breaks OBS streams
# Steps to reproduce

* Change Owncast stream key to ZPHwVFJglIHz
* Try streaming with OBS -> should succeed
* Change Owncast stream key to ZPHwV/FJglIHz
* Try streaming with OBS - fails (Owncast log says invalid streaming key; rejecting incoming stream)

# Expected behaviour

* OBS is able to stream with stream Key that includes a slash
* **OR** Owncast puts out a warning, that OBS does not support stream keys containing "/"

This was tested with Owncast v0.0.7 and OBS studio 26.1.2-2

jeyemwey:
Now allowing `/` in the stream secret.

This does rely on the middle part being `/live/` which was not enforced before. All of our docs use `/live/` in the examples, but there was nothing that stopped users to set this to other values.

Tangentely related to #1069 which reported that they're unable to use `/` secrets in OBS.

### DIFF
--- a/core/rtmp/rtmp.go
+++ b/core/rtmp/rtmp.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"net"
 	"os"
-	"strings"
 	"syscall"
 	"time"
 
@@ -78,9 +77,7 @@ func HandleConn(c *rtmp.Conn, nc net.Conn) {
 		return
 	}
 
-	streamingKeyComponents := strings.Split(c.URL.Path, "/")
-	streamingKey := streamingKeyComponents[len(streamingKeyComponents)-1]
-	if streamingKey != data.GetStreamKey() {
+	if !secretMatch(data.GetStreamKey(), c.URL.Path) {
 		log.Errorln("invalid streaming key; rejecting incoming stream")
 		nc.Close()
 		return
--- a/core/rtmp/utils.go
+++ b/core/rtmp/utils.go
@@ -78,16 +78,21 @@ func getVideoCodec(codec interface{}) string {
 }
 
 func secretMatch(configStreamKey string, url string) bool {
-  sep := "/live/"
-  givenParts := strings.Split(url, sep)
+	sep := "/live/"
+	givenParts := strings.Split(url, sep)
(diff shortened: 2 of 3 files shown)


aaronpk (changes requested):
Can you add some more failure tests, things like

* `rtmp://example.com/one/two/live/three`
* `rtmp://example.com/streamkey/live`
* `rtmp://example.com/somethingelse/streamkey`
*

jeyemwey:
:heavy_check_mark:

Added some more tests and added a test if there's stuff between the hostname and `/live/`.

jeyemwey:
The integration tests on 85601a1 showed a massive flaw in my function, namely the schema and the domain were only transmitted to the function in my tests, because `c.URL.Path` [is not what I thought it was](https://pkg.go.dev/net/url#URL.Path).

Anyways, the code is much clearer now and all the tests are happy, too.

gabek:
Could somebody catch me up?  I'm mis-interpreting this issue.  The original issue said that Owncast was fine, OBS doesn't support slashes so we should put something in the admin to let people know not to use them.   The reason I closed it was it was said there was not a technical problem, but there is?  Sorry if I'm not following what's going on here.

aaronpk:
The problem is that OwnCast is currently incorrectly parsing the RTMP URL when validating the stream key. This problem is not anything to do with OBS.

If the stream key in owncast is set to `one/two/three`, the client would send a URL like `rtmp://example.com/live/one/two/three` and the current code at line 81 (splitting on the slashes) would be grabbing just `three` and comparing it to `one/two/three` and reject the stream.

Another separate problem is OwnCast is essentially ignoring the first part of the RTMP path, `live`, so it will accept an RTMP URL with anything in that segment.

This PR fixes both issues by parsing the RTMP URL better when checking for the stream key.

gabek:
Thanks for fixing this!  Sorry for my mis-reading of the original issue.

PR #1083, opened 2021-06-07
inject custom style into app-video-only.js

MFTabriz:
This patch injects the user's custom style from the configs into the app-video-only.js


gabek:
Thank you!

PR #1091, opened 2021-06-10
Hide viewer count when offline

### LINKED ISSUE #1088: Don't show the viewer count in the embed player when offline
Just like the non-embed UI, we shouldn't show the viewer count if the stream is offline.

For anybody looking to work on this, it'll be in `app-video-only.js`.

image

MFTabriz:
closes (#1088)


MFTabriz:
I tried to restart the tests but golangci-lint is still failing by timeout.

gabek:
Thank you!

Issue #1082, opened 2021-06-07
Cleanup paths and make them configurable


ctsde-markus:
In the process of building the FreeBSD port for owncast an experienced commiter brought up some security considerations regarding writable files/directories to me.

The hier manpage of FreeBSD makes it clear that database files and variable data in general should go to /var/... (that makes it clear what to backup) and this consideration is reasonable to me.

I then tried to make a simple patch for the `transcoder.log` file only to find out that i would not only patch a config file, but parts of the source itself. So i tried to do that for `data` - which is defined in `config/constants.go` but already in `config/defaults.go` the `DatabaseFilePath` isn't built from the `DataDirectory` constant, but is defined as `data/backup.db`.

It would make porting owncast way easier if i could rely on `config/constants.go` and `config/defaults.go` to define paths for files and directories that need to be writable by the user running owncast.

What needs to be configurable from my perspective:

- backup
- data
- hls
- transcoder.log

While we are at it, we should collect the writeable files in a subdirectory in `webroot` like `webroot/local` (or something else)

- thumbnail.jpg
- (anything else?)

When this is done we could easy move `backup`, `data` and `hls` to `/var/db/owncast`. `transcoder.log` to `/var/log/owncast/transcoder.log` (or somewhere else). And have only `webroot/local` writeable for the user running owncast.

gabek:
Some other requests for customization have come through lately, resulting in a `-logdir` from: https://github.com/owncast/owncast/issues/1040.  

In general, as long as additional config options don't show up in the admin, adding command line flags isn't the worst thing in the world if there's valid use cases.  So data and backup could fall into that.

I'd like to simplify where hls segments are stored, since right now it depends on if you're using remote storage or not where the segments live, so it it makes sense to change that: https://github.com/owncast/owncast/issues/875.  Early on we had a flag for customizing the hls path but nobody used it, so it got pulled, so I figure we'd just move it under `data` since that's where user-writable data lives.

ctsde-markus:
For me it would be cool enough if i could rely on patching paths in config/*.go files to have the files where i (or the used UNIX) would like to see it. But command line switches are fine too.

fufexan:
What's the status on this? I'd also want to have Owncast write to `/var/lib/owncast`. Currently it fails to start because the binary is in a read-only location.

gabek:
I feel like this issue is a bit inactionable at this point, since the items listed above are a mix of directories, single files, things that are already configurable, etc.

[...]

PR #1099, opened 2021-06-12
updates to backups directory

### LINKED ISSUE #1093: Some small updates to backups
1. Change the default backup directory to be data directory + `backup`.  So by default `data/backup`.
1. Allow changing this directory via a command line flag (#1082)
1. Make sure this directory is created in `backup.go` when the backup is run and the directory doesn't exist.

MFTabriz:
* Changes the default backup directory to be data directory + backup.
* Creates the backup directory when the backup is run and the directory doesn't exist.
* Allows changing backup directory via a command line flag: backupdir (#1082)

closes #1093

### DIFF
--- a/core/data/data.go
+++ b/core/data/data.go
@@ -8,10 +8,10 @@ import (
 	"database/sql"
 	"fmt"
 	"os"
-  "path/filepath"
+ 	"path/filepath"
 	"time"
   
-  "github.com/owncast/owncast/config"
+	"github.com/owncast/owncast/config"
 	"github.com/owncast/owncast/utils"
 	log "github.com/sirupsen/logrus"
 )
@@ -96,7 +96,7 @@ func SetupPersistence(file string) error {
 
 	dbBackupTicker := time.NewTicker(1 * time.Hour)
 	go func() {
-                backupFile := filepath.Join(config.BackupDirectory, "owncastdb.bak")
+		backupFile := filepath.Join(config.BackupDirectory, "owncastdb.bak")
 		for range dbBackupTicker.C {
 			utils.Backup(_db, backupFile)
 		}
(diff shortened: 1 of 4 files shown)


utils/backup.go
```diff
@@ -59,6 +60,15 @@ func Restore(backupFile string, databaseFile string) error {
 func Backup(db *sql.DB, backupFile string) {
 	log.Traceln("Backing up database to", backupFile)
 
+  BackupDirectory := filepath.Dir(backupFile)
+  _, err := os.Stat(BackupDirectory)
```
gabek: There's a little utility function you can use so you don't have to do the extra Stat.
https://github.com/owncast/owncast/blob/develop/utils/utils.go#L29

MFTabriz: @gabek I used the DoesFileExists() and also changed the permission on the created folder to 700.

core/data/data.go
```diff
@@ -8,15 +8,19 @@ import (
...
 	"github.com/owncast/owncast/utils"
 	log "github.com/sirupsen/logrus"
 )
 
 const (
 	schemaVersion = 0
-	backupFile    = "backup/owncastdb.bak"
+)
+var (
+	backupFile    = filepath.Join(config.BackupDirectory, "owncastdb.bak")
```
gabek: Probably best just remove this since the `config.BackupDirectory` it's using is actually the default, so the customized directory isn't being used if you specify it on the command line.  Can just generate the full path when needed.
```
go run main.go --enableVerboseLogging -backupdir /tmp/testbackup
...
TRAC[2021-06-12T15:23:54-07:00] Backing up database to data/backup/owncastdb.bak
```

MFTabriz: @gabek Sorry about that! I hope the new commit fixes it.

core/data/data.go
```diff
@@ -95,6 +96,7 @@ func SetupPersistence(file string) error {
 
 	dbBackupTicker := time.NewTicker(1 * time.Hour)
 	go func() {
+                backupFile := filepath.Join(config.BackupDirectory, "owncastdb.bak")
```
gabek: Some weird spacing here.  If you happen to use VSCode, if you add the Go Language support it'll auto-format your code for you with [gofmt](https://golang.org/cmd/gofmt/).

MFTabriz: @gabek I should not use the GH editor for these kind of changes. That was a tab vs. space issue. I also removed the trailing spaces. The changes should also fix the linter errors.

gabek:
Thank you for your help! :1st_place_medal:

MFTabriz:
Sorry for the trouble and thanks for the checks.

PR #1118, opened 2021-06-18
Consolidate Dockerfile into root and consolidate version into config

### LINKED ISSUE #1041: Consolidate the two Dockerfiles
We have two Dockerfiles:

1. For running from source within a Docker container https://github.com/owncast/owncast/blob/develop/Dockerfile
2. For building and releasing a production Docker image.  It sets version numbers and git hashes and stuff when building. https://github.com/owncast/owncast/blob/develop/build/release/Dockerfile-build

This has been working fine but confuses everybody.  If anybody can spend some time consolidating these two that'd be awesome.

@geekgonecrazy Have any free cycles? :)

geekgonecrazy:
Consolidate Dockerfile to a single location at the root of the project.

While I was there also consolidated version string into config and adjusted the build args


gabek:
Thank you!

PR #1124, opened 2021-06-19
Remove empty action buttons container with extra space

### LINKED ISSUE #1119: Web frontend external actions container div creates blank space
`#external-actions-container` has a margin of `1em` even when there are no external action buttons, creating an empty spot in the UI.  Maybe we can not render that div at all if it's empty.

image

jyggiz:
### Description
Removed the unnecessary action buttons container that creates extra spacing when actions is empty. 

### Test
You need to go to the main page and see if the large space between the heading and the section with social networks has disappeared


gabek:
Awesome, thank you so much!  Looks much better :)

PR #1125, opened 2021-06-20
display last online time

### LINKED ISSUE #1111: Offline Web UI: Display when the last time there was a stream
Sometimes I'll visit an instance and I'll be curious when the last time they streamed was.  It would make sense to find a place to put this in the web UI somewhere.  There's already a `lastConnectTime` in the `/api/status` response we could use for this.

I'd like @gingervitis's opinion of where this could go and what it should look like before anybody actually makes the change.

gingervitis:
- address #1111 

- clean up status bar styles


gabek:
Thank you!!!

PR #1148, opened 2021-07-01
refactor: use io.Pipe and ffmpeg's stdin

### LINKED ISSUE #390: Fails to run on Windows
Trying to build the package for Windows currently fails. I'm seeing this both when running directly in a Windows command prompt, or cross-compiling on Linux with `GOOS=windows`.

My Windows env, FWIW:

```
PS C:\Users\Matt\code\owncast> go version
go version go1.15.5 windows/amd64
```

It fails to compile with this error:

```
# github.com/owncast/owncast/core/rtmp
core\rtmp\rtmp.go:94:10: undefined: syscall.Mkfifo
```

Cursory searching makes it look like this method just isn't supported on Windows: [1](https://gitlab.com/gitlab-org/gitlab-runner/-/merge_requests/2241) [2](https://github.com/containerd/fifo/issues/5)

I'm not really sure an alternative implementation. The above link mentioned https://github.com/Microsoft/go-winio might have useful Windows-specific utilities, so that might be a good place to start.

Also a reminder to myself, Windows folks almost certainly won't have a GCC toolchain installed, so will need to follow the requirements for the `sqlite3` package https://pkg.go.dev/github.com/mattn/go-sqlite3#readme-windows

Also CCing @bitspill as I think you originally reported this :)

jeyemwey:
This removes the usage of `syscall.Mkfifo` which was previously used and won't work on Windows systems and opens the door for other processes on the computer to interfere in the rtmp stream (dumping bad content in the fifo, removing the file, blocking the file in offline status).

Instead, this patch introduces an `io.Pipe` which pipes the RTMP stream to the ffmpeg command while staying in Owncast.

Further links:
* ffmpeg on using `pipe:0` (stdin) as an input: https://ffmpeg.org/ffmpeg-protocols.html#pipe

While I've tried my best to test the patch (streaming "works", also in Connect - Disconnect - Connect scenarios), please give it a try, too.

### DIFF
--- a/core/rtmp/rtmp.go
+++ b/core/rtmp/rtmp.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"io"
 	"net"
-	"os"
-	"syscall"
 	"time"
 
 	"github.com/nareix/joy5/format/flv"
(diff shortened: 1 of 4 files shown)


gabek:
Nice!  Thanks for taking care of this!  It's been on the todo list for a while.

I like that it's pretty much a one for one swap to using `io.Pipe` so not a lot needed to be changed.

It's working just fine for me.  I'll merge this in and get the next nightly build running on the demo server for some 24/7 testing.

PR #1149, opened 2021-07-01
🐛 use currentBroadcast for playlist overwrites

### LINKED ISSUE #1147: Don't remove non-existing files if video config changed while online
I have added a new output variant via the admin while the stream was live.

After the stream disconnected, Owncast tried to clean-up a track playlist that isn't there yet:

```log
INFO[2021-07-01T21:46:19+02:00] Video transcoder started using x264 with 1 stream variants. 
INFO[2021-07-01T21:50:35+02:00] Inbound stream disconnected.                 
WARN[2021-07-01T21:50:35+02:00] open hls/1/offline.ts: no such file or directory 
WARN[2021-07-01T21:50:35+02:00] open hls/1/offline.ts: no such file or directory 
ERRO[2021-07-01T21:50:35+02:00] open hls/1/stream.m3u8: no such file or directory 
ERRO[2021-07-01T21:50:35+02:00] invalid argument                             
WARN[2021-07-01T21:50:35+02:00] open hls/1/stream.m3u8: no such file or directory 
INFO[2021-07-01T21:50:40+02:00] Inbound stream connected.                    
INFO[2021-07-01T21:50:40+02:00] Video transcoder started using x264 with 2 stream variants. 
```

Previously, my config only had one track (in `hls/0/stream.m3u8`), but since the config has changed while I was online (now having two tracks), Owncast guesses that there must also be another track (in `hls/1/stream.m3u8`).

**To do:** Look into the clean up logic and don't try to delete files that don't exist.

jeyemwey:
Fixes #1147 since it looks into currentBroadcast and does not use the configuration of the next stream.

### DIFF
--- a/core/streamState.go
+++ b/core/streamState.go
@@ -88,7 +88,7 @@ func SetStreamAsDisconnected() {
 		_yp.Stop()
 	}
 
-	for index := range data.GetStreamOutputVariants() {
+	for index := range _currentBroadcast.OutputSettings {
 		playlistFilePath := fmt.Sprintf(filepath.Join(config.PrivateHLSStoragePath, "%d/stream.m3u8"), index)
 		segmentFilePath := fmt.Sprintf(filepath.Join(config.PrivateHLSStoragePath, "%d/%s"), index, offlineFilename)
 


gabek:
Nice, thank you!

PR #1150, opened 2021-07-01
fix video variant sorting

### LINKED ISSUE #1132: The sort order of player video qualities is buggy
It should show as high, medium, low.  
image

The sort in `controllers/video.go` probably isn't correct.

MFTabriz:
closes #1132


gabek:
Awesome, thank you!

PR #1163, opened 2021-07-05
Chat refactor + persistent backing chat users

gabek:
This PR includes a complete rewrite of the chat server with some frontend changes to accommodate it.  It's a lot, so I'd appreciate any help testing it.  Luckily it's pretty easy to test and doesn't take much time to play around with the changes.  See the testing steps below.

# Overview of major changes

- Chat server is completely rewritten.
- Turning off chat in the settings now completely removes the chat server functionality, not just hides it from the UI.
- You must now be a registered user to chat or get chat history.
- Registration takes place in the background and is not an interactive thing.
- Users can be banned from chat.
- Number of concurrent connections to chat is greatly increased.
- Old chat history is deleted so we don't keep more data than needed.
- See "Detailed notes" on the bottom of this PR for more info.

# Things to keep in mind

- When you are blocked from chat, while technically you can get unblocked, a boolean is set in your localstorage that will continue to stop you from reconnecting in the frontend. I didn't want to add more complexity and APIs around a chat client having to check if its been unblocked.
- You could wipe your localstorage, open a different browser, or open an incognito window to get around being banned.  We do not currently store IP addresses.  Originally I was going to, but I really didn't want to store that identifying information.  If in the future we need to add this additional layer of blocking we can, but users could always use VPNs or Tor to get around that, too.
- You cannot access `/api/chat` without an access token, just to make it a little bit more inconvenient if somebody wanted to harvest chat history.  [3rd parties have hopefully seen the notice around this](https://owncast.online/news/20210615-apis-future-releases/).

# Testing

The following sections walks you through some specific items to test.

## To begin

1. Since this test includes blocking you from chat you might want to test in an incognito window so you don't have to deal with blocked state later.
1. Run from the **`develop` branch**.
1. Add a test access token in the admin from the **`develop` branch**.
1. **Change to this branch** and run the new branch with the bundled admin: `go run main.go pkged.go`.
1. Make sure there are no errors.
1. Start a stream.

## Chat

1. Verify you successfully connect to chat.
1. Send some chat messages.
1. Make sure your username is the same as it was in 0.0.7.
1. Verify you can change your name to something.
1. Try changing your name to `owncast` and verify it tells you that is disallowed.
1. Reload the page and make sure you can still chat and your user is the same.

## Chat admin

1. Verify you show up as a connected user under the chat -> users page.
1. Block your chat user from this page.
1. Verify you get disconnected from chat and it gets hidden.
1. Refresh the page. Verify chat is still not accessible.
1. Open the page in a new incognito window. Verify the chat messages from your blocked user are hidden.

## Access tokens

- In the admin make sure the test access token you created above still shows up. 

----
# Detailed notes

- A user access token is required to connect to the websocket or to hit the chat history API.
- When a user first visits the Owncast page they register for an access token and it's saved to browser local storage.
- If they already had a username set from a previous visit to Owncast that name will be passed when registering and that name will continue to be used.
- Default usernames are no longer `userxx`, but instead a server-generated name, Docker style, like `excellent-pizza`.
- External/3rd party access tokens have been converted to be treated the same as any other chat user, except they're meant to be used pragmatically.  So you can think of them as "API users" .
- 3rd party integrations can take advantage of a new API where they can send chat messages as themselves. This makes sense for bots.
- The chat server now increases its max connection count to 70% of the process max connection count.  Most people were limited at 1024 concurrent connections before and now they're not.  Max still depends on how their operating system is configured.
- User colors are persistent between user name changes.
- You can specify a list of names that are prohibited from being used.  Defaults are: "owncast", "operator", "admin", "system"
- User's name history and user creation time is tracked to help validity of users.
- You can block users from accessing the chat via the admin.
- When a user is blocked all of their messages are marked as hidden and they are forcefully disconnected, and chat removed from their UI.
- Old chat messages are completely deleted from the database in order to reduce the amount of history we have for privacy and data exposure reasons.
- Added a simple chat load test that will connect 5000 users and have them send messages. This replaces the old artillery-based one since that was too limited.
- Websocket has been moved to a more obvious `/ws` path.


gingervitis:
Going to this page `/admin/access-tokens/` right after switching over to this branch gives an error.
image

gingervitis:
Changing your name to `"owncast"` - new name does not go away after error message is returned.

image

The forbidden "owncast" name remains even after reloading the page.

However, chatting will display your old user name.
image

gingervitis:
Some funkiness:

I have 2 different browsers going. "User 301" and "User 11".

1. From the Admin I banned "User 301".
2. Go to User 301's page. 301's chat clears out. But the panel is still there, and can still input and send messages. But this JS error does show up for it, post-ban.
image

3. Go to User 11's page. 11's chat panel has disappeared. (and it looks not-great)
4. Go reload the Admin chat user's page. 301 shows in both the Users table and the Blocked table. User 11 has disappeared.
image


Other funkiness: 
I noticed that in the Admin users table, the ordering of User 301 and User 11 would keep switching in the table. There might be something going on there, which may have confused which user actually got banned.

gingervitis:
This happens when trying to auto complete names (does not happen in `develop`)
image

gingervitis:
I restarted a new test with 2 browsers again , and did not come across these errors the 2nd time.

https://github.com/owncast/owncast/pull/1163#issuecomment-877687381

build/admin/bundleAdmin.sh
```diff
@@ -18,6 +18,7 @@ trap shutdown INT TERM ABRT EXIT
 echo "Cloning owncast admin into $INSTALL_TEMP_DIRECTORY..."
 git clone https://github.com/owncast/owncast-admin 2> /dev/null
 cd owncast-admin
+git checkout gek/chat-user-refactor
```
jeyemwey, line 21: Remove when merging into `develop`?

controllers/admin/serverConfig.go
```diff
@@ -15,6 +16,7 @@ import (
 // GetServerConfig gets the config details of the server.
 func GetServerConfig(w http.ResponseWriter, r *http.Request) {
 	ffmpeg := utils.ValidatedFfmpegPath(data.GetFfMpegPath())
+	usernameBlocklist := strings.Join(data.GetUsernameBlocklist(), ",")
```
jeyemwey: Why is `serverConfigAdminResponse.UsernameBlocklist` not a `[]string` but just a string? JSON offers a list encoding, and it usually makes sense to use it (if only for trimming spaces).

core/chat/events.go
```diff
@@ -0,0 +1,98 @@
...
+	oldName := savedUser.DisplayName
+
+	// Save the new name
+	user.ChangeUsername(eventData.client.User.Id, receivedEvent.NewName)
+
+	// Send chat event letting everyone about about the name change
+	savedUser.DisplayName = receivedEvent.NewName
+
+	// Update the connected clients associated user with the new name
+	eventData.client.User = savedUser
```
jeyemwey: Swap, so the comments stay correct:

```suggestion
	// Update the connected clients associated user with the new name
	eventData.client.User = savedUser

	// Send chat event letting everyone about about the name change
	savedUser.DisplayName = receivedEvent.NewName
```

core/chat/events/events.go
```diff
@@ -12,51 +12,72 @@ import (
...
+type EventPayload map[string]interface{}
+
+type OutboundEvent interface {
+	GetBroadcastPayload() EventPayload
+	GetMessageType() EventType
+}
+
+// Event is any kind of event.  A type is required to be specified.
+type Event struct {
+	Type      string    `json:"type"`
```
jeyemwey: Can we pin this to be of type `EventType`?

```suggestion
	Type      EventType `json:"type"`
```

core/chat/events/nameChangeEvent.go
```diff
@@ -0,0 +1,26 @@
...
+	UserEvent
+	Oldname string `json:"oldName"`
+}
+
+// GetBroadcastPayload will return the object to send to all chat users.
+func (e *NameChangeBroadcast) GetBroadcastPayload() EventPayload {
+	return EventPayload{
+		"id":        e.Id,
+		"timestamp": e.Timestamp,
+		"user":      e.User,
```
jeyemwey, line 22: Looking at the other events, this looks like a lot of duplicated code. Can we create a method which creates an `EventPayload` from any `e Event`, and add items to the `map[string]interface{}` later on?

core/chat/events/systemMessageEvent.go
```diff
@@ -0,0 +1,26 @@
...
+	Event
+	MessageEvent
+}
+
+// SystemMessageEvent will return the object to send to all chat users.
+func (e *SystemMessageEvent) GetBroadcastPayload() EventPayload {
+	return EventPayload{
+		"id":        e.Id,
+		"timestamp": e.Timestamp,
+		"body":      e.Body,
```
jeyemwey, line 16: See above.

core/chat/events/userDisabledEvent.go
```diff
@@ -0,0 +1,17 @@
...
+	UserEvent
+}
+
+// GetBroadcastPayload will return the object to send to all chat users.
+func (e *UserDisabledEvent) GetBroadcastPayload() EventPayload {
+	return EventPayload{
+		"type":      ErrorUserDisabled,
+		"id":        e.Id,
+		"timestamp": e.Timestamp,
+		"user":      e.User,
```
jeyemwey, line 15: See above.

core/chat/events/userJoinedEvent.go
```diff
@@ -0,0 +1,17 @@
...
+	UserEvent
+}
+
+// GetBroadcastPayload will return the object to send to all chat users.
+func (e *UserJoinedEvent) GetBroadcastPayload() EventPayload {
+	return EventPayload{
+		"type":      UserJoined,
+		"id":        e.Id,
+		"timestamp": e.Timestamp,
+		"user":      e.User,
```
jeyemwey, line 15: See above.

core/chat/events/userMessageEvent.go
```diff
@@ -0,0 +1,24 @@
...
+	MessageEvent
+}
+
+// GetBroadcastPayload will return the object to send to all chat users.
+func (e *UserMessageEvent) GetBroadcastPayload() EventPayload {
+	return EventPayload{
+		"id":        e.Id,
+		"timestamp": e.Timestamp,
+		"body":      e.Body,
+		"user":      e.User,
```
jeyemwey, line 16: See above (except for the body).

models/webhook.go
```diff
@@ -25,7 +29,8 @@ var validEvents = []EventType{
 // This is not a efficient method.
 func HasValidEvents(events []EventType) bool {
 	for _, event := range events {
-		if !findItemInSlice(validEvents, event) {
+		_, foundInSlice := utils.FindInSlice(validEvents, event)
+		if !foundInSlice {
```
jeyemwey:
```suggestion
		if _, foundInSlice := utils.FindInSlice(validEvents, event); !foundInSlice {
```

test/automated/chat.test.js
```diff
@@ -2,52 +2,43 @@ const { test } = require('@jest/globals');
...
 
-const testMessageId = Math.random().toString(36).substring(7);
-const username = 'user' + Math.floor(Math.random() * 100);
+var userDisplayName;
 const message = Math.floor(Math.random() * 100) + ' test 123';
-const messageRaw = message + ' *and some markdown too*';
-const messageMarkdown = '<p>' + message + ' <em>and some markdown too</em></p>'
-const date = new Date().toISOString();
+// const messageRaw = message + ' *and some markdown too*';
+// const messageMarkdown = '<p>' + message + ' <em>and some markdown too</em></p>'
```
jeyemwey: Remove code dumps

```suggestion
```

test/automated/chat.test.js
```diff
@@ -2,52 +2,43 @@ const { test } = require('@jest/globals');
...
+  request
+    .get('/api/admin/chat/messages')
+    .auth('admin', 'abc123')
+    .expect(200)
+    .then((res) => {
+      const message = res.body.filter(function (msg) {
+        return (msg.body = testMessage.body);
+      })[0];
+
+      expect(message.body).toBe(message.body);
```
jeyemwey: Will always be true. You probably want to diff against the `testMessage.body`?

test/automated/chat.test.js
```diff
@@ -2,52 +2,43 @@ const { test } = require('@jest/globals');
...
-
-            done();
-        });
+  request
+    .get('/api/admin/chat/messages')
+    .auth('admin', 'abc123')
+    .expect(200)
+    .then((res) => {
+      const message = res.body.filter(function (msg) {
+        return (msg.body = testMessage.body);
```
jeyemwey: Missing equal sign, thus not filtering properly.

test/automated/chat.test.js
```diff
@@ -2,52 +2,43 @@ const { test } = require('@jest/globals');
...
+      })[0];
+
+      expect(message.body).toBe(message.body);
+      expect(message.user.displayName).toBe(userDisplayName);
+      // expect(message.date).toBe(testMessage.date);
```
jeyemwey: Remove commented code.
```suggestion
```

test/automated/chat.test.js
```diff
@@ -2,52 +2,43 @@ const { test } = require('@jest/globals');
...
-    origin: 'http://localhost',
-});
 
-    function onOpen() {
-        ws.send(JSON.stringify(testMessage), function() {
-            ws.close();
-            done();
-        });
-    }
+test('can send a chat message', async (done) => {
```
jeyemwey, line 16: The `can send a chat message` and `can fetch a chat message` test need to work together to work properly (which you can smell from the global `var userDisplayName`. IMO, it would be better to wrap both tests in a `describe` block, so it's obvious that they work with the same data.

test/automated/chatmoderation.test.js
```diff
@@ -2,38 +2,30 @@ const { test } = require('@jest/globals');
...
-    }
-
-    ws.on('open', onOpen);
-});
-
-var messageId;
+test('can send a chat message', async (done) => {
+    const registration = await registerChat();
+    const accessToken = registration.accessToken;
+    userDisplayName = registration.displayName;
```
jeyemwey: `userDisplayName` is not declared for this test, is never used later. I'd vote to remove it from the test.

test/automated/chatusers.test.js
```diff
@@ -0,0 +1,73 @@
...
+};
+
+var userId
+var accessToken
+test('can register a user', async (done) => {
+    const registration = await registerChat();
+    userId = registration.id;
+    accessToken = registration.accessToken;
+    done();
+});
```
jeyemwey, line 20: Same as with `test/automated/chat.test.js`, the tests are not independent and need to run in a specific order. It would be more obvious what is happening where if we were using `describe` or `beforeEach`/`beforeAll`.

test/load/chatLoadTest.js
```diff
@@ -0,0 +1,78 @@
...
+  try {
+      const response = await fetch('http://localhost:8080/api/chat/register', options);
+      const result = await response.json();
+      return result;
+  } catch(e) {
+      console.error(e);
+  }
+}
+
+async function sendMessage() {
```
jeyemwey: From a review POV, it's not clear what is happening when. Is this function only running once? Is it calling itself (with a timeout)? What is the `send` function doing inside of it?

utils/phraseGenerator.go
```diff
@@ -0,0 +1,636 @@
+package utils
```
jeyemwey, line 1: This is the Docker list, isn't it? Is there a reason why we can't import their package, but have to copy it over? Is that fine with Docker's (and our) IP Licensing?

jeyemwey (commented):
Hi Gabe, I've added a bunch of comments regarding the Go code and some testing bits. Mostly smaller things, so no worries.

I did not touch the `webroot/` folder, and most of the business logic, so it would be helpful if someone would look on that with a clear mind.

@ whoever shies away from the line count: The PR also includes 1,200 diff lines of Documentation, 6,300 diffs of a `package-lock.json` and 600 lines for a phrase generator. So it's "only" around 2,200 additions that need to be looked at :joy:

build/admin/bundleAdmin.sh
gabek: Yup!

utils/phraseGenerator.go
gabek: I wanted our own list we could customize.  No headers were stripped from the original.

core/chat/events/nameChangeEvent.go
gabek: I did this on purpose so the outbound payloads were explicit and obvious and are not bound by an abstraction, since these values are specifically for client consumption.  I like being able to look at a specific event and know exactly how it's being generated, and it makes it clear when adding more in the future.  If we wanted to type the outbound events (like we do with the internal objects) then we can abstract it more, but I chose this flexibility over having an additional specific `BroadcastNameChangeEvent` type, even though that would also be fine.

test/automated/chat.test.js
gabek: As far as I've seen tests have always run in order, so I've never seen any issues from it.  But I'm no expert in Jest.  Just for clarification I checked with @gingervitis and she believes they always run in order.  This could be updated in the future if we want to spend some time grouping them together.  But all the tests currently work this way, and they're not broken now and are performing as expected.

test/load/chatLoadTest.js
gabek: I renamed and added comments it to be clearer.  I could totally spend a bunch of time building a far cleaner script, but this just needs to do one specific thing to assist in testing, so it was thrown together for that because the load testing framework wouldn't allow me to do what I needed to do.  Hopefully we can get rid of it if Artillery's update comes out some day.

gingervitis:
`previousNames` for a user keeps coming back `null`

test/automated/chat.test.js
gabek: Great find, fixed!

controllers/admin/serverConfig.go
gabek: It's being stored as a string, so one side or the other needs to do the translation, so I just took the route of the least amount of code and allowed it to be sent as a string.  I went ahead and changed it to be an array over the wire, but the end result is the same, with the trimming of whitespace still happening [during the comparison](https://github.com/owncast/owncast/pull/1163/files#diff-249ac4ddbe2945013dfb90c7fc32f299ea8e9a3326e67033eb14b476323ff004R25-R28).

PR #1167, opened 2021-07-07
proper cleanup and permission check for the hls directory

### LINKED ISSUE #1166: Detect if the directories we expect to write to have bad permissions
Right now we fail silently if we can't write to the hls directories.  We should be nice and proactively check things like this at launch to alert on this issue.  This can probably done at the same time as https://github.com/owncast/owncast/issues/875.

See https://owncast.rocket.chat/channel/general?msg=d8i9CoqERqji3ejSt for context.

MFTabriz:
If the process doesn't have the right (write) permission for the hls directory:

```
INFO[2021-07-07T18:11:12Z] Owncast v0.0.8-dev (20210707)                
TRAC[2021-07-07T18:11:12Z] Creating webhooks table...                   
TRAC[2021-07-07T18:11:12Z] Creating access_tokens table...              
TRAC[2021-07-07T18:11:12Z] http_listen_address sql: no rows in result set 
TRAC[2021-07-07T18:11:12Z] Resetting file directories to a clean slate. 
FATA[2021-07-07T18:11:12Z] unlinkat ./hls/0/stream-offline***********.ts: permission denied 
exit status 1
```

closes #1166

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -114,28 +114,13 @@ func resetDirectories() {
 	log.Trace("Resetting file directories to a clean slate.")
 
 	// Wipe the public, web-accessible hls data directory
-	err := os.RemoveAll(config.PublicHLSStoragePath)
-	if err != nil {
-		log.Fatalln(err)
-	}
-	err = os.RemoveAll(config.PrivateHLSStoragePath)
-	if err != nil {
-		log.Fatalln(err)
-	}
-	err = os.MkdirAll(config.PublicHLSStoragePath, 0777)
-	if err != nil {
-		log.Fatalln(err)
-	}
-
-	err = os.MkdirAll(config.PrivateHLSStoragePath, 0777)
-	if err != nil {
-		log.Fatalln(err)
-	}
+	utils.CleanupDirectory(config.PublicHLSStoragePath)
+	utils.CleanupDirectory(config.PrivateHLSStoragePath)
 
 	// Remove the previous thumbnail
 	logo := data.GetLogoPath()
 	if utils.DoesFileExists(logo) {
-		err = utils.Copy(path.Join("data", logo), filepath.Join(config.WebRoot, "thumbnail.jpg"))
+		err := utils.Copy(path.Join("data", logo), filepath.Join(config.WebRoot, "thumbnail.jpg"))
 		if err != nil {
 			log.Warnln(err)
 		}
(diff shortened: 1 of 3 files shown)


core/core.go
```diff
@@ -114,9 +114,15 @@ func resetDirectories() {
 	log.Trace("Resetting file directories to a clean slate.")
 
 	// Wipe the public, web-accessible hls data directory
-	os.RemoveAll(config.PublicHLSStoragePath)
-	os.RemoveAll(config.PrivateHLSStoragePath)
-	err := os.MkdirAll(config.PublicHLSStoragePath, 0777)
+	err := os.RemoveAll(config.PublicHLSStoragePath)
+	if err != nil {
+		log.Fatalln(err)
+	}
```
gabek: Just to save a line (and I'm not always good at remembering to do this myself), you can write these error checks as

```
if err := os.RemoveAll(config.PublicHLSStoragePath); err != nil {
   ...
}
```

core/transcoder/utils.go
```diff
@@ -91,9 +91,19 @@ func handleTranscoderMessage(message string) {
...
 func createVariantDirectories() {
 	// Create private hls data dirs
+
+	err := os.RemoveAll(config.PublicHLSStoragePath)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	err = os.RemoveAll(config.PrivateHLSStoragePath)
+	if err != nil {
+		log.Fatalln(err)
```
gabek: Is it possible to fire `resetDirectories()` so the directory cleanup doesn't need to be duplicated?

MFTabriz: @gabek Moved the code to CleanupDirectory function in utils. What do you think about it?

core/core.go
MFTabriz: Thanks for pointing it out. I think this should be refactored once in the whole codebase to prevent people from copying the style in their contribution.

utils/utils.go
```diff
@@ -226,3 +226,14 @@ func VerifyFFMpegPath(path string) error {
...
 }
+
+// Removes the directory and makes it again. Throws fatal error on fail.
+func CleanupDirectory(path string) {
+	log.Traceln("Cleaning", path)
+	if err := os.RemoveAll(path); err != nil {
+		log.Fatalln(err)
+	}
+	if err := os.MkdirAll(path, 0777); err != nil {
+		log.Fatalln(err)
```
gabek: Just because we're adding these errors to help users and not developers, could you add a little human readable blurb?  Maybe just a `log.Fatalln("Unable to remove/create directory, check ownership and permissions", err)` or something.  Other than that I think we're good to go!

MFTabriz: @gabek Done!

PR #1170, opened 2021-07-09
style fix for error checking

MFTabriz:
some style fix for error checking (according to https://github.com/owncast/owncast/pull/1167#discussion_r665644437) and additional linting.


gabek:
Awesome!  Thanks for going through this!  Clean up is always good, I appreciate you taking the time.

PR #1173, opened 2021-07-09
Do not return true from DoesFileExists() on misc. errors

MFTabriz:
Just a small patch for better handling of os.Stat() errors.


gabek:
Nice tweak, thank you!

PR #1175, opened 2021-07-10
Cache the custom emoji list

MFTabriz:
We cache the emoji list and return it for the API calls. If the emoji directory’s modification time is not the same as the cache’s timestamp, we renew the cache.

PS: and the answer to [this comment](https://github.com/owncast/owncast/pull/108/commits/53b7bd77ad1ec5a3547cdc9e09893bfd6647f4c3#diff-db61f30213576cd7f03ff9fa695ce59656366854fa9b9dfa5e96e4c496857aadR31-R32) is: no! there is no need to restart the server after adding emojis :)

### DIFF
--- a/controllers/emoji.go
+++ b/controllers/emoji.go
@@ -22,6 +22,7 @@ const emojiDir = "/img/emoji" // Relative to webroot
 var emojiCache = make([]models.CustomEmoji, 0)
 var emojiCacheTimestamp time.Time
 
+// getCustomEmojiList returns a list of custom emoji either from the cache or from the emoji directory.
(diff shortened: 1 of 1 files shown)


gabek:
Seems totally reasonable, thanks for this!  I totally forgot about the fact that at some point this should be done.  That comment was set and forgotten, thanks for taking care of it!

MFTabriz:
@gabek marking these kind of comments with a tag such as TODO can be really useful. Later on, a todo list can be generated from them using the documentation tools e.g. godoc.

PR #1180, opened 2021-07-12
improve emoji cache timestamping

MFTabriz:
In the old version, if the content of the emoji directory was modified during the caching process, it was possible for the cache to get a wrong timestamp.


MFTabriz:
patch is not required! As long as FileInfo.Sys() is not used, the returned FileInfo will refer to the state of the directory when the Stat() was requested.

PR #1185, opened 2021-07-13
fix meta description, og:description & twitter:description blank tags

### LINKED ISSUE #1184: og:description tag is blank in HTML headers when sharing an Owncast link
It's not getting set [here](https://github.com/owncast/owncast/blob/73615784122bff635bb62d3fbcd00e5144fc9926/controllers/index.go#L104) therefore it's empty when being used [here](https://github.com/owncast/owncast/blob/develop/static/metadata.html#L13).

image

MFTabriz:
image

closes #1184


gabek:
Nice!  Thanks for taking care of that. :1st_place_medal:

Issue #922, opened 2021-04-13
Feature request: Automatic TLS via Let’s Encrypt


aral:
It would help deployments if TLS was automatically supported (currently the documentation states that a server like Caddy should be used for this but that’s quite a bit of additional complexity).

In [Site.js](https://sitejs.org), for example, https is automatically supported both on localhost (using mkcert via [auto-encrypt-localhost](https://github.com/small-tech/auto-encrypt-localhost)) and at hostname/public domains (using Let’s Encrypt via [auto-encrypt](https://github.com/small-tech/auto-encrypt)). 

While auto-encrypt can be used to whip up a tiny, lightweight proxy server for this purpose, there seems to be native support for what it does in go via the autocert library. And, it seems that there’s already a proxy forwarding server:

https://github.com/samthor/https-forward

I’d suggest integrating this functionality into Owncast to make secure deployments seamless.

aral:
Note: this is now implemented for Owncast in Site.js. Please feel free to steal and implement natively in Owncast :)

https://github.com/owncast/owncast/discussions/935

gabek:
Caddy released this that would be a good drop-in candidate to add this to Owncast.

https://github.com/caddyserver/certmagic

If we did add this:

1. What is the default with Owncast?  HTTP or HTTPS?
1. While it's easy to say "of course HTTPS by default" most Owncast (almost all) servers have their own HTTPS setup already. We wouldn't want to get in their way.
1. We may be able to attempt SSL by default and then fallback to HTTP if we can't bind to 443.

<blockquote><img src="https://repository-images.githubusercontent.com/161109499/c6b40280-8e29-11e9-9beb-8654a93b3a5d" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/caddyserver/certmagic">caddyserver/certmagic</a></strong></div><div>Automatic HTTPS for any Go program: fully-managed TLS certificate issuance and renewal - caddyserver/certmagic</div></blockquote>

Kunsi:
Please make this a configuration option. Users may want to use owncast on a standalone server, however others may want to run it only behind a reverse proxy, which then would handle SSL.

PR #1187, opened 2021-07-14
link assets under the same protocol for bots

MFTabriz:
When serving the page over a secure connection, the links should be also over the https.

### DIFF
--- a/controllers/index.go
+++ b/controllers/index.go
@@ -74,12 +74,17 @@ func IndexHandler(w http.ResponseWriter, r *http.Request) {
 // to give to Opengraph clients and web scrapers (bots, web crawlers, etc).
 func handleScraperMetadataPage(w http.ResponseWriter, r *http.Request) {
 	tmpl := template.Must(template.ParseFiles(path.Join("static", "metadata.html")))
+	protocol := "http"
 
-	fullURL, err := url.Parse(fmt.Sprintf("http://%s%s", r.Host, r.URL.Path))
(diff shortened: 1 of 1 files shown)


jeyemwey:
Question because I don't know Go good enough: Currently, [Owncast refers to a proxy server for TLS](https://owncast.online/docs/sslproxies/), so the proxy sends an individual (plain http) request to the local Owncast. It'd thus assume that `r.TLS` is always `nil` since OC never really get's in touch with any TLS?

MFTabriz:
@jeyemwey you are right. Totally forgot about that!

@gabek Are there any plans for the native TLS support in Owncast in future? Can't we just set relative addresses for the metadata [here](https://github.com/owncast/owncast/blob/bf0720aba73db7f557ffa9909d01a047497a72f3/controllers/index.go#L109)?

gabek:
There are no immediate plans, though there's [a discussion](https://github.com/owncast/owncast/issues/922).

Does the `//` trick work?

```
imageURL, err := url.Parse(fmt.Sprintf("//%s%s", r.Host, "/logo"))
```

MFTabriz:
@gabek This works fine in a browser but I'm not sure how real bots will interpret it.

gabek:
Let's give it a spin and find out! :)  Thanks for looking at this.  It'll get auto deployed to `nightly.owncast.tv` tomorrow and we can do some experimenting with search engines and social debuggers.

PR #1190, opened 2021-07-15
Support text input of :emoji: in chat

### LINKED ISSUE #480: Support text input of :emoji:, and show the title of emoji in title attribute

MFTabriz:
Replaces the typed `:emoji:` in the chat on client-side with the emoji ``. Only the valid emoji names get replaced [+bonus point!]. Emoji name lookup is case-insensitive. Supports pasting text with `:emoji:` in it (you can paste `This is a :portalparrot:!` and it will be rendered). Copy-pasting of the rendered emoji is also supported! [+extra bonus point! 😉]

closes #480

### DIFF
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -181,7 +181,7 @@ export default class ChatInput extends Component {
 
     if (textValue != convertToText(processedHTML)) {
       this.setState({
-        inputHTML: processedHTML
+        inputHTML: processedHTML,
       });
       return true;
     }
@@ -251,6 +251,7 @@ export default class ChatInput extends Component {
       }
     }
     this.setState({
+      inputText: textValue,
       inputCharsLeft: CHAT_MAX_MESSAGE_LENGTH - textValue.length,
     });
   }
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -193,8 +193,7 @@ export default class ChatInput extends Component {
       });
 
       if (emojiIndex != -1) {
-        const url = location.protocol + '//' + location.host + '/' + emojiList[emojiIndex].emoji;
-        const emojiImgElement = '<img class="emoji" alt="' + emojiList[emojiIndex].name + '" title="' + emojiList[emojiIndex].name + '" src="' + url + '"/>';
+        const emojiImgElement = '<img class="emoji" alt="' + emojiList[emojiIndex].name + '" title="' + emojiList[emojiIndex].name + '" src="' + emojiList[emojiIndex].emoji + '"/>';
 
         processedHTML = processedHTML.replace(":" + typedEmoji + ":", emojiImgElement)
         foundEmoji = true;
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -182,12 +182,12 @@ export default class ChatInput extends Component {
       return false;
     }
 
-    let typedEmoji = inputHTML.substring(at + 1, position).trim();
-    const emojiIndex = emojiList.findIndex(function (emojiItem) { return emojiItem.name === typedEmoji; });
+    const typedEmoji = textValue.substring(at + 1, position).trim();
+    const emojiIndex = emojiList.findIndex(function (emojiItem) { return emojiItem.name.toLowerCase() === typedEmoji.toLowerCase(); });
 
     if (emojiIndex != -1) {
       const url = location.protocol + '//' + location.host + '/' + emojiList[emojiIndex].emoji;
-      const emojiItem = '<img class="emoji" alt="' + typedEmoji + '" src="' + url + '"/>';
+      const emojiItem = '<img class="emoji" alt="' + emojiList[emojiIndex].name + '" src="' + url + '"/>';
 
       this.setState({
         inputHTML: inputHTML.replace(":" + typedEmoji, emojiItem)
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -180,12 +180,12 @@ export default class ChatInput extends Component {
     let processedHTML = inputHTML;
     for (var lastPos = textValue.length; lastPos >= 0; lastPos--) {
       const endPos = textValue.lastIndexOf(':', lastPos);
-      if (endPos === -1) {
+      if (endPos <= 0) {
         break;
       }
       const startPos = textValue.lastIndexOf(':', endPos - 1);
-      if (startPos === endPos) {
-        continue;
+      if (startPos === -1) {
+        break;
       }
       const typedEmoji = textValue.substring(startPos + 1, endPos).trim();
       const emojiIndex = emojiList.findIndex(function (emojiItem) {
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -8,6 +8,7 @@ import ContentEditable, { replaceCaret } from './content-editable.js';
 import {
   generatePlaceholderText,
   getCaretPosition,
+  getCaretCharacterOffsetWithin,
   convertToText,
   convertOnPaste,
   createEmojiMarkup,
@@ -176,7 +177,7 @@ export default class ChatInput extends Component {
   injectEmoji() {
     const { inputHTML, emojiList } = this.state;
     let textValue = this.formMessageInput.current.textContent;
-    const position = getCaretPosition(this.formMessageInput.current);
+    const position = getCaretCharacterOffsetWithin(this.formMessageInput.current);
     const at = textValue.lastIndexOf(':', position - 1);
     if (at === -1) {
       return false;
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -234,7 +234,7 @@ export default class ChatInput extends Component {
 
   handleMessageInputKeyup(event) {
     const formField = this.formMessageInput.current;
-    let textValue = formField.textContent; // get this only to count chars
+    const textValue = formField.textContent; // get this only to count chars
 
     const { key } = event;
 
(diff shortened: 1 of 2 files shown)


MFTabriz:
@gingervitis I have modified the code to replace all `:emoj:`s in the text with the image. Unfortunately, I couldn’t figure out how to trigger this functionality when someone pastes text with `:emoji:` in it. 
I'd really appreciate it if you can give me some pointers on how this should be done (and how to improve my js).

gingervitis:
Oh cool! I'll check it out this weekend and see what we can do for it.

webroot/js/components/chat/chat-input.js
```diff
@@ -171,6 +173,41 @@ export default class ChatInput extends Component {
...
+      if (startPos === -1) {
+        break;
+      }
+      const typedEmoji = textValue.substring(startPos + 1, endPos).trim();
+      const emojiIndex = emojiList.findIndex(function (emojiItem) {
+        return emojiItem.name.toLowerCase() === typedEmoji.toLowerCase();
+      });
+
+      if (emojiIndex != -1) {
+        const url = location.protocol + '//' + location.host + '/' + emojiList[emojiIndex].emoji;
```
gingervitis: We should just leave this  as `emojiList[emojiIndex].emoji;`. Those urls start with `/` and should automatically just come from self domain.

webroot/js/components/chat/chat-input.js
```diff
@@ -227,6 +264,12 @@ export default class ChatInput extends Component {
     if (CHAT_KEY_MODIFIERS.includes(key)) {
       this.modifierKeyPressed = false;
     }
+    if (key === ':' || key === ';') {
```
gingervitis, line 247: Why `;`?

gingervitis:
@MFTabriz Ok some thoughts about the paste functionality..  So I'm guessing that the question is when you want to paste in something like `I love :partyparrot: and :bananadance: a lot!` , and you want it to show up like
image?


I would make the `injectEmoji` function an external util instead in `chat.js`.  Then somewhere in the `convertOnPaste` function, you could do the parsing in there to return the emoji image there?

gingervitis:
Interesting, Safari in Incognito mode does not inject the emojis 🤔

webroot/js/components/chat/chat-input.js
MFTabriz: if the user releases the shift key first, the `;` keyup event will trigger instead of `:`

webroot/js/components/chat/chat-input.js
MFTabriz: done!

gingervitis:
Hmm interesting, the `title` and `alt` attributes aren't being rendered.  I feel like React's `dangerouslySetInnerHTML` method is stripping some attributes.. let me dig a little more into that. 🤔

gingervitis:
> Hmm interesting, the `title` and `alt` attributes aren't being rendered. I feel like React's `dangerouslySetInnerHTML` method is stripping some attributes.. let me dig a little more into that. 🤔

Ok actually it looks like the server is stripping those attributes when it returns the message to be displayed. I'll open a ticket.

webroot/js/components/chat/chat-input.js
gingervitis: OH! Wow, I guess my right hand has always been faster to release than my left hand and this never occurred to me!

gingervitis:
@MFTabriz I made a quick update to the `gek/chat-rebuild-with-access-tokens` that adds a util to generate the emoji `` markup (since it looks like we'll now be doing it in at least a couple of places).  Feel free to use it!

https://github.com/owncast/owncast/commit/e6abee927b08d8f7424710bcd04d583be06dbece#diff-91877633ef1d5fe9e1c1c5b03ed30966b09faac26799cdaf840fe2aac11ff322R145

MFTabriz:
> @MFTabriz I made a quick update to the `gek/chat-rebuild-with-access-tokens` that adds a util to generate the emoji `` markup (since it looks like we'll now be doing it in at least a couple of places). Feel free to use it!
> 
> [e6abee9#diff-91877633ef1d5fe9e1c1c5b03ed30966b09faac26799cdaf840fe2aac11ff322R145](https://github.com/owncast/owncast/commit/e6abee927b08d8f7424710bcd04d583be06dbece#diff-91877633ef1d5fe9e1c1c5b03ed30966b09faac26799cdaf840fe2aac11ff322R145)

@gingervitis I rebased the branch on the top of `gek/chat-rebuild-with-access-tokens` and used createEmojiMarkup(). Although, the `:emoji:` in the alt/title tags were being replaced too. Just removed the `:` from those.

MFTabriz:
> 
> 
> @MFTabriz Ok some thoughts about the paste functionality.. So I'm guessing that the question is when you want to paste in something like `I love :partyparrot: and :bananadance: a lot!` , and you want it to show up like
> image?
> 
> I would make the `injectEmoji` function an external util instead in `chat.js`. Then somewhere in the `convertOnPaste` function, you could do the parsing in there to return the emoji image there?

@gingervitis You are right. That was my question. Thanks for the helps. I guess the main part of my problem is not fully understanding exactly how/when these values are updated and how they relate to each other:
- this.formMessageInput.current.textContent
- this.state{inputHTML}
- this.state{inputText}

gingervitis:
Oh! Ha, I actually intentionally put the `:` in the tags because I _think_ if you copy/paste the the message the alt tag gets copied with it. So if you paste it as a message maybe the emoji will populate?

If that doesn't work, the please disregard this comment!

gingervitis:
@MFTabriz 
> @gingervitis You are right. That was my question. Thanks for the helps. I guess the main part of my problem is not fully understanding exactly how/when these values are updated and how they relate to each other:
> 
> * this.formMessageInput.current.textContent
> * this.state{inputHTML}
> * this.state{inputText}

Ok it's been a while!  Let me try and remember this correctly..

- `[someFormField].textContent`--  check out https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent for what textContent does.

- So `this.state.inputText` is the textContent, or basically just chars generated from any plain text content that is read from the form field.  We're keeping this in state as one types, just to count how many text chars are being typed.
- `this.state.inputHTML` is are all the chars that make up the actual rendered markup code that needs to get created and rendered in that ContentEditable form field, while a user is typing or pasting.  We're keeping this in state to know how many actual characters are in there to be able to manipulate the cursor position, and to send that raw form to the server.

<blockquote><div><strong><a href="https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent">Node.textContent - Web APIs | MDN</a></strong></div><div>The textContent property of the Node
	interface represents the text content of the node and its descendants.</div></blockquote>

gingervitis:
> Oh! Ha, I actually intentionally put the `:` in the tags because I _think_ if you copy/paste the the message the alt tag gets copied with it. So if you paste it as a message maybe the emoji will populate?
> 
> If that doesn't work, the please disregard this comment!

To clarify, my intention was such that I often find myself doing something like this where I want to copy an existing emoji in the message
image
... and then paste it multiple times in the same same message without having to open the emoji picker again and search for the same emoji..
image
Not sure how popular of a use case this is. But I do often see users flooding a message with the same emoji multiple times and I just assume it's done with copy paste.

MFTabriz:
@gingervitis Now everything should be working as expected. Can you please check to see if it works for you too?
I created a new function `emojify()` in `webroot/js/utils/chat.js` which does the :emoji: to `` conversion and is called when user types `:` or when pasting the text from clipboard. I have added zwnj characters to the emojis' `` title & alt-tag to prevent replacing them with a new emoji img. The `convertToText()` function removes them during the pasting and the sanitized :emoji: gets pasted which then gets replaced with the image.

gingervitis:
Hey @MFTabriz awesome this is looking great! The copy/paste is working well.
I think the only minor issue I'm getting is after an :emojicode: is replaced with an image, the cursor cannot be moved backwards until more text is typed. Or more like the cursor always wants to stay at the end. But after you type something new, or blur then refocus the field, everything is fine again.

webroot/js/components/chat/chat-input.js
```diff
@@ -226,6 +244,12 @@ export default class ChatInput extends Component {
...
+      if (this.injectEmoji()) {
+        // value could have been changed, update char count
+        textValue = formField.textContent;
+      }
+    }
     this.setState({
       inputCharsLeft: CHAT_MAX_MESSAGE_LENGTH - textValue.length,
```
gingervitis, line 251: I think you'll need to add `inputText: textValue` to update state as well here

MFTabriz: Done!

MFTabriz: On second thought, we shouldn't even change the textValue here. The emojis should not add/deduct anything from the text content itself.

MFTabriz:
> Hey @MFTabriz awesome this is looking great! The copy/paste is working well.
> I think the only minor issue I'm getting is after an :emojicode: is replaced with an image, the cursor cannot be moved backwards until more text is typed. Or more like the cursor always wants to stay at the end. But after you type something new, or blur then refocus the field, everything is fine again.

@gingervitis Thanks a lot for testing this. You are right after the replacement of emoji in the text, the cursor does not go back. But unfortunately I couldn't figure out the reason behind it. Can you please look into it or refer this to someone for debugging?

gingervitis:
> > Hey @MFTabriz awesome this is looking great! The copy/paste is working well.
> > I think the only minor issue I'm getting is after an :emojicode: is replaced with an image, the cursor cannot be moved backwards until more text is typed. Or more like the cursor always wants to stay at the end. But after you type something new, or blur then refocus the field, everything is fine again.
> 
> @gingervitis Thanks a lot for testing this. You are right after the replacement of emoji in the text, the cursor does not go back. But unfortunately I couldn't figure out the reason behind it. Can you please look into it or refer this to someone for debugging?

Yeah I also tried digging into this yesterday for a bit and did not make good progress. 😞 
But since this is a pretty minor issue, let's just make a new Issue out of it and merge and close out this PR! Thank you so much for working on this @MFTabriz !

gabek:
Thank you! :1st_place_medal:

Issue #1217, opened 2021-07-18
Improve calculation and UI for when server's message size limit is reached (on new Chat branch)


gingervitis:
Currently the frontend limits to 500 text chars, excluding images/custom emojis. Apparently the backend has a limit of 512 bytes per message.

Despite the server limit, emoji images and other markup generate many extra raw characters that get sent with the message, which the server then rejects. 

What is the best way to calculate when a message size limit is reached? 

We may need to:
  - when typing a message, we'll render the emoji image to the user, but we'll be counting the chars in  `:emoji_codename:` string.
  - when the message is sent, it would get sent as `:emoji_codename:`
  - the backend would do the conversion of  `:emoji_codename:` to html markup `<img src="...emoji.png" alt="emoji".. />` and return that to clients
  - clients render the full html in the chat message when received.

^ That does seem like quite a bit of extra work that would go on top of the current Chat Refactor work, if not out of scope?

Perhaps a hacky temp solution could be:
A
- Just have the server send back an error message when too many chars are sent.
  - Surface that error when on the client side when received.

B 
- Limit the number of custom emojis to send?
  - Base that number on the longest emoji name in the custom emoji list and do the math on what the html markup for one would be? Let's say that number is X..
  - Every time an emoji is typed , we'll just deduct X from the current typed char count.
  - 😕 Hm I can see this leading to a bunch of other UX considerations that would need to be factored in.

C
- something else?

gingervitis:
I might be overthinking `B`.
@gabek has added a field in the config `maxSocketPayloadSize` which will tell us the max bytes a message can be.

We can use that number to calculate total char calculation per message. Instead of counting visible text characters in a message, we'll count all the bytes in the payload instead.

gabek:
I cranked up the max size on the backend to 2048 bytes for now until we have a real fix for this.

PR #1221, opened 2021-07-19
use maxSocketPayloadSize to calculate total bytes of message payload

gingervitis:
Addresses idea `B` in #1217


gabek:
Resolved the merge conflicts and seems to be good to go.  Works for me.  Thank you for fixing this unexpected issue!

PR #1226, opened 2021-07-20
Override default protocol if one is set in settings

### LINKED ISSUE #1225: Fix share tags
We experimented with using `//` without the protocol in the tags, but it seems like the things don't like that after all.  Oh well.  It was worth trying.

https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fnightly.owncast.tv

> Provided og:image URL, //nightly.owncast.tv/logo was not a valid URL.

And the image then doesn't show for Twitter cards either.

image

gabek:
This continues the discussion from https://github.com/owncast/owncast/pull/1187.

It restores the default protocol to `http` but overrides it if one is set in the server settings.  This is probably a good middle ground.

Closes #1225 

@jeyemwey @MFTabriz

### DIFF
--- a/controllers/index.go
+++ b/controllers/index.go
@@ -97,7 +97,7 @@ func handleScraperMetadataPage(w http.ResponseWriter, r *http.Request) {
 	// If the thumbnail does not exist or we're offline then just use the logo image
 	var thumbnailURL string
 	if status.Online && utils.DoesFileExists(filepath.Join(config.WebRoot, "thumbnail.jpg")) {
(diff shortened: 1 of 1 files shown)


jeyemwey (approved):
LGTM

MFTabriz:
@gabek I think [this line](https://github.com/owncast/owncast/blob/a84e8870eb8b7c62173adc0d4f4075798c76ba96/controllers/index.go#L100) should be adjusted too.

gabek:
Yes, thanks for catching!

PR #1232, opened 2021-07-21
In name change notification missing space added

### LINKED ISSUE #1231: The name change notification is missing a space
Whoops.

image

https://github.com/owncast/owncast/blob/develop/webroot/js/components/chat/message.js#L30

soham4abc:
Fixes #1231

### DIFF
--- a/webroot/js/components/chat/message.js
+++ b/webroot/js/components/chat/message.js
@@ -27,7 +27,7 @@ export default function Message(props) {
           <div
             class="text-white text-center opacity-50 overflow-hidden break-words"
           >
-            <span class="font-bold">${oldName}</span> is now known as
+            <span class="font-bold">${oldName}</span> is now known as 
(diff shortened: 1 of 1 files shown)


gabek:
Hi there.  This doesn't work for me, and I'm surprised this worked for you, as the trailing whitespace won't be rendered as a space.  You'll have to use a `&nbsp;`

image

soham4abc:
> Hi there. This doesn't work for me, and I'm surprised this worked for you, as the trailing whitespace won't be rendered as a space. You'll have to use a `&nbsp;`
> 
> image

Here you go! Added `&nbsp` this should do!

gabek:
Because it's in a literal string `&nbsp;` doesn't work either.  I don't think you're testing these changes before pushing them, so it's probably easier for me just to figure out a solution.  But thanks for offering!

PR #1238, opened 2021-07-22
Fix exception being thrown while typing in chat

gabek:
It looked like some values were missing so I added them.  I want to make sure these are what's expected.

### DIFF
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -25,6 +25,7 @@ import {
   CHAT_CHAR_COUNT_BUFFER,
   CHAT_OK_KEYCODES,
   CHAT_KEY_MODIFIERS,
+  CHAT_MAX_MESSAGE_LENGTH,
 } from '../../utils/constants.js';
 
 export default class ChatInput extends Component {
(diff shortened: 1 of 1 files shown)


webroot/js/components/chat/chat-input.js
```diff
@@ -25,6 +25,7 @@ import {
   CHAT_CHAR_COUNT_BUFFER,
   CHAT_OK_KEYCODES,
   CHAT_KEY_MODIFIERS,
+  CHAT_MAX_MESSAGE_LENGTH,
```
gingervitis, line 28: We shouldn't need this anymore

gabek: I was seeing an exception thrown because it's used here:

https://github.com/owncast/owncast/blob/develop/webroot/js/components/chat/chat-input.js#L248

gingervitis: Looks like line 252 needs to be updated so that it uses `props.inputMaxBytes`.

gingervitis: Actually no. `inputText` is no longer needed as we are no longer counting plain text chars, we are counting `inputHTML` markup chars instead.

webroot/js/components/chat/chat-input.js
```diff
@@ -234,6 +235,9 @@ export default class ChatInput extends Component {
 
   handleMessageInputKeyup(event) {
     const { key } = event;
+    const { inputHTML } = this.state;
```
gingervitis, line 238: These are no longer needed. 
The setState() at the end is therefore also not needed.
Instead of counting chars here, we'll let the downstream component update on `handleContentEditableChange` handle the updates of the charsLeft in state.

PR #1245, opened 2021-07-23
Prevent remote image injection with /img/emoji/ in url

MFTabriz:
Remote images with `/img/emoji/` in their url were not being filtered properly. Added a test [4fc5528] (which failed) and patched it [9785b04].


gabek:
Good catch! I never thought of that.

PR #1246, opened 2021-07-23
Fix autocomplete in case there are html elements in the chat

MFTabriz:
New getCaretPosition() returns the position counting also the html elements. Old implementation returned the caret position assuming only pure text in the box which was breaking the compatibility with emojis.

Animation2


gabek:
Works great!  Thanks for spending so much effort and finding edge cases with the text input.

PR #1250, opened 2021-07-23
Autocomplete emoji names

MFTabriz:
@gabek what is the best approach to save [emoji names list](https://github.com/owncast/owncast/compare/develop...MFTabriz:autocomplete-emoji?expand=1#diff-4435c87ff3087498a4b89d5997b45109f78aeef360a55cd0ea99a3bfc8d6a327R228) instead of generating it each time?

### DIFF
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -136,7 +136,7 @@ export default class ChatInput extends Component {
   }
 
   // autocomplete text from the given "list". "token" marks the start of work lookup.
-  autoComplete(token , list) {
+  autoComplete(token, list) {
     const { inputHTML } = this.state;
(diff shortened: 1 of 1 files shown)


gabek:
> @gabek what is the best approach to save [emoji names list](https://github.com/owncast/owncast/compare/develop...MFTabriz:autocomplete-emoji?expand=1#diff-4435c87ff3087498a4b89d5997b45109f78aeef360a55cd0ea99a3bfc8d6a327R228) instead of generating it each time?

Since it's going through them only when you hit the `Tab` key it's probably not the worst thing in the world that it regenerates the list, since it's not that common of an operation.  But if we wanted to save just the list of emoji names we could do that when emoji are pulled in via `getCustomEmojis`, and we could store them as a separate simple `emojiNames` string array in the state.

PR #1253, opened 2021-07-24
Insert emoji in the caret position

### LINKED ISSUE #1200: Insert emoji in the caret position
Currently, the [emojis are added to the end of the text](https://github.com/owncast/owncast/blob/4067bcf757bbdbd94452c4f67807a64ba34629a6/webroot/js/components/chat/chat-input.js#L120) but they should be added to the caret position instead.

MFTabriz:
closes #1200


gabek:
Works great, thank you!

PR #1258, opened 2021-07-24
Remove user's own name from the autocomplete suggestions

### LINKED ISSUE #1256: Username autocomplete suggestions should not include the user's own name

MFTabriz:
Additionally, wrong return values in the updateAuthorList() are fixed.

closes #1256


gabek:
Nice update, thanks for doing that!

PR #1261, opened 2021-07-25
Improve performance of retrieving chat history by avoiding double-sort

### LINKED ISSUE #1260: Improve performance of the chat history query
This query gets very slow if there are lots of chat messages.  Since old messages get removed this doesn't impact most streamers, but if somebody were to have thousands of chat messages within a window the query will become quite slow.

I haven't dug deep into options, but If anybody is keen on doing some database query optimization, I'd love to have your input!  

Return the most visible recent 50 messages, joined by enabled user data.

```sql
SELECT id, user_id, body, eventType, hidden_at, timestamp, display_name, display_color, created_at, disabled_at, previous_names, namechanged_at FROM (SELECT * FROM messages LEFT OUTER JOIN users ON messages.user_id = users.id WHERE hidden_at IS NULL AND disabled_at IS NULL ORDER BY timestamp DESC LIMIT 50) ORDER BY timestamp asc
```

muesli:
Avoid the double sort when we really just want to reverse the order of the result set.

### DIFF
--- a/core/chat/persistence.go
+++ b/core/chat/persistence.go
@@ -156,8 +156,15 @@ func GetChatModerationHistory() []events.UserMessageEvent {
 
 func GetChatHistory() []events.UserMessageEvent {
 	// Get all visible messages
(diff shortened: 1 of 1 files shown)


gabek:
Awesome!  Huge improvement, thank you for looking at this.

PR #1264, opened 2021-07-26
Prevent sending empty messages

### LINKED ISSUE #1263: Prevent posting empty messages in chat
If after sanitation of the markdown formatted text, the message content is empty (or just `<p></p>`), do not post it.

Screenshot 2021-07-25 at 20-09-16 Owncast Demo Server

MFTabriz:
The patch can be tested by sending a markdown image with external source.
closes #1263


gabek:
It feels a bit dirty comparing against html, but I don't know of a better way of doing it. So this accomplishes the goal, thanks!

MFTabriz:
@gabek #1266 should be better.

PR #1266, opened 2021-07-26
better chat message sanitization

MFTabriz:
improves #1264

### DIFF
--- a/core/chat/events/events.go
+++ b/core/chat/events/events.go
@@ -136,7 +136,9 @@ func sanitize(raw string) string {
 	p.AddTargetBlankToFullyQualifiedLinks(true)
 
 	// Allow breaks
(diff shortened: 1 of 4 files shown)


gabek:
I took a quick glance, but couldn't find an answer.  But really I'd prefer if we didn't render the `<p>` tags at all.  We don't need them, and if we're just going to strip them out then we need them even less.  I'm guessing there must be a way to force it to not render a container element.  I'll spend some more time looking at Goldmark at some point to see if there's an option for that.

MFTabriz:
@gabek That simplifies everything. I think this new solution must be better.

gabek:
Awesome!  Yeah this is nice and simple, and even removes a bit of the unnecessary payload.  Thank you!

PR #1270, opened 2021-07-27
Add standalone chat with ability to send messages

### LINKED ISSUE #531: Add standalone chat with ability to send messages
With a standalone chat, you could have the stream fullscreen on one display, while still having the chat open separately on a different display or device

MFTabriz:
The old embed chat is kept at `embed/chat` and two new redirects are added:` embed/chat/readonly` and `embed/chat/readwrite`
`embed/chat/readwrite` is a full-fledged chat which can be embedded. 
`app-standalone-chat.js` have been updated and now includes all the features of the main app.

closes #531

- [x] separate pages for chat overlay and embed chat functionality
- [x] chat activation and deactivation
- [x] username changing
- [x] emoji picker
- [x] full width box
- [x] username autocomplete
- [x] emoji autocomplete
- [x] emoji copy-paste
- [x] chat scroll
- [x] CSS makeup
- [x] refactor
- [x] test
- [x] update documentation


MFTabriz:
@gabek Can you please take a look at this? If you agree on the changes, I’ll update the documentations. This should also be mentioned in the release notes as a breaking change. The old chat overlay have been moved to `/embed/chatoverlay`

gabek:
Can we name things so we don't break things for people who are using the existing chat in their OBS?  Seems like we can work around that inconvenience for people.  Keep the old one `embed/chat` -> `index-standalone-chat.html` and call the new one something else.

MFTabriz:
> Can we name things so we don't break things for people who are using the existing chat in their OBS? Seems like we can work around that inconvenience for people. Keep the old one `embed/chat` -> `index-standalone-chat.html` and call the new one something else.

The new one is what people expect from an "embed chat" and the old one simply wasn’t that! I couldn’t come up with a better name. Do you have any suggestion?

gabek:
We now have chat logic in three different places, and it's almost the same in all three places.  I think it's natural to think that people who make changes to chat probably won't remember to make changes and test the two alternative UIs (especially me, I just ran into an issue with embed chat this morning that I didn't know about).  So I think we need to make it a priority to consolidate all the chat registration, parsing, rendering, etc.  It's too late in the game to do this now for 0.0.8, but I think it's the most important first thing to do for 0.0.9.

I filed this TODO: https://github.com/owncast/owncast/issues/1280

MFTabriz:
> We now have chat logic in three different places, and it's almost the same in all three places. I think it's natural to think that people who make changes to chat probably won't remember to make changes and test the two alternative UIs (especially me, I just ran into an issue with embed chat this morning that I didn't know about). So I think we need to make it a priority to consolidate all the chat registration, parsing, rendering, etc. It's too late in the game to do this now for 0.0.8, but I think it's the most important first thing to do for 0.0.9.
> 
> I filed this TODO: #1280

Actually both of the embed ones are using the same `app-standalone-chat.js` now. But this is definitely much needed!

gabek:
> The new one is what people expect from an "embed chat" and the old one simply wasn’t that! I couldn’t come up with a better name. Do you have any suggestion?

I don't think it matters too much.  Maybe both should have more specific names.  Maybe leave the existing one in and we add something like:

1. `/embed/chat/readonly`
2. `/embed/chat/readwrite`

And start referring them to the new paths, leaving the old one around so it doesn't break existing users. Not pretty names, but they're descriptive and normal people won't see them anyway.

MFTabriz:
> > The new one is what people expect from an "embed chat" and the old one simply wasn’t that! I couldn’t come up with a better name. Do you have any suggestion?
> 
> I don't think it matters too much. Maybe both should have more specific names. Maybe leave the existing one in and we add something like:
> 
>     1. `/embed/chat/readonly`
> 
>     2. `/embed/chat/readwrite`
> 
> 
> And start referring them to the new paths, leaving the old one around so it doesn't break existing users. Not pretty names, but they're descriptive and normal people won't see them anyway.

Done! I also renamed a couple of functions and variables to better reflect this.

gabek:
For those looking to test here's some steps to go through:

1. Start the server.
2. Start a stream (messages get rejected when a stream isn't active now)
3. Send some messages in the main UI.
4. Visit `/embed/chat/readonly` and `/embed/chat/readwrite` and verify your message from chat history shows up.
5. Go into the `test` directory and `node fakeChat.js` (you might have to `npm install` first if you've never used it)
6. Verify readonly and readwrite starts updating with new chat messages.
7. Stop the server and close the main frontend.
8. `rm -rf data`
9. Start the serever, start a stream, start `fakeChat.js` but don't visit the main frontend.
10. Visit `/embed/chat/readonly` or `/embed/chat/readwrite` and verify it detects the registration is invalid and re-registers.
11. Verify new messages come in.
12. Verify you can send messages on `/embed/chat/readwrite`
13. Verify you can change your name on `/embed/chat/readwrite`
14. Verify `/embed/chat` loads up the existing chat embed so existing users aren't broken.

gabek:
Looks like `/embed/chat` is redirecting to a 404.
---
Edit: i bet it's because of the `StatusMovedPermanently` and the browser has cached what the redirect is, and that old file has been removed.  So you'll probably have to use the existing name so people don't get 404s :(.  Maybe we can change that redirect to not be set as Permanent as well.

gabek:
Aside from that redirect issue things look to be working ok.  I filed something to remove the redirect completely in the future.  Though we'll still have to deal with that cached in existing browsers.  https://github.com/owncast/owncast/issues/1281

webroot/js/components/chat/chat.js
```diff
@@ -333,7 +333,7 @@ export default class Chat extends Component {
       // update document title if window blurred
       if (
         this.numMessagesSinceBlur &&
-        !this.props.messagesOnly &&
+        !this.props.readonly &&
         this.windowBlurred
       ) {
         this.updateDocumentTitle();
```
gabek, line 339: Now that people might be opening this standalone chat in a browser, people will see the title of the page.  Right now it sets it to `undefined` and looks broken, since I'm guessing there's some missing value somewhere.  It didn't matter before because the chat embed wasn't really being opened in the browser, but now people will probably see that weird title. 
image

gabek:
Tested disabling/re-enabling chat and refreshing the chat embeds and that all looks to be doing the right thing.

webroot/js/components/chat/chat.js
MFTabriz: I was not expecting people to open the embed chat directly in the browser. Just put it into an iframe and you won't even notice it!😉 
But it should be fixed now.

MFTabriz:
> ## Looks like `/embed/chat` is redirecting to a 404.
> 
> Edit: i bet it's because of the `StatusMovedPermanently` and the browser has cached what the redirect is, and that old file has been removed. So you'll probably have to use the existing name so people don't get 404s :(. Maybe we can change that redirect to not be set as Permanent as well.

I replaced the redirects with 307. We can also use a link to the new filename to keep the cached links alive for now. But at some point we have to stop working with the old links.

gabek:
Let's get it tested on actual servers! :1st_place_medal:

PR #1279, opened 2021-07-28
safely generate the thumbnail.jpg & preview.gif

rAcHekLoS:
ffmpeg writes directly into the filesystem and does not lock the file for the webserver. This change generate the files with a temp filename and move after a sucsessful ffmepg job to the final filenames.

### DIFF
--- a/core/transcoder/thumbnailGenerator.go
+++ b/core/transcoder/thumbnailGenerator.go
@@ -2,6 +2,7 @@ package transcoder
 
 import (
 	"io/ioutil"
+	"os"
 	"os/exec"
 	"path"
 	"strconv"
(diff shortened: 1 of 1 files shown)


gabek:
This is a great change, thank you!

jeyemwey:
Great findings! Thanks for finding those edgy issues and fixing them directly. :) 

I'd suggest creating a temporary directory (with os.Mktemp) which holds the temporary files instead of having them in the usual folder.

(One step further, ffmpeg can just dump the image/gif to stdout and Go can write the buffer to the public files. I can give it a try soon, but feel free to hack around, too).

gabek:
While those could be nice additions, they're certainly not necessary.  I wouldn't take the stdout route since that would just add extra  complexity with very little benefit since our code doesn't need to have a handle on that image or know the state of the write.  As long as it's not throwing errors there's nothing more we need to know.

PR #1282, opened 2021-07-28
Sort after build variants index number, fix broken index for HLS path. 

rAcHekLoS:


gabek:
Hi there!  Could you elaborate? Your change doesn't look to be touching HLS paths at all, so I'd love some additional detail around what you're addressing with this change.

controllers/video.go
```diff
@@ -8,27 +8,38 @@ import (
 )
 
 type variantsResponse struct {
-	Name  string `json:"name"`
-	Index int    `json:"index"`
+	Name               string `json:"name"`
+	Index              int    `json:"index"`
+	VideoBitrate       int    `json:"VideoBitrae"`
+	IsVideoPassthrough bool   `json:"IsVideoPassthrough"`
```
gabek: The frontend web client shouldn't need to know the specifics around the bitrate or what passthrough is.  If you're only using these values for sorting, can you change them to be `json:"-"` ?

gabek: And don't export them `VideoBitrate` -> `videoBitrate`, etc.

rAcHekLoS:
> Hi there! Could you elaborate? Your change doesn't look to be touching HLS paths at all, so I'd love some additional detail around what you're addressing with this change.

The HLS index (hls/0/, hls/1/) is the sequence of stream output settings stored in the database. When sorting, an index is taken from the FOR loop, the index number is used for variant selection in Video.js and this is not necessarily identical to the HLS paths.

You can test this as follows.  Create two variants, the first variant with a lower bitrate than the second variant and the player uses the wrong variants.

controllers/video.go
jeyemwey: > The frontend web client shouldn't need to know the specifics around the bitrate

VideoJs does know this because ffmpeg writes it in the master hls playlist.

MFTabriz:
@rAcHekLoS I just want to make sure you are aware of this: As long as you have this PR open, any commits to the PR's branch in your repository (rAcHekLoS:develop) will be reflected here as a part of your PR.

controllers/video.go
gabek: > VideoJs does know this because ffmpeg writes it in the master hls playlist.

I'm talking specifically about this API response that was changed.  Not playback.

rAcHekLoS:
sorry for my mistake, i didn't know that this automatically goes into this PR.

muesli:
Careful, there's still a commit that has slipped in from @rAcHekLoS's previous PR #1279. We can probably resolve this by simply squashing that PR instead of rebasing / merging.

controllers/video.go
```diff
@@ -10,8 +10,8 @@ import (
 type variantsResponse struct {
 	Name               string `json:"name"`
 	Index              int    `json:"index"`
-	VideoBitrate       int    `json:"VideoBitrae"`
-	IsVideoPassthrough bool   `json:"IsVideoPassthrough"`
+	VideoBitrate       int    `json:"-"`
+	IsVideoPassthrough bool   `json:"-"`
```
gabek: Your commit says you're changing them to not be exported, but they're still being exported.

gabek:
Looks like there's a conflict with your previous PR, as @muesli mentioned.  If you rebase `develop` into this PR you can update it and resolve the conflict.

muesli:
@gabek We could squash these commits in GitHub now. Alternatively I can offer to squash them directly in `rAcHekLoS/owncast` if you'd prefer that.

gabek:
Sorry to keep dragging this on, I just wanted to clarify something.

I have this example video configuration with two bitrates and one passthrough:

```
gabek@gabe-linux ~ ❯ curl --silent -u admin:abc123 http://localhost:8080/api/admin/serverconfig |jq -r '.videoSettings.videoQualityVariants[] | "\(.name) \(.videoBitrate) Passthrough: \(.videoPassthrough)"'

1.2Mbps@24fps 1200 Passthrough: false
super low 400 Passthrough: false
source 800 Passthrough: true
```

And it's sorting the passthrough between the two bitrates, is that what was expected?

`gabek@gabe-linux ~ ❯ curl --silent http://localhost:8080/api/video/variants | jq .`

```json
[
  {
    "name": "1.2Mbps@24fps",
    "index": 0
  },
  {
    "name": "source",
    "index": 2
  },
  {
    "name": "super low",
    "index": 1
  }
]
```

rAcHekLoS:
Oh, i see it. The problem is i have used the function "sort.Slice" with the variable result and for the sort criteria the variable outputVariants.  I dont have a better solution as build first the index with the other needed datas for the sort. Then use sort.slice and copy only the needed data in a new array for the JSON Output. Other solution would be use an index in the saved Stream output data in the database. but my current solution works for now.

grafik

```
[
  {
    "index": 1,
    "name": "s2"
  },
  {
    "index": 3,
    "name": "s"
  },
  {
    "index": 2,
    "name": "6000"
  },
  {
    "index": 4,
    "name": "3000"
  },
  {
    "index": 0,
    "name": "1800"
  }
]
```

controllers/video.go
```diff
@@ -7,35 +7,53 @@ import (
...
 	Index int    `json:"index"`
+	Name  string `json:"name"`
 }
 
 // GetVideoStreamOutputVariants will return the video variants available.
 func GetVideoStreamOutputVariants(w http.ResponseWriter, r *http.Request) {
 	outputVariants := data.GetStreamOutputVariants()
 
-	result := make([]variantsResponse, len(outputVariants))
+	StreamSortVariants := make([]variantsSort, len(outputVariants))
```
gabek: Rename `StreamSortVariants` to `streamSortVariants`.

controllers/video.go
```diff
@@ -7,35 +7,53 @@ import (
 	"github.com/owncast/owncast/core/data"
 )
 
+type variantsSort struct {
+	Index              int
+	Name               string
+	IsVideoPassthrough bool
+	VideoBitrate       int
```
gabek, line 14: Unexport these properties by renaming:

`IsVideoPassthrough` -> `isVideoPassthrough`
`VideoBitrate` -> `videoBitrate`

gabek (changes requested):
Just a couple renames, but other than that it looks to be doing what you expect.  Thanks for looking into this!

gabek (approved):

gabek:
Thank you, @rAcHekLoS!

PR #1290, opened 2021-08-03
Fix status code for options requests

YarmoM:
This PR fixes the HTTP status code sent by owncast upon receiving OPTIONS requests.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS), OPTIONS requests expect the 204 No Content status code instead of 200 OK.

The integration unit test has also been updated to reflect this change.


gabek:
Thank you!

PR #1295, opened 2021-08-04
Fixed bug with not correctly split IP address.

rAcHekLoS:
Quick fix for viewer counter problem with the same browsers. Better would be something more unique than the IP, users who go over a carrier-grade NAT to the Internet can not be counted uniquely.


gabek:
Looks like there's 12 commits in this PR.  I suggest creating a new branch (off of an up to date `develop`) for each PR so things are organized a bit better.

utils/clientId.go
```diff
@@ -13,9 +13,9 @@ import (
 // GenerateClientIDFromRequest generates a client id from the provided request.
 func GenerateClientIDFromRequest(req *http.Request) string {
 	ipAddress := GetIPAddressFromRequest(req)
-	ipAddressComponents := strings.Split(ipAddress, ":")
+	ipAddressComponents := strings.Split(ipAddress, ".")
 	ipAddressComponents[len(ipAddressComponents)-1] = ""
-	clientID := strings.Join(ipAddressComponents, ":") + req.UserAgent()
+	clientID := strings.Join(ipAddressComponents, ".") + req.UserAgent()
```
gabek: It probably makes sense just to simplify this, since the added complexity isn't needed to generate this identifier and relies on a specific IP version.  How about just:

```suggestion
  clientID := ipAddress + req.UserAgent()
```

And remove the rest?

rAcHekLoS:
> 
> 
> Looks like there's 12 commits in this PR. I suggest creating a new branch (off of an up to date `develop`) for each PR so things are organized a bit better.

Ok, next time I used a new branch for patches and I use rebase and not merge.

MFTabriz:
> > Looks like there's 12 commits in this PR. I suggest creating a new branch (off of an up to date `develop`) for each PR so things are organized a bit better.
> 
> Ok, next time I used a new branch for patches and I use rebase and not merge.

@rAcHekLoS You can always rebase your branch on `owncast/develop` and either drop or fixup the unnecessary commits. [This guide](https://github.com/k88hudson/git-flight-rules) may come in handy if you need any help with git.

gabek (approved):

gabek:
I think this should work just fine.  If you can clean up those unrelated commits from this PR then we can merge this in.

gabek:
I went ahead and manually cherry pick'ed the single commit in order to get this in.

PR #1313, opened 2021-08-13
keep the chat enabled on timer

### LINKED ISSUE #1311: Re-joining chat in the 5min window after disconnect shows the chat as disabled
In general the rule is chat should be available for 5min after the stream ends, even for somebody just showing up.

1. End stream
2. Chat will be enabled for 5 minutes as long as you stay on the page
3. Reload page
4. Chat will be immediately shown as disconnected

The disconnect delay countdown that's handled in `handleOfflineMode` is not being fired if you visit the page and it's already offline.

This is low importance, since it optimizes for the scenario where the stream is over, and somebody is _just_ showing up.  But it used to work.

MFTabriz:
<s>This is not so efficient as the handleOfflineMode() will be constantly called when the stream is offline. But I couldn't find any simple and elegant solution. :\ </s>

I think I have found a better solution. 

This PR also applies the changes in #1310 to the embed chat.

closes #1311

### DIFF
--- a/webroot/js/app-standalone-chat.js
+++ b/webroot/js/app-standalone-chat.js
@@ -131,7 +131,7 @@ export default class StandaloneChat extends Component {
       streamOnline: online,
     });
 
-    if (status.online != curStreamOnline) {
+    if (status.online !== curStreamOnline) {
       if (status.online) {
         // stream has just come online.
         this.handleOnlineMode();
--- a/webroot/js/app-standalone-chat.js
+++ b/webroot/js/app-standalone-chat.js
@@ -131,12 +131,14 @@ export default class StandaloneChat extends Component {
       streamOnline: online,
     });
 
-    if (status.online && !curStreamOnline) {
-      // stream has just come online.
-      this.handleOnlineMode();
-    } else if (!status.online && (curStreamOnline || curStreamOnline === null)) {
-      // stream has just flipped offline or we have just loaded the app and it is offline.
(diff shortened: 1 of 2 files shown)


webroot/js/app.js
```diff
@@ -259,12 +259,14 @@ export default class App extends Component {
       lastDisconnectTime,
     });
 
-    if (status.online && !curStreamOnline) {
-      // stream has just come online.
-      this.handleOnlineMode();
-    } else if (!status.online && curStreamOnline) {
-      // stream has just flipped offline.
-      this.handleOfflineMode(lastDisconnectTime);
+    if (status.online != curStreamOnline) {
```
gingervitis:
```suggestion
    if (status.online !== curStreamOnline) {
```

webroot/js/app-standalone-chat.js
```diff
@@ -126,27 +126,36 @@ export default class StandaloneChat extends Component {
...
-      // stream has just flipped offline.
-      this.handleOfflineMode();
-    }
-
     this.setState({
       lastDisconnectTime,
       streamOnline: online,
     });
+
+    if (status.online != curStreamOnline) {
```
gingervitis:
```suggestion
    if (status.online !== curStreamOnline) {
```

gingervitis (approved):

gabek:
Awesome, glad to see this fixed.  Thank you!

PR #1336, opened 2021-08-18
set max 10 retries for videojs playlist

### LINKED ISSUE #1335: Set a value for player maxPlaylistRetries
https://github.com/videojs/video.js/issues/7232 highlights an issue that [aaronpk is having](https://owncast.rocket.chat/channel/general?msg=JNG6aXDREaHub72Bn) where an unavailable playlist eventually grinds the browser to a halt.

Sounds like we can simply set a value for `maxPlaylistRetries` to have it eventually give up.

aaronpk:
for #1335

10 seems like a reasonable number but I chose it pretty arbitrarily. This makes the infinite loop stop after about 10-15 seconds.


gabek:
Thanks for adding this!

PR #1340, opened 2021-08-19
Reverse the behaviour of chat/disable api

### LINKED ISSUE #1330: "Disable chat" setting is confusing
It's kind of confusing having the chat enabled setting be opposite of the toggle like this:

image

image

It would be clearer if the setting were "Enable chat" and the default were "On". That way the "on" state of the toggle switch matches when the chat is visible.

MFTabriz:
For owncast/owncast-admin#286

### DIFF
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -517,7 +517,7 @@ func SetChatDisabled(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := data.SetChatDisabled(configValue.Value.(bool)); err != nil {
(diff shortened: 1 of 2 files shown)


MFTabriz:
<s>I think I must be able to handle everything client side</s> [apparently not so easily!]

gabek:
Could you share the thought behind this?  I thought you were going to change this on the frontend instead.  If you have to flip the boolean on the way in and the way out on the backend then the value being sent from the frontend seems like it's wrong.

MFTabriz:
Sorry for the trouble. I moved all the changes to the frontend.

PR #1351, opened 2021-08-26
Feature/oc 1316 add support for system message to single user

### LINKED ISSUE #1316: Create external API to enable sending a message directly to a client instead of publicly
An example is a bota sending welcome messages as somebody joins.  See https://github.com/owncast/owncast-examples/pull/1 for some use cases.

unclearParadigm:
Hello,

I've implemented an additional HTTP endpoint `/api/integrations/chat/system/client/{clientId}` and some very simple `gorilla/mux`-like REST-Helpers that make it easier to parse REST-Like params from an URL in the future.

This http-endpoint calls into core/chat.go, which will find the ChatClient in the connected clients by its Id. 
If the Id is found, I have used the already existing sendMessageToClient function (with the addition that contents are now markdown rendered, hope that's allright?)

Every webhook firing an Event of Type "UserEvent" will now have a ClientId included, which can be utilized by thirdparty integrations to use this endpoint.

2 further Merge Requests incoming (Demo-Bot, which I uesd for testing of everything, and a short update to the Docs).

BR, Ruffy

### DIFF
--- a/core/chat/chat.go
+++ b/core/chat/chat.go
@@ -41,6 +41,12 @@ func GetClientsForUser(userID string) ([]*Client, error) {
 	return clients[userID], nil
 }
 
+// FindClientByID will return a single connected client by ID.
+func FindClientByID(clientID uint) (*Client, bool) {
+	client, found := _server.clients[clientID]
+	return client, found
+}
+
 // GetClients will return all the current chat clients connected.
 func GetClients() []*Client {
 	clients := []*Client{}
--- a/core/chat/chat.go
+++ b/core/chat/chat.go
@@ -111,6 +111,13 @@ func SendAllWelcomeMessage() {
 	_server.sendAllWelcomeMessage()
 }
 
+// SendSystemMessageToClient will send a single message to a single connected chat client.
+func SendSystemMessageToClient(clientID uint, text string) {
+	if client, foundClient := FindClientByID(clientID); foundClient {
+		_server.sendSystemMessageToClient(client, text)
+	}
+}
+
 // Broadcast will send all connected clients the outbound object provided.
 func Broadcast(event events.OutboundEvent) error {
 	return _server.Broadcast(event.GetBroadcastPayload())
--- a/utils/restendpointhelper.go
+++ b/utils/restendpointhelper.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 )
 
-const restUrlPatternHeaderKey string = "Owncast-Resturl-Pattern"
+const restUrlPatternHeaderKey = "Owncast-Resturl-Pattern"
 
 // takes the segment pattern of an Url string and returns the segment before the first dynamic REST parameter.
 func getPatternForRestEndpoint(pattern string) string {
--- a/utils/restendpointhelper.go
+++ b/utils/restendpointhelper.go
@@ -3,7 +3,6 @@ package utils
 import (
 	"errors"
 	"fmt"
-	"log"
 	"net/http"
 	"strings"
 )
@@ -50,7 +49,6 @@ func readParameter(pattern string, requestUrl string, paramName string) (string,
 
 func ReadRestUrlParameter(r *http.Request, parameterName string) (string, error) {
 	pattern, found := r.Header["OWNCAST-RESTURL-PATTERN"]
-	log.Print(r.Header)
 	if !found {
 		return "", errors.New(fmt.Sprintf("This HandlerFunc is not marked as REST-Endpoint. Cannot read Parameter '%s' from Request", parameterName))
 	}
@@ -62,7 +60,6 @@ func RestEndpoint(pattern string, handler http.HandlerFunc) (string, http.Handle
 	baseUrl := getPatternForRestEndpoint(pattern)
 	return baseUrl, func(w http.ResponseWriter, r *http.Request) {
 		r.Header["OWNCAST-RESTURL-PATTERN"] = []string{pattern}
-		log.Print(r.Header)
 		handler(w, r)
 	}
 }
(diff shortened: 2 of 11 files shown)


gabek:
Interested in extra credit? :)

Our API documentation is published via an OpenAPI 3 spec at https://github.com/owncast/owncast/blob/develop/openapi.yaml.  Would you be interested in adding your API to this document?  

It's very manual, but I use the `42crunch.vscode-openapi` VSCode extension to help catch syntax errors and preview the doc, in case you use VSCode.

unclearParadigm:
Will do. Didn't know this is done manually.

MFTabriz:
@unclearParadigm Please check the [linter report](https://github.com/owncast/owncast/actions/runs/1173168948) and address the errors if possible.

unclearParadigm:
Whoops Poopsie. People call me the Techdebt Factory ;) I'm on it.

core/chat/chat.go
```diff
@@ -40,6 +40,11 @@ func GetClientsForUser(userID string) ([]*ChatClient, error) {
 	return clients[userID], nil
 }
 
+func FindClientById(clientId uint) (*ChatClient, bool) {
```
gabek: If this isn't being used outside of the `chat` package you can un-export it (`findClientById`).

unclearParadigm: Allright, made it private now.

utils/restendpointhelper.go
```diff
@@ -0,0 +1,67 @@
+package utils
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+const restUrlPatternHeaderKey string = "Owncast-Resturl-Pattern"
```
gabek: I'm a little confused about this.  It seems like you're setting a header value as a guard, so only requests with that header is allowed to use the path parameters.  But it feels kind of weird.  Can we just return no parameters if none exist?

gabek: Oh and as just a random aside, you don't need to specify the type `string` if you're assigning a value, but if you're declaring an empty var then you do.

unclearParadigm: Okay, I feel like I should have outlined how the REST-endpoint magic works (or how its supposed to work).

Initially I was investigating the `gorilla/mux` framework for matching REST endpoints. As discussed in a previous thread, there is and will be the need for REST endpoint matching, but we should stay away from third-party dependencies if possible and rather implement our own. So I implemented a very very simple way to handle such endpoints.

First I want to explain how I intended to parse Variables from the Path/Segment of an Url:
suppose we have a URL-segment that looks like this `/user/12/friend/13/name` and we want to parse out `12` and `13`.  A simple (but very unsafe way) to parse these out would be to split the string by '/' and access the vars by accessing split[0] and split[2]. The more reliable and secure version, is to match the Url against a pattern. So this is where this `Rest-Url-Pattern` comes into play. Basically we define a pattern (other web-frameworks like Flask, ASP.NET modelbinder, refer to this as "Segment-Mask") that is used as Pattern-String for the Input URL. Given the above example, the mask would look like this: `/user/{userId}/friend/{friendId}/name`. That mask helps to safely extract the actual parameters out of the URL path. 

#### Extracting an parameter out of the URL path is implemented like this

* need the REST-pattern and the URL-segment (current URL Request Path) as string 
* Split both the REST-pattern and the actual URL-segment by `/`
* the resulting arrays/slices  must have equal length/dimensions (if not, something is wrong, abort)
* to lookup the location of the `{clientId}` parameter in the Url:
    *  find the Index in the split of the REST-pattern (the index that contains the placeholder `{clientId}`)
    *  and then use this index to access the element at the same index in the URL-Segment split array.
 * evoila we've safely extracted the input parameter without relying on any assumptions, and without relying more than necessary on user-crafted input

#### You might ask, why this REST-pattern is there at all

* it's a very performant way to parse parameters
* it makes these requests more readable and easier to structure in the mux
* it's safer than directly parsing arbitrary user-input (reducing attack-surface for parameter tampering)
* you can identify faults/errors in the URL-segment immediately (e.g. different dimensions)
* you could additional type-checks something like: `/user/{clientId: int}/friend/{friendId: int}/name` (at one point in the future)

#### Okay, if you're cool with that part so far - let's close the loop

The [`RestEndpoint`- Request-Middleware](https://github.com/owncast/owncast/pull/1351/files#diff-a2a44c38aa0276fd98d206598799c9eec7f0d0700704975ff05ec1f32775168dR61) takes this REST-Pattern and registers a HandlerFunc for this Pattern. Internally this middleware will reduce `/user/{clientId: int}/friend/{friendId: int}/name` to `/user/` and register the passed HandlerFunc with `/user/` to the `net/http` mux.  I had to do it like this, because the vanilla `net/http` server does not support partial Url matching (or placeholders in Urls). 

Inside the Controller-Function in [`controllers/admin/chat.go`](https://github.com/owncast/owncast/blob/5fc62eefdc39f9374911ca373d8b4df05a52f505/controllers/admin/chat.go#L131) I'm using the `clientIdText, err := utils.ReadRestUrlParameter(r, "clientId")` function to parse out the ClientId from the template. Now this function requires the REST-pattern to be present to safely work (see REST-parameter parsing part above). The solution that I came up with, is to pass the REST-Pattern from the `RestEndpoint`-middleware further down the middleware chain until it finally reaches a handler that needs the pattern for parsing out a parameter. The nicest way I could think of, was utilizing the HTTP headers. Now the middleware sets an HTTP Header var with the Pattern, that then can be read from the `utils.ReadRestUrlParameter(r, "clientId")` inside the actual HandlerFunc.

That guard clause (that I think you are referring to) inside the `ReadRestUrlParameter`-func is actually just a guard making sure, that the Rest-Pattern is set by the middleware (or in other words, the developer actually decorated the HandlerFunc with the `RestEndpoint` middleware). In the real world, this guard clause won't ever cause a problem. I see it more or less as a "fail-fast"-shortcircuit for developers using this helpers.

Excuse me for slashing this together like this, also excuse me for the wall of text. I hope this explanation at least somewhat makes sense and is acceptable for you.

gabek: Thank you for all the detailed clarification!

I went ahead and pushed a few tweaks (mostly renames, comments) needed by the linter and resolved some merge conflicts.

PR #1355, opened 2021-08-28
do not exit after successful stream key change

nothub:
this makes running preconfigured instances in docker easier and i could not find any reason why the application should exit after the stream key change.


gabek:
Totally fair!  Thanks!

PR #1356, opened 2021-08-29
Add support for loading the owncast URL in mpv

MFTabriz:
mpv first tries to load the URL with the user-agent set as a typical browser. The returned page gets parsed with the yt-dl. In this phase we cannot detect/intercept it. 

In case of failure in the first step, mpv will load the same URL with the user-agent set as “mpv” and pipes the result directly into ffmpeg. If we redirect it to the bot page here, ffmpeg will not be able to parse the html page and extract the hls link. But instead we can just redirect it to the hls link and ffmpeg should handle the rest.

PS: +Bonus VLC support! 😎

### DIFF
--- a/controllers/index.go
+++ b/controllers/index.go
@@ -44,6 +44,11 @@ func IndexHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


gabek:
:1st_place_medal:

PR #1367, opened 2021-08-31
Improve no-JS landing page

### LINKED ISSUE #1139: Add proper landing page for browsers without the JS support
I understand that we need the client-side JS support for the app to work. But the JS-disabled browsers should be presented with

1. A proper explanation of why Owncast needs JS to work, and
2. Some alternative ways of watching the stream (maybe just throw a HLS link for now)

What they are currently encountering is this:
Screenshot 2021-06-26 at 18-08-33 Owncast

MFTabriz:
This PR depends on the changes in #1356. Please fix the grammar mistakes and/or improve the explanation if needed.

Screenshot 2021-08-31 at 23-09-40 Owncast

closes #1139

### DIFF
--- a/webroot/index.html
+++ b/webroot/index.html
@@ -133,22 +133,22 @@
         This website is powered by <a href="https://owncast.online" rel="noopener noreferrer" target="_blank">Owncast</a>.
       </p>
       <p>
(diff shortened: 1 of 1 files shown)


gabek:
Nice, thank you!

PR #1393, opened 2021-09-10
Simplify HLS storage paths

### LINKED ISSUE #875: Simplify where HLS segments live on the filesystem
Right now we move HLS segments around depending on if they're being served from the server itself or from external storage.

We should change it so we always just keep files in the `hls` (or possibly even move to `data/hls`) directory, completely getting rid of the `webroot/hls` directory.  And then in our router we just simply enable or disable a new route to allow serving files from that directory.

This would simplify the LocalStorageProvider and remove some filesystem action.

gabek:
Remove the "public" vs "private" HLS storage directories and create a new HLS controller for managing access to the files.

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -7,8 +7,6 @@ const (
 	StaticVersionNumber = "0.0.9" // Shown when you build from develop
 	// WebRoot is the web server root directory.
 	WebRoot = "webroot"
-	// PrivateHLSStoragePath is the HLS write directory.
-	PrivateHLSStoragePath = "hls"
 	// FfmpegSuggestedVersion is the version of ffmpeg we suggest.
 	FfmpegSuggestedVersion = "v4.1.5" // Requires the v
 	// DataDirectory is the directory we save data to.
@@ -18,8 +16,9 @@ const (
 )
 
 var (
-	// PublicHLSStoragePath is the directory we write public HLS files to for distribution.
-	PublicHLSStoragePath = filepath.Join(WebRoot, "hls")
 	// BackupDirectory is the directory we write backup files to.
 	BackupDirectory = filepath.Join(DataDirectory, "backup")
+
+	// HLSStoragePath is the directory HLS video is written to.
+	HLSStoragePath = filepath.Join(DataDirectory, "hls")
 )
--- a/controllers/hls.go
+++ b/controllers/hls.go
@@ -1,0 +1,66 @@
+package controllers
+
+import (
+	"io/ioutil"
+	"net/http"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/owncast/owncast/config"
+	"github.com/owncast/owncast/core"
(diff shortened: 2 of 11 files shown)


controllers/hls.go
```diff
@@ -0,0 +1,66 @@
...
+		core.SetViewerIdActive(id)
+	} else {
+		cacheTime := utils.GetCacheDurationSecondsForPath(relativePath)
+		w.Header().Set("Cache-Control", "public, max-age="+strconv.Itoa(cacheTime))
+	}
+
+	writeHLSResponse(fullPath, w)
+}
+
+func writeHLSResponse(filePath string, w http.ResponseWriter) {
```
jeyemwey: There are things missing by throwing out `http.FileServer` and just using `io.Copy`, like 404 Not Found, Range Requests, If-Unmodified-Since, or Mime-Types. Id suggest to use [`http.ServeFile`](https://pkg.go.dev/net/http#ServeFile) or [`http.ServeContent`](https://pkg.go.dev/net/http#ServeContent) which deals with all of the above.

config/constants.go
```diff
@@ -18,8 +16,9 @@ const (
 )
 
 var (
-	// PublicHLSStoragePath is the directory we write public HLS files to for distribution.
-	PublicHLSStoragePath = filepath.Join(WebRoot, "hls")
 	// BackupDirectory is the directory we write backup files to.
 	BackupDirectory = filepath.Join(DataDirectory, "backup")
+
+	// HLSStoragePath is the directory HLS video is written to.
+	HLSStoragePath = filepath.Join(DataDirectory, "hls")
```
jeyemwey, line 23: I have a hard time understanding why the data directory is your preferred location for storage. Previously, it was clear that HLS files were served „with the webserver“ if they were in the `/webroot` folder. Now, the segments are in another folder which also serves anything that’s in it, but is not marked as „public for everyone to access“ as `/webroot`.

IMO, `HLSStoragePath` should always be `/webroot/hls`, even if S3 is activated, but I’m open for a debate here.

gabek: It's because people have asked us to centralize where we write things so they can run Owncast from read-only locations and only have to map a single writable directory. If we work towards having `data` be the single place where we read and write user data then it simplifies concerns when people deploy.  We're not there yet as we're still writing images to `webroot`, but I'd like to move those too.  But going forward we should not be writing to `webroot`.
 
It also very much makes sense to simplify and always write files to the same location instead of the weird "sometimes here and sometimes there" logic we currently have. A single location where if public access is needed then a endpoint specifically built for that purpose should be used, not just throwing the files under `webroot`.  

And absolutely these files should not be publicly available if the configuration says video is not being served locally.

jeyemwey: Ack, thanks for clarifying.

controllers/hls.go
gabek: Good thoughts!  No reason not to use them.

MFTabriz (approved):

PR #1396, opened 2021-09-11
Expanded linting + fix warnings

gabek:
Every once in a while I discover my linter locally is not setup properly and I'm missing all kinds of warnings.  And because we let linter failures pass on push they just silently go through.

So I did this:

1. Made it so Linter failures break the build.
2. I added some expanded checks. 
4. Pulled some specific ignores from the default list of golang-ci lint ignores.
3. I went through and fixed everything.

Hopefully going forward they won't silently get through, even if they don't show up in my own local environment, they'll complain remotely.

Some of these things are trivial and worth ignoring, but I'd like to be better about consistency, so I'd rather follow some arbitrary guide to have something to follow. (Id -> ID, Url -> URL, etc).

This PR is both a sanity check, open to suggestions, and a heads up.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           # Optional: golangci-lint command line arguments.
-          args: --issues-exit-code=0 --timeout 2m0s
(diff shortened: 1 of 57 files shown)


muesli (approved):
Looks good to me!

Issue #1293, opened 2021-08-04
Add chat message design for external Federated engagement


gabek:
This would be similar to the "system message", but be able to show an avatar of a federated user and have a unique look to make it clear it's something that happened on the fediverse.

Example events:

- [x]  x@y.com favorited the "Go Live" federated message
- [x]  x@y.com boosted/reposted the "Go Live" federated message.
- [ ]  x@y.com started following the Owncast server on the fediverse.

gingervitis:
Something like this?
<img width="323" alt="image" src="https://user-images.githubusercontent.com/9563255/132949913-b1d7b274-ec2a-4554-829d-ef3aa1eb9ef2.png">


https://github.com/owncast/owncast/pull/new/gw/20210911-external-fed-chat-eng

gingervitis:
We should probably start sticking with a consistent icon library at some point. Or we could keep the emojis, we just have to make sure the emojis aren't too new so that most OS's will support them.

gabek:
Because these are users with username, avatar, as well as the full account name, I wonder if we can display all of that so we look like we're fully tied into the fediverse.  It might also encourage more people to interact if they get their full name and image displayed.

![image](https://user-images.githubusercontent.com/414923/132965843-c4b8ba51-028f-4f2c-b3a2-4ff0a5d2dc07.png)

gabek:
Yeah emoji are rough, I know on Linux I run into a lot of missing emoji, so I probably would rather using a standardized set of images.

gabek:
First pass at this is in.

![image](https://user-images.githubusercontent.com/414923/144171413-28a4771b-6c44-4cef-a3f7-b0ee674fe98e.png)

PR #1397, opened 2021-09-11
WIP: Fediverse interaction prototype for #1293

gingervitis:
Some mock placeholders

There are a bunch of stubs for
- websocket message type
- message content
- how will the fediverse action message be presented? I'm assuming that the server will generate the actual "action" message sentence? And it should just return that + action type?


gabek:
Closing this in favor of using the same follower look as the follower list (avatar, name, account).

PR #1400, opened 2021-09-13
Adjust error messages to conform to Go best practices

lindner:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings


lindner:
Just getting my workflow setup and doing a simple change.  More to come!

gabek:
Awesome, thank you! I'm trying to get better about consistency going forward, so this kind of thing matters.

PR #1411, opened 2021-09-16
Enable CORS in the hls handler

### LINKED ISSUE #1409: Re-enable CORS in the HLS handler
CORS didn't get added in the new controller and it needs to be.

https://github.com/owncast/owncast/blob/develop/controllers/hls.go#L18

u5surf:
Fixes #1409 
Add CORS response header in the hls handler.


gabek:
Thank you, @u5surf!

PR #1413, opened 2021-09-16
Add hashes for standalone video/chat

rAcHekLoS:
Add hashes for standalone chat and video pages, the commit "Merge pull request from GHSA-2hfj-cxw7-g45p" add the CSP header only for the index.html.


gabek:
This is great, thank you for catching this!

PR #1415, opened 2021-09-17
Automated browser testing

### LINKED ISSUE #388: Create automated browser-based testing
Suggested by @mattdsteele in https://github.com/owncast/owncast/pull/384#issuecomment-731433136.

We can use the same appraoch as the API-based testing but instead of firing off Jest it can run Puppeteer tests.  This can live in a Github workflow in the same way.

I like Matt's suggestions of some top level things to test:

1. Make sure there's no console errors.
1. Make sure the page renders.
1. Save a screenshot for review.
1. Make sure some obvious components exist.

I could see us also wanting to test the chat because it's one of the few actions a user takes.
1. Send a test chat message.
2. Make sure that chat message is visually rendered.

gabek:
# Automated browser tests

- It uses Puppeteer + Jest.
- It it similar to how the API tests work

The tests currently address the following interfaces:

1. The main web frontend of Owncast
1. The embeddable video player
1. The embeddable read-only chat
1. the embeddable read-write chat

Each have a set of test to make sure they load, have the expected elements on the screen, that API requests are successful, and that there are no errors being thrown in the console.

The main web frontend additionally iterates its tests over a set of different device characteristics to verify mobile and tablet usage and goes through some interactive usage of the page such as changing their name and sending a chat message by clicking and typing.

While it emulates the user agent, screen size, and touch features of different devices, they're still just a copy of Chromium running and not a true emulation of these other devices. So any "it breaks only on Safari" type bugs will not get caught.

It can't actually play video, so anything specific about video playback cannot be verified with these tests.

## Setup

`npm install`

## Run

`./run.sh`

## Screenshots

After the tests finish a set of screenshots will be saved into the `screenshots` directory to aid in troubleshooting or sanity checking different viewport sizes.


jeyemwey:
Wouldn't you commit the screenshots and diff against them to see if something breaks?

gabek:
If you've set that up in the past or want to look into it feel free to add it to these tests, but visual regression wasn't the goal for me here.

PR #1416, opened 2021-09-17
Do not send user joined messages for already active users

### LINKED ISSUE #1406: Do not show user joined chat message if user has recently joined
Keep track of user ids and last joined and skip the message if they recently joined.

gabek:
This aims to stop the scenario where somebody is on a bad network, or is refreshing their browser for whatever reason, forcing User Joined messages to show over and over again in the chat.  It's annoying enough for me to do something about it.

With this change if a user ID has been active within 10 minutes then don't alert everyone that the user has joined, because they've already been told once previously.  If we think 10 minutes is too aggressive we can reduce it.

Closes https://github.com/owncast/owncast/issues/1406


gabek:
Reduced to 5 minutes.  Merging.

PR #1441, opened 2021-09-29
fix insecure math/rand use for access token generation

ghost:


gabek:
I don't believe there to be any downside to using `crypto/rand` instead.

PR #1442, opened 2021-09-29
refactor geoip

ghost:
- Introduce a new Client type to remove the global variables from the file
- Use the sync package to prevent race conditions with the cache and
  enabled flag
- Cache results for IPs, even if the result is nil

There are still data races around the client.Geo variable, but that can be
resolved in a future commit.

### DIFF
--- a/core/chat/server.go
+++ b/core/chat/server.go
@@ -38,6 +38,8 @@ type Server struct {
 
 	// unregister requests from clients.
 	unregister chan uint // the ChatClient id
+
+	geoipClient *geoip.Client
 }
 
 // NewChat will return a new instance of the chat server.
@@ -51,6 +53,7 @@ func NewChat() *Server {
 		inbound:                  make(chan chatClientEvent),
 		unregister:               make(chan uint),
 		maxSocketConnectionLimit: maximumConcurrentConnectionLimit,
+		geoipClient:              geoip.NewClient(),
 	}
 
 	return server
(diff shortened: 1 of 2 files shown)


gabek:
I've never seen all package-level state as a problem.  But I see it from the standpoint that if a package is a singleton then keeping private state in a struct that is passed around just amounts to passing around more state than necessary (even when not exported).  Like for example in this case the geoip package is a singleton and the cache belongs to the package, and passing it around feels unnecessary. It would feel different if there needed to be multiple unique instances of it that needed to be encapsulated.

That being said, anything that requires a lock and isn't getting locked should be resolved for sure. And don't get me wrong, I don't have a problem wrapping everything in a struct instead of package-level state, I just don't think it's necessarily wrong or required in all cases.  But I could also be totally wrong in my thoughts :)

ghost:
Having a package as a singleton has more downsides than upsides in my opinion:
- Having testing of the package means that the state of the package has to be reset every test iteration, rather than just instantiating a type
- You cannot have a package implement an interface, only a type (e.g. the non-singleton `geoip.Client` type could implement a `GeoipResolver` interface, but the singleton package `geoip` could not).
- Similar to this previous point, but passing around and using interfaces allows components of the owncast system to be more plugable, rather than being tied to a concrete implementation by a package (i.e. allows the system to be more decoupled).
- Packages cannot be imported and used multiple times by different instances of the core system without conflict. Maybe this isn't a major issue now, but likely would be if owncast ever wants to support multiple incoming streams. Or if owncast ever wanted to be exposed as a library that others could import and use.

gabek:
I'm on board with your examples for sure, they make a lot of sense, thanks for sharing your thoughts!  While there are certainly times where I don't think everything needs to be wrapped up in a struct, more things will to allow for swapping out implementations of interfaces.

Thanks again for chiming in and for your help.

Issue #1439, opened 2021-09-29
invalid memory address or nil pointer dereference 


ghost:
```
INFO[2021-09-29T12:13:43-05:00] Video transcoder started using x264 with 1 stream variants.
INFO[2021-09-29T12:14:53-05:00] Inbound stream disconnected.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xc5d3fd]

goroutine 3431 [running]:
github.com/owncast/owncast/core.SetStreamAsDisconnected()
	/go/src/github.com/owncast/owncast/core/streamState.go:97 +0x16d
github.com/owncast/owncast/core.setStreamAsConnected.func1.1(0x0, 0x0)
	/go/src/github.com/owncast/owncast/core/streamState.go:62 +0x20
github.com/owncast/owncast/core/transcoder.(*Transcoder).Start(0xc000f3d2b0)
	/go/src/github.com/owncast/owncast/core/transcoder/transcoder.go:126 +0x501
github.com/owncast/owncast/core.setStreamAsConnected.func1(0xc001266028)
	/go/src/github.com/owncast/owncast/core/streamState.go:67 +0x7e
created by github.com/owncast/owncast/core.setStreamAsConnected
	/go/src/github.com/owncast/owncast/core/streamState.go:59 +0x263
```

Streaming with external S3 storage. I haven't tried to reproduce the crash.

Version: v0.0.9

gabek:
If you have a chance to reproduce this, or can share any specific configuration that makes this occur, please do.

ghost:
To reproduce:

1. Start v0.0.9 with empty data directory
2. Open OBS Studio to configure to stream to owncast instance
3. Rapidly click "Start Streaming"/"Stop Streaming"
4. owncast will crash

I believe there's data races happening. In just taking a peek at the code, I notice there is a lack of locking on global variables (e.g. `_hasInboundRTMPConnection`), which would account for variables being in an invalid state.

bontibon:
I believe there's data races happening. In just taking a peek at the code, I notice there is a lack of locking on global variables (e.g. `_hasInboundRTMPConnection`), which would account for variable be in an invalid state.

gabek:
> Rapidly click "Start Streaming"/"Stop Streaming"

Certainly sounds like a race condition to me.

gabek:
[...]

PR #1444, opened 2021-09-30
remove global variables from core/storageproviders

ghost:


gabek:
Could you provide some background on why you're proposing these changes?

ghost:
Related to #1439, I am concerned about the amount of global state that is used throughout the source code. I believe it is in the best interest of stability and memory safety that the amount package-wide variables be reduced and converted to struct fields with locking, where appropriate.

gabek:
These changes make sense since things like the uploader is specific to s3 logic and not to the storageprovider package as a whole, so architecturally it shouldn't be available to other logic in the package.  Thank you!

gabek:
Thanks again for cleaning this up.  Just did some testing and all looks good.

PR #1448, opened 2021-10-03
:rocket: fix: split EXPOSE into multiple statements

### LINKED ISSUE #1447: Plesk does not support our Dockerfile syntax
Hello.

Background:

I'm trying to get Owncast running via Docker within Plesk.  It isn't as simple as docker-compose or just simply running docker normally .

-----

In my testing to see if indeed it was something that I screwed up on I tested portainer as it uses three ports ( see https://github.com/portainer/portainer/blob/develop/build/linux/Dockerfile#L8..L10 ) and **that** does work.

I followed the same sort of setup to get Owncast going, but it seems your `EXPOSE` combines two ports in the same line.

I've also done some other digging and experimenting to see if this is just a one off on Plesk being a pain in the you know where but it isn't.

It seems your `EXPOSE` of combining two ports on the same line causes issues - see https://github.com/owncast/owncast/blob/develop/Dockerfile#L3

Sidenote: I actually have Plesk running on a Vultr VPS as I saw on #1446

jeyemwey:
This is a change in the Dockerfile because Plesk seems to be bad with having multiple ports in one `EXPOSE` statement.

Related to #1447.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # Perform a build
 FROM golang:alpine AS build
-EXPOSE 8080 1935
+EXPOSE 8080
+EXPOSE 1935
 RUN mkdir /build
 ADD . /build
 WORKDIR /build


jeyemwey:
Not necessary since #1463 fixes the issue. See https://github.com/owncast/owncast/issues/1447#issuecomment-939384799 for explanation.

PR #1453, opened 2021-10-05
Replaces redirect of /embed URLs to return content directly.

### LINKED ISSUE #1281: Replace redirect of /embed URLs to return content directly
I'm not sure why I did it this way in the first place, but we should change it.

Example: Instead if redirecting `/embed/video` to `index-video-only.html` it should read `index-video-only.html` and return the content directly.  Same with the chat versions.

Some import changes may need to be made in the files since the URL for the content will be different, but this will be cleaner since it probably confuses people as to why a different file is loaded when they use an embed url.

krashanoff:
# Description

- Removes relative imports from embedded player, chat.
- Replaces `Redirect` calls with `ServeFile` in `embed.go`.

This PR replaces the temporary redirect of owncast's `/embed` URLs with calls to `http.ServeFile` to directly supply the embeds. This will reduce confusion when embedding owncast streams. In the process of making this change, relative imports in `index-standalone-chat-.*` and `index-video-only.html` were replaced with absolute imports. This way, the proper resources are loaded by the pages regardless their path.

Fixes #1281.

### DIFF
--- a/controllers/embed.go
+++ b/controllers/embed.go
@@ -6,15 +6,15 @@ import (
 
 // GetChatEmbedreadwrite gets the embed for readwrite chat.
 func GetChatEmbedreadwrite(w http.ResponseWriter, r *http.Request) {
-	http.Redirect(w, r, "/index-standalone-chat-readwrite.html", http.StatusTemporaryRedirect)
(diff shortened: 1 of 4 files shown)


gabek:
This is great, thank you for changing this and resolving the import paths needed to make it work. I appreciate your help!

PR #1471, opened 2021-10-13
Fix FAQ link

mscherer:
https://owncast.online/docs/faq seems to return a 404 error.


jeyemwey:
Thanks for noticing (:

PR #1472, opened 2021-10-13
Fix hashes for standalone video/chat

rAcHekLoS:
Fix the hashes for inline javascript on the index-video-only.html, index-standalone-chat-readonly.html and index-standalone-chat-readwrite.html

### DIFF
--- a/controllers/embed.go
+++ b/controllers/embed.go
@@ -2,19 +2,30 @@ package controllers
 
 import (
 	"net/http"
+
+	"github.com/owncast/owncast/router/middleware"
 )
 
 // GetChatEmbedreadwrite gets the embed for readwrite chat.
 func GetChatEmbedreadwrite(w http.ResponseWriter, r *http.Request) {
+	// Set our global HTTP headers
+	middleware.SetHeaders(w)
+
 	http.ServeFile(w, r, "webroot/index-standalone-chat-readwrite.html")
 }
 
 // GetChatEmbedreadonly gets the embed for readonly chat.
 func GetChatEmbedreadonly(w http.ResponseWriter, r *http.Request) {
+	// Set our global HTTP headers
+	middleware.SetHeaders(w)
+
 	http.ServeFile(w, r, "webroot/index-standalone-chat-readonly.html")
 }
 
 // GetVideoEmbed gets the embed for video.
 func GetVideoEmbed(w http.ResponseWriter, r *http.Request) {
+	// Set our global HTTP headers
+	middleware.SetHeaders(w)
+
 	http.ServeFile(w, r, "webroot/index-video-only.html")
 }
--- a/router/middleware/headers.go
+++ b/router/middleware/headers.go
@@ -21,7 +21,7 @@ func SetHeaders(w http.ResponseWriter) {
 	}
 	// Content security policy
 	csp := []string{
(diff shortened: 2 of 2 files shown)


gabek:
It doesn't seem to be broken as far as I can tell in `develop`:

https://nightly.owncast.tv/embed/chat/readwrite
https://nightly.owncast.tv/embed/chat/readonly
https://nightly.owncast.tv/embed/video

How are you experiencing the errors?

rAcHekLoS:
Firefox and Edge says "An attempt to execute inline scripts has been blocked".

Under https://owncast.racheklos.com/embed/chat/readwrite runs the version with the new hashes.

You run your build with BROWSER_TEST == "true"?

gabek:
Are you saying the links above, pointing to the nightly build off of `develop` are throwing those errors on Firefox and Edge? Or just your local instance?

I just tested them with Firefox and I'm not seeing any issues.

`BROWSER_TEST` is used in browser tests [here](https://github.com/owncast/owncast/blob/f0bd7d2528049ca5956a568de19505004fa1b811/test/automated/browser/run.sh#L22) to allow injection of test scripts.

rAcHekLoS:
I had checked out the latest dev version and with that i had the error "An attempt to execute inline scripts has been blocked". Then i changed the hashes and created a new build.

The nightly build from the links works on my system. Strange ;)

rAcHekLoS:
I have calculated the hashes from the javascript on https://nightly.owncast.tv and this is exact the same hash as from my commit!

<blockquote><div><strong><a href="https://nightly.owncast.tv/">Owncast</a></strong></div></blockquote>

gabek:
One other person in chat mentioned they were having hash mismatches. As far as I can tell you two are the only two that it's broken for. I wish I knew what was going on here.

rAcHekLoS:
The fact is that the hashes have changed due to the following commit: https://github.com/owncast/owncast/commit/958c78641e8adb2ebf6abb71c126622c0399866c#diff-b4457c24a0e2302c29a678250532790aca45822e4b4aa58d20d9c2435139efbf

I think it has something to do with the server settings, e.g. if gzip is active or not. The CSP header is not delivered for the main html, but only for js, css etc. files.

grafik
grafik

gabek:
It makes sense that if you have a proxy in between, and that's changing the content, but is that what's happening to you?

Since I'm still not sure what's going on, I just want to clarify how I'm seeing things:

1. It works for me in Chrome and Firefox, locally and deployed through a proxy.
2. Only one other person, who hasn't been able to give details, has mentioned something that sounds similar.
3. So is there a problem with Owncast? Or is it some proxy configuration that's out of our hands?  Why does it not work for you?

Since it works right now for others I'm not eager to arbitrarily merge in new hashes without understanding if there's actually a problem in the first place. If Owncast works for people out of the box, and it does, then I'm having a hard time finding a problem to fix.

rAcHekLoS:
It is clear why this partly does not work, for the addresses /embed/chat/readwrite etc. no CSP header is set at all. Depending on the order in which the browser receives the data from the server, the header of the other files is evaluated or not.

I have set the header for the embed, please test this with the old hashes and then you see that not work.

gabek:
The lack of headers on the embeds explains why the tests (test/automated/browse) weren't throwing any issues with them, because they weren't being enforced. Thank you for catching that!

Though because they weren't being enforced at all for embeds, I'm not sure how you were getting CSP errors in the first place? But that's less important, seems like it's all doing the right thing with the headers.  Merging!

Thanks again!

PR #1481, opened 2021-10-19
strip any leading hash symbols from admin tag input using strings.TrimLeft

### LINKED ISSUE #1476: Strip hashtag prefix when saving new tags via the admin
It should be saved without the symbol.

image

nthall:
# Description

This is a small change to admin behavior, removing any/all hashtag symbols at the start of a tag before it is saved.

Fixes #1476 

Note I'm fairly new to learning Go, so I apologize if this isn't idiomatic or anything.


jeyemwey:
Amazing, thanks for your help.

PR #1502, opened 2021-10-29
Optimise regexp usage

Jack073:
This PR optimises a couple of uses of RegExps throughout the go backend. In two cases it moves the compilation of the RegExp to the global scope and outside of the function definition so they only need to get compiled once at program startup as opposed to everytime the function gets called. The third RegExp I've taken out and instead replaced with a simple function to parse it manually, in the benchmarks I ran (essentially using various JSON strings, valid and malformed) it generally runs atleast hundreds of times faster and has no additional allocations - I'll attach one set of results below (for one data input), but their relative results were pretty consistent. I also considered simply moving this regex compilation to the global scope as well, but the introduction of the new function instead still offered much better performance.

[No issue attached - I saw this and liked the project, so I wanted to become a little familiar with the codebase, and saw these contributions I could make which would have a positive effect]

```
BenchmarkRegExpRecompile        176647              6544 ns/op            1929 B/op         23 allocs/op
BenchmarkRegExpCompileOnce      471300              2516 ns/op             192 B/op          2 allocs/op
BenchmarkNewParser            16669629                67.69 ns/op            0 B/op          0 allocs/op
```
(first row is original, second row is moving compilation to global scope, third row is using the new function)

### DIFF
--- a/core/chat/events/events.go
+++ b/core/chat/events/events.go
@@ -124,6 +124,11 @@ func RenderMarkdown(raw string) string {
 	return buf.String()
 }
 
+var (
+	_sanitizeReSrcMatch      = regexp.MustCompile(`(?i)^/img/emoji`)
+	_sanitizeReAltTitleMatch = regexp.MustCompile(`:\S+:`)
+)
+
 func sanitize(raw string) string {
 	p := bluemonday.StrictPolicy()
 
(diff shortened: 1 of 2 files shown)


gabek:
That's awesome, thank you for investigating!

It's a good catch, as recompiling that regex on every chat message that needs to be sanitized doesn't make any sense, so that's great to move that out of there. On any kind of chat load this change will make a real difference for sure.

Now that you've pointed it out I'm wondering if creating a single sanitizer object would be helpful at all, since that's being recreated on every pass as well, and I'm guessing internally there's some regex being compiled. Not something that needs to be done now, but you got me thinking!

As for that metadata parsing, that only happens once on every inbound stream, so it's run very rarely. So while I totally appreciate you rewriting that in a specific implementation, and it's clear that it is faster, I'm not sure the tradeoff in this case is there of having custom string parsing for a questionable win.  What do you think?

Jack073:
Ah okay that's pretty fair then, I'd probably agree then that if it's run rarely its probably better for maintainability and readability sake to just stick with the regex, so I'll revert that one in a sec. As for the other bit though (assuming I'm understanding you correctly) these are the only 3 uses of regex I can see in the backend, well, bar config files.

gabek:
Ok, awesome!  And yeah, I just meant we're using [bluemonday](https://github.com/microcosm-cc/bluemonday) for sanitizing and there's probably some regex in there.  I did a quick search and there's [a few](https://github.com/microcosm-cc/bluemonday/search?q=regexp) that look to get compiled when creating a new policy. But that's something that can be addressed another time.

Thanks again for your suggestions and insight. Feel free to chime in on any other thoughts you have elsewhere in the codebase, I'm sure there's lots of opportunities for optimizations such as this :)

PR #1505, opened 2021-10-30
:zap: execute webhooks in parallel (and :bug: update webhooks.last_used correctly)

### LINKED ISSUE #1504: Slow webhooks servers might cause delays
The webhooks are run sequentially for one event, meaning that a chat message was first send to the first applicable webhook, and only to the next when it has closed the call.

This might create significant delay if one webhook recipient is slow in its response (e.g. if they're doing expensive operations on their end).

jeyemwey:
Change how webhook requests are send out.

Previously, this was done sequentially, as described in #1504.

With the change, each call of `SendEventToWebhooks(payload WebhookEvent)` creates `n` go routines for the `n` webhook URLs which are then run in parallel / non-blocking. After the HTTP request went through and the last-used cell is updated, the go routine is destructed.

PS: While looking into it, I found that the `SetWebhookAsUsed` call does not work as expected. I have attempted to fix it, in the second commit (0b4705ca). Since the field is not used anywhere (from what I can see), I didn't want to create new PR.

### DIFF
--- a/core/data/webhooks.go
+++ b/core/data/webhooks.go
@@ -102,15 +102,15 @@ func GetWebhooksForEvent(event models.EventType) []models.Webhook {
 	webhooks := make([]models.Webhook, 0)
 
 	var query = `SELECT * FROM (
-		WITH RECURSIVE split(url, event, rest) AS (
-		  SELECT url, '', events || ',' FROM webhooks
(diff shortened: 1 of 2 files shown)


gabek:
LGTM! It does what it says.

PR #1506, opened 2021-10-31
Fix build for Windows (#1377)

### LINKED ISSUE #1377: Fix Windows support
Re: https://github.com/owncast/owncast/pull/1376

We're using Unix system calls in https://github.com/owncast/owncast/blob/develop/core/chat/concurrentConnections.go#L13-L19 and https://github.com/owncast/owncast/blob/04bb97bffc17a29d5b3388a84df2d48f7ec087e3/core/chat/utils.go#L7  

Just like we have a FreeBSD specific version of his code, we may need to create one that's Windows-only that's simply a no-op.

Jack073:
Fixes #1377. This PR adds specific build constraints for windows, including a no-op setSystemConcurrentConnectionLimit and a getMaximumConcurrentConnectionLimit which simply returns a constant for the closest value I could find for windows. This also adds the new style build constraints (added in go 1.17) but also keeps the original version.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       matrix:
         go-version: [1.15.x, 1.16.x]
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
--- a/core/chat/concurrentConnections.go
+++ b/core/chat/concurrentConnections.go
@@ -1,5 +1,6 @@
 // nolint:goimports
-// +build !freebsd
+//go:build !freebsd && !windows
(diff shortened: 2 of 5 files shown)


gabek:
Awesome!  Would you mind adding `windows-latest` to this line here https://github.com/owncast/owncast/blob/develop/.github/workflows/test.yaml#L9 so it runs the test build against Windows?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast">owncast/test.yaml at develop · owncast/owncast</a></strong></div><div>Take control over your live stream video by running it yourself.  Streaming + chat out of the box. - owncast/test.yaml at develop · owncast/owncast</div></blockquote>

Jack073:
Ah yep, no problem. Doing it now

gabek:
This is great, thanks for doing this!  I don't actively use Windows myself, but as far as I know this was the last thing that needed to be resolved to fix the build.  Thanks again! :100:

PR #1523, opened 2021-11-12
Auto updater APIs

### LINKED ISSUE #924: Feature request: Auto update
It would help non-technical people if the server auto updated.

If [automatic TLS](https://github.com/owncast/owncast/issues/922), [single-command server setup](https://github.com/owncast/owncast/issues/923), and this feature request are implemented, Owncast would become a [Small Web](https://small-tech.org/research-and-development/) app that could easily be installed and maintained by everyday people without any technical knowledge whatsoever (imagine: you hit a web site, choose a domain, and boom, your Owncast server is up and running and you never have to maintain it). 

In [Site.js](https://sitejs.org), this is implemented by a simple GET request that checks the latest version of the single binary, downloads it if necessary, and restarts the systemd service.

Something similar can be implemented for Owncast.

gabek:
These are the APIs for

1. Determining if an in-place update or auto-restart is available.
2. Firing the updater.
3. Force-quitting to restart the process.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,7 @@
         "Mbps",
         "nolint",
         "Owncast",
+        "ppid",
         "preact",
         "RTMP",
         "rtmpserverport",
--- a/build/release/build.sh
+++ b/build/release/build.sh
@@ -71,7 +71,7 @@ build() {
 
   pushd dist/${NAME} >> /dev/null
 
-  CGO_ENABLED=1 ~/go/bin/xgo --branch ${GIT_BRANCH} -ldflags "-s -w -X github.com/owncast/owncast/config.GitCommit=${GIT_COMMIT} -X github.com/owncast/owncast/config.BuildVersion=${VERSION} -X github.com/owncast/owncast/config.BuildPlatform=${NAME}" -targets "${OS}/${ARCH}" github.com/owncast/owncast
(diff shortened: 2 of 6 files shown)


gabek:
https://github.com/owncast/owncast/pull/1523/files#diff-b7add41520314d7ff6812fbd9cee90b3607ea665eecf7c29949e0e296fff4a52R160 is currently how it's checking against systemd to see if a PID is managed by systemd.

Still need to handle errors and failure states.  Currently the frontend doesn't know if the installer step failed.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast/pull/1523">Auto updater APIs by gabek · Pull Request #1523 · owncast/owncast</a></strong></div><div>These are the APIs for

Determining if an in-place update or auto-restart is available.
Firing the updater.
Force-quitting to restart the process.</div></blockquote>

gabek:
@muesli Do you have any thoughts on any more reliable/less hacky ways of knowing if a specific PID is managed by systemd?

muesli:
@gabek Could check the parent's ID (`getppid`): for systemd it should be `1`. Alternatively, since systemd 232, there's an environment variable called `INVOCATION_ID`. Obviously both approaches aren't entirely fool-proof, but certainly a bit less hacky than the current method.

gabek:
Those are awesome pointers, thank you!

PR #1525, opened 2021-11-12
use worker pool to limit webhooks to 10 concurrent http executions (#1510)

### LINKED ISSUE #1510: Build outbound webhook execution queue
See https://github.com/owncast/owncast/issues/1504 for some discussion.

Right now each event fires off its own concurrent webhooks. In the case of an abundance of events this could cause both a small DDoS to the webhook endpoint, and cause a lot of concurrent outbound requests on our side.

The solution is to queue them up and rate limit the number of webhooks that can be in flight.

jeyemwey:
Using a simple worker pool model.

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -13,8 +13,6 @@ const (
 	DataDirectory = "data"
 	// EmojiDir is relative to the webroot.
 	EmojiDir = "/img/emoji"
-	// WebhookWorkerPoolSize defines the number of concurrent HTTP webhook requests.
-	WebhookWorkerPoolSize = 10
 )
 
 var (
--- a/config/constants.go
+++ b/config/constants.go
@@ -13,6 +13,8 @@ const (
 	DataDirectory = "data"
 	// EmojiDir is relative to the webroot.
 	EmojiDir = "/img/emoji"
+	// WebhookWorkerPoolSize defines the number of concurrent HTTP webhook requests.
+	WebhookWorkerPoolSize = 10
 )
 
 var (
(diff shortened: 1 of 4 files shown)


config/constants.go
```diff
@@ -13,6 +13,8 @@ const (
 	DataDirectory = "data"
 	// EmojiDir is relative to the webroot.
 	EmojiDir = "/img/emoji"
+	// WebhookWorkerPoolSize defines the number of concurrent HTTP webhook requests.
+	WebhookWorkerPoolSize = 10
```
gabek: Since this value is an internal detail to the worker pool itself and not to the application as a whole, maybe just move it to that file.

core/webhooks/webhooks.go
```diff
@@ -1,17 +1,11 @@
 package webhooks
 
 import (
-	"bytes"
-	"encoding/json"
-	"net/http"
-	"time"
+  "time"
```
gabek: https://pkg.go.dev/golang.org/x/tools/cmd/goimports?utm_source=godoc

Might be a whitespace thing that goimports is complaining about?

gabek:
Thank you! This is great, exactly what I had in mind. It will certainly make me feel better knowing we are rate limiting events that are directly connected to chat actions.

At some point we should have a test suite for webhooks so changes can run some sanity checks.  But I tested manually and all works as expected! :100:

jeyemwey:
There we go, a spoonful of automated formating helps the linter errors go down.

gabek:
Thank you for doing this!

PR #1539, opened 2021-11-17
Specify required Go version to compile from source

mattdsteele:
# Description

A small documentation change I noticed when building from source today; I was running into issues compiling on a slightly older VPS that had Go 1.15 installed, and it wasn't immediately clear to me why (the `io/fs` module in particular was missing).


gabek:
Nice, thanks!

PR #1546, opened 2021-11-19
refactor: move from io/ioutil to io and os packages

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since owncast has been upgraded to Go 1.17 (#1464), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

While the `io/ioutil` package is still usable, it is best for us to move to the new definitions now to ensure code consistency in the future.

### DIFF
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -4,9 +4,9 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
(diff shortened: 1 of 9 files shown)


gabek:
It seems like the `io` and `os` versions also exist in 1.16? The tests aren't failing running against 1.16 so that seems to be the case.  If this deprecates using 1.16 completely then we should update the readme and remove the 1.16 tests.

Juneezee:
> It seems like the `io` and `os` versions also exist in 1.16? The tests aren't failing running against 1.16 so that seems to be the case. If this deprecates using 1.16 completely then we should update the readme and remove the 1.16 tests.

Yes, the new definitions in `io` and `os` packages exist since Go 1.16. So 1.16 tests will still work without problems

gabek:
Awesome! Yeah, I see no reason not to standardize on moving away from `ioutil`. Thanks for the suggestion, the change and the clarification!

Issue #1497, opened 2021-10-27
List of configurable username suggestions


jeyemwey:
The Docker-style terminology might not suite everyone, especially in a more serious context. That's why some admins are asking for an option to have a custom list of predefined usernames which should be used instead of the Docker phrases.

INFO: The phrase generator is here: https://github.com/owncast/owncast/blob/develop/utils/phraseGenerator.go

INFO: The "old" way (custom prefix and random number) would be great too.

PR #1644, opened 2022-01-04
feat: add custom username suggestions pool

jeyemwey:
The reasoning was described in https://github.com/owncast/owncast/issues/1497. Basically, some audiences might find it disturbing to be called after some random scientists, so we should give server admins the option to create a pool of suggested usernames.

With this PR, we're creating the option. It runs alongside the "forbidden usernames" feature and most code was inherited from it.

To return to "the old way" of `user-$randInt(0,100)", an admin can add a large number of usernames to the pool.
The user can still modify the assigned username. The same username might be given to multiple people in the audience, if the sample pool size is too small.

Changes to `owncast-admin` in this PR: https://github.com/owncast/owncast-admin/pull/415

### DIFF
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -603,6 +603,27 @@ func SetForbiddenUsernameList(w http.ResponseWriter, r *http.Request) {
 	controllers.WriteSimpleResponse(w, true, "forbidden username list updated")
 }
 
+// SetSuggestedUsernameList will set the list of suggested usernames that newly registered users are assigned if it isn't inferred otherwise (i.e. through a proxy).
+func SetSuggestedUsernameList(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


gabek:
While I understand the use case, I do fear that people might set one or two names here and then cause a real mess when everyone has the same name. I know that's their problem, but that result doesn't seem like something anybody would want.  But if this is what they asked for, then they can have it.  But I just hope people don't experiment with it and cause a mess and not understand why everyone has the same name.

gabek (approved):

gabek:
Can we make it so you need to have at least 5 or 10 names or something for the name pool to be active?

jeyemwey:
> Can we make it so you need to have at least 5 or 10 names or something for the name pool to be active?

Added a check. 10 is probably a better number.

PR #1663, opened 2022-01-13
Outbound live stream notifications

### LINKED ISSUE #1609: Internal 3rd party auto-poster
Add a new feature where a server admin could configure specific services that get notified when a stream goes live.  This could use the https://github.com/nikoksr/notify library that has built-in support for a handful of different services. The downside, however, is it pulls in SDKs from all of those services and I'm curious how much it would increase our shipped binary size.

Right now we encourage people to utilize our webhook functionality to do things like this, but having this support built-in would be a huge win for streamers, and possibly increase their viewership as they are empowered to post about their streams directly.

### LINKED ISSUE #1656: Web push notifications
I know everyone says no to these, but I'm thinking if we added a YouTube style "ring the bell to be notified" contextual action button people might say yes to being notified when the stream goes live, just like how people install the Owncast browser extension. Except this would be for a single stream, and for people who don't want, or isn't familiar with the browser extension.

### LINKED ISSUE #1685: Outbound email notifications
Email, while not a perfect way to contact people, exists, and it works for many.  Plus people get it anywhere they are, so my dream of "let me know when X starts streaming, even when I'm not at my computer" helps with this.

As a bonus, it would allow streamers to optionally send out newsletters or their schedule for the week.  Something I think would be very cool.

## Requirements

- We don't store the email addresses 
- A single newsletter provider with a free tier must be selected
- This provider must have a good API we can use
- This provider is GDPR compliant and handles opt-in and unsubscribe logic
- We add email support to the notify modal on the fronend

gabek:
This is a set of outbound notification features to alert people and places when a stream goes live.

While previously I felt strongly about "APIs over Features", nobody is building on the APIs, even though I had [previously already wrote sample code](https://owncast.online/thirdparty/samples/#notify-followers-via-email-or-sms-when-you-go-live) people could use. So now I'll build the features to fill in the gaps.

- Sends a notification to a Discord channel
- Uses browser push notifications to notify people, though it only seems to work on Chrome-based browsers
- Uses text messages to notify people

Obviously I can't ship this for a number of reasons including that I'd need a much nicer designed modal to convince people they want to get notified and we can't be storing people's phone numbers. Technically the Discord notifier could be shipped since it has no user-facing modal that needs to be designed.

It's possible I can find an alternative to Twilio that operates more like Mailchimp and is responsible for storing phone numbers so a feature like this can still work in the future, without Owncast storing anything personal.

image

owncast-browser-notifications

image


aaronpk:
FYI the prompts in Chrome and Firefox on mac do not include those options for notification lifetime.

image

image

gabek:
- [x] Figure out why Firefox browser push notifications don't work.
- [x] Design the modal to support email and/or push notifications.
- [x] Hide push support for Safari, but show for others as long as the browser says it supports them.

gabek:
The Mozilla server error when delivering Firefox subscriptions is as follows:

```
{"code": 401, "errno": 109, "error": "Unauthorized", "more_info": "http://autopush.readthedocs.io/en/latest/http.html#error-codes", "message": "Request did not validate missing authorization header: Key mismatch"}
```

Chrome works fine, so there's something different between the two.

gabek:
Updated modal.

image

image

gabek:
I've removed support for sending emails.

PR #1678, opened 2022-01-18
Fixed multiple short variable declaration for 'err' var.

goodtiding5:
# Fix multiple short variable declaration for err in core:Start() function

---

I believe we should have only one declaration within function scope, or we can't guarantee which err we will be using in logic checking.

In my build with golang 1.17, an error 'failed to create the initial offline state' is reported in core:Start().  It shouldn't happen as function core:createInitialOfflineState() will only return nil as the error.

Without the fix, owncast will stop with error.  With the fix, it will start correctly.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -33,19 +33,19 @@ var (
 )
 
 // Start starts up the core processing.
-func Start() error {
+func Start() (err error) {
 	resetDirectories()
 
 	data.PopulateDefaults()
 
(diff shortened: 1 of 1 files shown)


gabek:
> Without the fix, owncast will stop with error. With the fix, it will start correctly.

Could you clarify how Owncast is stopping with an error?  It's working for me.

PR #1681, opened 2022-01-19
Add support for established chat user mode. #1587

### LINKED ISSUE #1587: Troll Mode (like +m on IRC)
Was discussing with @YarmoM's and crew on his stream today about moderation ideas and this one came out of some brainstorming that I think is a neat idea.

The idea of something akin to an "allow list" of chatters when a troll comes in would reverse the moderation flow, a "Troll Mode", if you will. Instead of banning the bad people, you temporarily allow only the "good" ones.  And while having to manually mark those you trust in the chat is additional work, this would persist and be a one time thing per person. This can also be accomplished by moderators.

It would seen as a temporary measure, and not something expected to be on all the time in hopes that a few minutes of not being able to send messages would discourage a troll from wanting to spend more time there.

gabek:
When enabled it rejects chat messages from users who are newer than 15 minutes old.

https://github.com/owncast/owncast/issues/1587

### DIFF
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -1,6 +1,10 @@
 package config
 
-import "github.com/owncast/owncast/models"
+import (
+	"time"
+
+	"github.com/owncast/owncast/models"
+)
 
 // Defaults will hold default configuration values.
 type Defaults struct {
@@ -27,6 +31,8 @@ type Defaults struct {
 
 	FederationUsername      string
 	FederationGoLiveMessage string
+
+	ChatEstablishedUserModeTimeDuration time.Duration
 }
 
 // GetDefaults will return default configuration values.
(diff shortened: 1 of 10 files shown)


Semisol:
> When enabled it rejects chat messages from users who are newer than 15 minutes old.

What about a customizable one? Depending on what you do you may want longer.

Also, three seperate things that may be worth doing:
1. Override for specific users. Useful with longer times like specified above.
2. Put all messages sent in a queue. Spammy users will still think they are getting their messages through, while real users can get an override.
3. Related to the above one, a "shadowban" that pretends the messages came through and does not put them into the queue. May be useful for extending the useful lifespan of a block.

Last 2 ones can be detected, but it would still be helpful.

Semisol:
Also, why does this close #1587? This issue only adds a time limiter and not "only approved users". I would probably add an "override only" option to the customizable time requirement of "established" proposal in the above comment.

gabek:
Like with everything else I'm taking the route of one step at a time. I could build all of those specific implementations, but I believe this will work for most people. If it doesn't, then I'll take another feature off the queue and build that one with whatever I learn from this one.

Semisol:
> Like with everything else I'm taking the route of one step at a time. I could build all of those specific implementations, but I believe this will work for most people. If it doesn't, then I'll take another feature off the queue and build that one with whatever I learn from this one.

Understandable. These were just proposals. Still, adding a customizable time requirement may be helpful.

PR #1683, opened 2022-01-19
Sanitize display names. 

### LINKED ISSUE #1541: Strip out problematic and confusing characters from chat display names
Currently, nicknames can contain any character, such as unicode.
This cam cause problems, such as people using a banned nickname with ZWSPs or adding dots/etc at the end.
I recommend the character set `A-Z`, `a-z`, `0-9`, `-` and `_` for nicknames.

gabek:
This sanitizes/normalizes characters from display names from extended character sets.  It still allows for pretty liberal usage of characters, but the really annoying stuff that trolls would use to junk up chat messages is covered here.

Closes #1541

### DIFF
--- a/core/chat/events.go
+++ b/core/chat/events.go
@@ -10,6 +10,7 @@ import (
 	"github.com/owncast/owncast/core/data"
 	"github.com/owncast/owncast/core/user"
 	"github.com/owncast/owncast/core/webhooks"
+	"github.com/owncast/owncast/utils"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -20,7 +21,12 @@ func (s *Server) userNameChanged(eventData chatClientEvent) {
 		return
 	}
 
-	proposedUsername := receivedEvent.NewName
+	proposedUsername := utils.SanitizeString(receivedEvent.NewName)
+	if proposedUsername == "" {
+		// Empty string is rejected.
+		return
+	}
+
 	blocklist := data.GetForbiddenUsernameList()
 
 	for _, blockedName := range blocklist {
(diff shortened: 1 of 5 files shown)


jeyemwey:
Can we have more positive tests for „okayish“ usernames? Not everyone uses ascii characters and from looking at the code, I fear that it might be problematic if we don’t test them. For example:

* Thomas Müller (German Football Player)
* Łukasz (Polish version of Lucas)
* something with ß (I don’t have examples here)
* Tromsø (city in Norway)
* Asian characters (I don’t know how those work)

All of those are perfectly legitimate usernames, so we should add them to our test suite.

Semisol:
Does it remove Zalgo text? I don't know how it works, but if it doesn't, you may want to do overflow: hidden (if not done already).

jeyemwey:
I've added some more test cases which are legit and should not be replaced.

```
=== RUN   TestSanitizeString
    utils_test.go:55: Incorrect sanitization of string. Expected 'Thomas Müller', got 'Thomas Muller'.
    utils_test.go:55: Incorrect sanitization of string. Expected 'Łukasz', got 'ukasz'.
    utils_test.go:55: Incorrect sanitization of string. Expected 'Gießkanne', got 'Giekanne'.
    utils_test.go:55: Incorrect sanitization of string. Expected 'Tromsø', got 'Troms'.
    utils_test.go:55: Incorrect sanitization of string. Expected 'ру́сский алфави́т', got ' '.
    utils_test.go:55: Incorrect sanitization of string. Expected '日本語の文字', got ''.
--- FAIL: TestSanitizeString (0.00s)
```

While the tool replaces ü to u (which it shouldn't do in the first place), some characters are removed entirely which turns them into empty strings.

gabek:
> All of those are perfectly legitimate usernames
> While the tool replaces ü to u (which it shouldn't do in the first place)

It's doing what I expect it to do, to normalize extended characters to simple ones.  While @Semisol wanted an even more limited character set than that, I thought it was a good compromise.

It sounds like you'd prefer we leave it how it is now (anything goes). I'm fine with that, I don't actually care about this at all either way.  I'm just trying to respond to people's concerns.  I'm just going to close this. If people create shitty names then ban them.

PR #1732, opened 2022-02-12
:adhesive_bandage: Chat fed. event: acc name shld truncate (#1730)

### LINKED ISSUE #1730: Chat federated event: Account should truncate, not overflow
image

jeyemwey:
Fixes #1730. `flex` and `text-overflow` is weird, so the existing thing did not work as expected. Now it _should_ work:

image


gabek:
Nice, thanks for tracking that down!

PR #1754, opened 2022-02-28
Auto-select old username text when changing name.

le-fractal:
# Description

When a user want to change their username, they have to manually click the input field to edit it. This change makes it so that the old username in the input field is selected automatically and the user can just start typing their new chosen name.
I saw that the username features are planned to be changed, but I believe in the meantime this is a welcome addition.

This is my first PR on here, and the code seems to work. However I don't know anything about preact (I just found enough on Google to understand what I was doing).

I marked this as draft as I am not sure I'm not doing something terrible :)

### DIFF
--- a/webroot/js/components/chat/username.js
+++ b/webroot/js/components/chat/username.js
@@ -79,6 +79,12 @@ export default class UsernameForm extends Component {
     }
   }
 
+  componentDidUpdate({ }, { displayForm }) {
+    if (this.state.displayForm && !displayForm ) {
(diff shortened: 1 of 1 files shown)


gabek:
Great idea! I'll test it out today and we can get it in.

gabek:
Works as I'd expect!  Feel free to remove the draft status when you feel comfortable.

I added this to the 0.0.12 milestone because 0.0.11 will be released this week and it's best not to throw this in at the last minute, so feel free to take you time with it.

PR #1785, opened 2022-03-15
Stream performance metrics

### LINKED ISSUE #793: Playback metrics
It would be nice to surface, at a high level, the overall performance of a stream.  Some streamers don't know that they are having issues and showing re-buffering and variant switches, while not a perfect way to talk about overall streaming health, is a place to start.

gabek:
Accept, aggregate, and expose playback metrics to the admin. Closes #793


gabek:
I have no idea why the browser tests fail in this PR, they run fine locally and I can't seem to fix them. Just going to play it by ear and figure out in time why they only fail in Github Actions.

PR #1811, opened 2022-04-01
IndieAuth support

### LINKED ISSUE #1214: Add a small indicator to messages indicating the person who sent the message is authenticated
For example, when an admin authenticates against their Owncast server via IndieAuth, we should show something in the message to show that the user authenticated somewhere.  Possibly allow clicking on this indicator to send them to where they authenticated against.

### LINKED ISSUE #1272: Add an IndieAuth layer on top of the admin auth
## Goal

Authenticate the admin of an Owncast instance as a next step in supporting authentication for chat users.

## Examples

1. Gabe, who runs goth.land is chatting on goth.land but it is not clear he is the admin. He starts the the auth flow, types in "goth.land" and now Gabe is shown as the admin of goth.land, upgrading the "shadow account" to an authorized, "verified" account.
2. Gabe, who runs goth.land visits live.w0rmh0le.net. He has an auto-generated "shadow account" on live.w0rmh0le.net but it's not verified. He starts the the auth flow, types in "goth.land" and live.w0rmh0le.net verifies Gabe is verified by goth.land, and that "shadow account" is now an authorized, "verified" account.

History behind this exists in this discussion: https://github.com/owncast/owncast/discussions/440#discussioncomment-551174

### LINKED ISSUE #1273: Initial chat auth UI to support IndieAuth
@aaronpk wrote [this article](https://aaronparecki.com/2018/07/07/7/oauth-for-the-open-web) with this example login UI:

image

Since Owncast will support IndieAuth our UI can have some more specific verbiage specifically speaking to Owncast server admins, since that's the primary audience as most other people won't have access to IndieAuth.

See https://github.com/owncast/owncast/discussions/440#discussioncomment-532137 for the previous discussion about this topic.

Actual backend implementation is a prerequisite: https://github.com/owncast/owncast/issues/1272

### LINKED ISSUE #1810: Do not allow use of a display name owned by an authenticated user
Because everyone feels so strongly about display names being singularly used I'm making a compromise. 

1. If a display name is currently used by a user that has been authenticated, then that name is locked and nobody else can use it.
1. If a display name is currently used by a user that has not been authenticated, then that name can still be used.
1. If the authenticated user is banned, then the name is no longer locked.

> Why is it conditional?

Because if somebody is unauthenticated then they should still be able to use the name from different devices, clients, browsers, etc. It also doesn't lock away a name for eternity if somebody casually drops by and uses it for a few seconds and then never again. Somebody has to authenticate to use that name.

> What if some authenticated jerk uses that name that somebody else wanted?

Ban the jerk and free up the name.

gabek:
This is both the client and server-side support for IndieAuth.

https://indieauth.net/

- Adds new "chat settings" modal where IndieAuth is located.
- Authenticated users now display in chat with a checkmark.
- A display name in use by an authenticated user is now no longer available to other users.
- Adds IndieAuth server support so people can use Owncast itself as an IndieAuth server, primarily so people can authenticate at other Owncast servers, but also other IndieAuth sites as well.


auth/indieauth/client.go
```diff
@@ -0,0 +1,87 @@
...
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var response Response
+	if err := json.Unmarshal(body, &response); err != nil {
+		return nil, nil, err
+	}
+
```
aaronpk, line 73: This needs some more error handling in case it gets back an error from the IndieAuth server before continuing on to assume there is a successful response returned.

auth/indieauth/client.go
```diff
@@ -0,0 +1,87 @@
...
+	// Trim any trailing slash so we can accurately compare the two "me" values
+	meResponseVerifier := strings.TrimRight(response.Me, "/")
+	meRequestVerifier := strings.TrimRight(request.Me.String(), "/")
+
+	// What we sent and what we got back must match
+	if meRequestVerifier != meResponseVerifier {
+		return nil, nil, errors.New("indieauth response does not match the initial anticipated auth destination")
```
aaronpk, line 93: Returning a JSON error here isn't great UX since this ends up being visible JSON to the user.

auth/indieauth/client.go
```diff
@@ -0,0 +1,87 @@
...
+
+	data := url.Values{}
+	data.Set("grant_type", "authorization_code")
+	data.Set("code", code)
+	data.Set("client_id", request.ClientID)
+	data.Set("redirect_uri", request.Callback.String())
+	data.Set("code_verifier", request.CodeVerifier)
+
+	codeChallenge := createCodeChallenge(request.CodeVerifier)
+	if codeChallenge != request.CodeChallenge {
```
aaronpk: What's the purpose of checking this here? The client stored the code verifier itself, and generated the code challenge in the previous request, so the code challenge here isn't coming from an external source, so it seems like the check is redundant.

controllers/auth/indieauth/server.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		// Return a human readable, HTML page as an error. JSON is no use here.
+		return
+	}
+
+	// Redirect the client browser with the values we generated to continue
+	// the IndieAuth flow.
+	u, err := url.Parse(redirectURI)
+	if err != nil {
+		controllers.WriteSimpleResponse(w, false, err.Error())
```
aaronpk: This needs to return a proper OAuth error response with `{"error":"...","error_description":"..."}`. Currently it sends back an Owncast-style JSON response

aaronpk (commented):
I sent a PR in #1812 with some changes that I was able to do myself. The comments here are things I wasn't sure how you wanted to handle or didn't know enough Go to do myself.

aaronpk:
How would you feel about having this change the username to the `name` returned in the IndieAuth profile response after logging in? (Keeping the same uniqueness check of course). For owncast-to-owncast logins it would end up being the name of the server, and if they log in with their own website it would be the user's name. I'd imagine it would still send a username change event and everything so you'd see a notification like this in the chat:

> competent-newton is now known as Aaron Parecki

jamietanna:
Was there any reason we've hand-crafted the core part of the flow instead of using core go libraries for OAuth2?

E: I say this because recent changes to Indieauth allow us to use standard OAuth2 implementations now ☺ see https://gitlab.com/jamietanna/micropub-go/-/commit/76e6430bb90f6b74e418b5b311d4d3bdb3e472d8 as an example

auth/indieauth/helpers.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer r.Body.Close()
+
+	scrapedHTMLDocument, err := html.Parse(r.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to parse html at remote auth host")
+	}
+	authorizationEndpointTag := cascadia.MustCompile("link[rel=authorization_endpoint]").MatchAll(scrapedHTMLDocument)
```
jamietanna, line 85: It'd be great if we could support the Discovery endpoint, as although fairly new it's going to be fairly widespread before long https://indieauth.spec.indieweb.org/#indieauth-server-metadata

jamietanna: We could also outsource this logic to github.com/hawx/indieauth

auth/indieauth/helpers.go
```diff
@@ -0,0 +1,120 @@
...
+
+			return u, nil
+		}
+	}
+
+	return nil, fmt.Errorf("unable to find href value for authorization_endpoint")
+}
+
+func createCodeChallenge(codeVerifier string) string {
+	hasher := sha1.New() // nolint:gosec
```
jamietanna: :thinking: does this create a SHA1 hash? PKCE is commonly used with SHA256

auth/indieauth/response.go
```diff
@@ -0,0 +1,16 @@
...
+	Name  string `json:"name"`
+	URL   string `json:"url"`
+	Photo string `json:"photo"`
+}
+
+// Response the response returned when completing
+// the IndieAuth flow.
+type Response struct {
+	Me      string  `json:"me"`
+	Profile Profile `json:"profile"`
```
jamietanna: this may not always be returned - what happens if it's absent?

auth/indieauth/client.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, nil, errors.New("code challenge mismatch")
+	}
+
+	client := &http.Client{}
```
jamietanna, line 51: should we cache HTTP clients? Not sure if we do that elsewhere

auth/indieauth/helpers.go
gabek: I apologize in advance for my ignorance, I'm quite detached from the ecosystem and how all these things work together. 
 But I'm confused. There's a not-yet-formalized update to IndieAuth, but a library that hasn't been updated in two years supports it?

I also didn't use an OAuth library because I wasn't building an OAuth implementation, I specifically was following the documentation around IndieAuth.  And not the "new" IndieAuth, the "original" IndieAuth (again I don't really understand the difference, or if they're compatible, etc).

Admittedly I'm a little frustrated by the entire thing and if IndieAuth is going to be an always changing spec that is not going to be used by people who visit Owncast streams I should probably cut my losses because I don't want the cost of maintaining something that's constantly changing but nobody is using.

gabek:
All the feedback is completely appreciated, but this PR doesn't even pass tests and fails in every way. It's a draft PR that is looking less likely to be going in.

auth/indieauth/helpers.go
aaronpk: I fixed this in my PR to this branch

auth/indieauth/helpers.go
jamietanna: No I completely empathise with you and agree that it is a little frustrating!

Some of the recent changes we've made very recently have been to align much closer to OAuth2, so it's a much more straightforward implementation

And no you're right, that library doesn't yet support the new means, it was something I'm planning on contributing so we'd not need the logic in here either

gabek:
> How would you feel about having this change the username to the `name` returned in the IndieAuth profile response after logging in? (Keeping the same uniqueness check of course). For owncast-to-owncast logins it would end up being the name of the server, and if they log in with their own website it would be the user's name. I'd imagine it would still send a username change event and everything so you'd see a notification like this in the chat:
> 
> > competent-newton is now known as Aaron Parecki

This, and more, is something I did want to do. In fact my original vision was to set their display name, link their profile, and even put a little Owncast logo if what they're authenticating against is an Owncast server. But then I realized I'm making a lot of assumptions about what data they want to make public. While they may want to authenticate against a service, that doesn't mean they want their identity, as known by that service, to be made public. Even sharing that this person has an Owncast instance is making an assumption that they want others to know about it.

I could, of course, add more switches, knobs and prompts to ask people what they want to share, but I think that would just add confusion and make the whole flow feel like Facebook Connect, where Facebook would hand over data about a user when you used them for authentication purposes. So saying it's auth, and nothing more, is probably the most fair way to go about it.

aaronpk:
That's a fair point. The way this usually works with IndieAuth servers is the server will show a prompt to indicate what info will be shared with the site you're logging in to, just like you see in other OAuth servers. There's also scopes that the site can request to tell the server whether it wants to know their name or email in addition to just doing the authentication.

auth/indieauth/response.go
gabek: The profile isn't required. I might as well just remove it.

gabek: Oh sorry, misread, you're looking at our server response.

There's currently no checks for these kind of things put in. That still needs to happen. IndieAuth support (client or server) will not be available unless you've set the required configuration (URL mostly). Errors need to be surfaced in the case that it's not configured.

auth/indieauth/client.go
gabek: These requests will be rare, seems like that optimization wouldn't be needed. It's not something that's done elsewhere. Owncast doesn't do many outbound requests.

auth/indieauth/client.go
gabek: Since every implementation is unique, is it safe to rely on http status codes to determine there was an error as opposed to errors in the return payload? There's no way to know what the error structure will look like.

aaronpk: The error structure is OAuth 2's structure with "error" and "error_description" keys in the JSON response

gabek: I didn't realize that, I'm certainly not using that structure in my IndieAuth response. I'll make sure to do that, but if I didn't know it's possible other implementations didn't know either.

gabek: What should be returned as `error` vs `error_description`?

aaronpk: Yeah we chose to link out to the OAuth spec instead of copy all the error responses into the IndieAuth spec. Here's the list: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

controllers/auth/indieauth/server.go
gabek: hah! If I would have read your comment before today I would have known about the OAuth error response.

auth/indieauth/helpers.go
gabek: Filed an issue pointing to this thread to possibly work on the next version of IndieAuth if other servers start supporting it and people find the support within Owncast useful. https://github.com/owncast/owncast/issues/1837

auth/indieauth/client.go
gabek: Removed redundant check.

gabek:
I think it's in better shape now. Hopefully the above feedback is implemented as intended.  I'm guessing there's still some rough edges in real world use, however.

auth/indieauth/client.go
gabek: A simple, but plain text, error is surfaced now. I filed #1838 to have something nicer.

gabek:
I'm going to merge this into `develop` so daily development can live side by side with this and issues can be surfaced. If it all falls apart it's easy enough to turn off.

PR #1812, opened 2022-04-02
Fixes for IndieAuth client

aaronpk:
* Fixes frontend to not redirect unless the server returns a URL to redirect to
* Improve the error if the owncast server URL isn't set (this would have saved me some time hunting this down when I was testing)
* Fixes the IndieAuth client PKCE implementation


auth/indieauth/helpers.go
```diff
@@ -1,15 +1,15 @@
 package indieauth
 
 import (
-	"crypto/sha1" // nolint:gosec
+	"crypto/sha256" // nolint:gosec
```
jamietanna, line 4: Can probably remove the `nolint`?

gabek:
I went ahead and cherry-pick'ed your commits into [my branch](https://github.com/owncast/owncast/tree/gek/indieauth) so I could fix the linter errors and simplify fixing the conflicts.

PR #1831, opened 2022-04-15
Change CMD to ENTRYPOINT in Dockerfile

IonCannon218:
Hello,

This should be easier to allow docker (and docker-compose) users to optionally use the verbose, debug or other command line arguments without breaking the container start up. From my understanding, using ENTRYPOINT in the Dockerfile is more suitable since the `/app/owncast` start command within the container should not be overridden but can be appended to if desired.

I was trying to get `--enableVerboseLogging` and `--enableDebugOptions` to work with docker-compose `command:` but it would not work until I modified the Dockerfile since it was overriding the original command.


gabek:
Makes sense!  Thanks for the change!

PR #1833, opened 2022-04-15
Eliminate client close deadlocks.

funkyhippo:
# Description

Currently, a deadlock exists when the chat server tries to broadcast a message to a blocked client `send` channel. The existing behaviour calls the client's `close()` method, which also writes to the server's `unregister` channel. At this point, the receiver to the `unregister` channel also tries to acquire the server's mutex, which it cannot because the broadcast method is still holding it. The broadcast won't continue since the `unregister` channel is unbuffered, causing a deadlock.

## Reproducing

I've encountered the deadlock issue (which kills chat until we restart the Owncast instance) a couple times during my community's streams. This issue is reproducible with a modified `test/fakeChat.js` script that constantly creates and disconnects hundreds of chat clients.

## Changes

This change moves the client close into a goroutine so that client management can be reconciliated after the broadcast is complete.

I've also cleaned up the client ID set management by removing it from the broadcast method. This means that a client will only be truly unregistered when its ID is passed and handled in the `unregister` channel.

Finally, since the close is also asynchronous now, I've added a mutex to the client in order to prevent multiple closes (ie. the Go scheduler runs two broadcasts back to back; this will trigger two calls to a given client's `close()` method) or writing to a nil channel.

### DIFF
--- a/core/chat/chatclient.go
+++ b/core/chat/chatclient.go
@@ -3,6 +3,7 @@ package chat
 import (
 	"bytes"
 	"encoding/json"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -17,6 +18,7 @@ import (
 
 // Client represents a single chat client.
 type Client struct {
+	mu          sync.RWMutex
 	id          uint
 	accessToken string
 	conn        *websocket.Conn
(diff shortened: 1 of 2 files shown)


gabek:
This is a fantastic find and fix! I haven't seen this occur, but I've probably just never run into quick connects/disconnects, so I really appreciate you discovering this and looking into it. :1st_place_medal: 

I have a WIP load test that I've been wanting to finish that I bet could have detected this, so this will encourage me to get that working and into the project.

PR #1834, opened 2022-04-16
Fixed Firefox positioning caret on the right in chat input

### LINKED ISSUE #1561: Cursor when composing chat message appears on the right-hand side of the text field in Firefox
This is probably a bug, small but somewhat irritating.
The cursor initially appears on the right-hand side of the text field rather than on the left side.
cursor_bug
Once you start typing it behaves as expected.

t1enne:
# Description

Fixes #1561 

---

Firefox places caret on the right in contenteditable elements. 
[Bug 904846] (https://bugzilla.mozilla.org/show_bug.cgi?id=904846)

Simply changing position to absolute on the ::before pseudo-element fixes it.

Tested on Chromium and Firefox.

### DIFF
--- a/webroot/styles/chat.css
+++ b/webroot/styles/chat.css
@@ -64,6 +64,7 @@
 #message-input:empty:before {
   content: attr(placeholderText);
   pointer-events: none;
+  position: absolute; /* Fixes firefox positioning caret on the right */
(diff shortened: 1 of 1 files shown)


gabek:
This is awesome, thank you so much for tracking down this fix!

PR #1844, opened 2022-04-20
Chat menu restyle

### LINKED ISSUE #1648: Replace hide chat toggle and username change form with chat actions dropdown
In addressing #1339, people don't know how to change their name.

Removing these two actions in the header and instead replacing with a single drop down menu would enable us to be clearer about specific actions.

This "viewer settings" menu would only have "Hide chat" and "Change name" to begin with, but this is also where we could put future features and actions as needed, such as IndieAuth login.

It clarifies these actions because when people see a menu button they know they can click it, where currently they don't know they should click their username to change it, or the chat toggle to hide chat.

t1enne:
Restyled the chat menu. It now has a popout with two options:

- change username
- toggle chat

### DIFF
--- a/webroot/js/app.js
+++ b/webroot/js/app.js
@@ -9,8 +9,8 @@ import SocialIconsList from './components/platform-logos-list.js';
 import UsernameForm from './components/chat/username.js';
 import VideoPoster from './components/video-poster.js';
 import Followers from './components/federation/followers.js';
-
 import Chat from './components/chat/chat.js';
+import { ChatMenu } from './components/chat/chat-menu.js'
 import Websocket, {
   CALLBACKS,
   SOCKET_MESSAGE_TYPES,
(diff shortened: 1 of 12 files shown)


webroot/js/app.js
```diff
@@ -941,36 +941,22 @@ export default class App extends Component {
                 >${streamOnline && streamTitle ? streamTitle : name}</span
               >
             </h1>
-            <div
+            <!-- <div
               id="user-options-container"
               class="flex flex-row justify-end items-center flex-no-wrap"
```
gabek, line 946: Remove commented out code.

webroot/js/components/chat/chat-menu.js
```diff
@@ -0,0 +1,126 @@
...
+//       <${UsernameForm}
+//         username=${username}
+//         isModerator=${isModerator}
+//         onUsernameChange=${onUsernameChange}
+//         onFocus=${onFocus}
+//         onBlur=${onBlur}
+//       />
+//     </div>
+//   `
+// }
```
gabek: Remove commented out code.

webroot/js/components/chat/username.js
```diff
@@ -98,65 +100,61 @@ export default class UsernameForm extends Component {
...
-    `;
-    const userIcon = html`
-      <img src="/img/user-icon.svg" class="user-icon-flag" />
-    `;
+    // const moderatorFlag = html`
+    //   <img src="/img/moderator-nobackground.svg" class="moderator-flag" />
+    // `;
+    // const userIcon = html`
+    //   <img src="/img/user-icon.svg" class="user-icon-flag" />
+    // `;
```
gabek: Remove commented out code.

webroot/styles/app.css
```diff
@@ -310,12 +327,12 @@ header {
...
-    width: 75vw;
+    /* width: 75vw; */
   }
   #user-info-display {
-    max-width: 30vw;
-    overflow: hidden;
-    text-overflow: ellipsis;
+    max-width: none;
+    /* overflow: hidden; */
+    /* text-overflow: ellipsis; */
```
gabek: Commented code.

webroot/js/components/chat/username.js
```diff
@@ -98,65 +100,61 @@ export default class UsernameForm extends Component {
...
-            type="button"
-            class="bg-gray-900 hover:bg-gray-800 py-1 px-2 mx-1 rounded cursor-pointer user-btn text-white text-xs uppercase text-opacity-50"
-            title="cancel"
-          >
-            X
-          </button>
+          <div class="username-buttons-wrapper flex ml-2">
+            <button
+              id="button-update-username"
+              onClick=${this.handleUpdateUsername}
```
gabek, line 136: This is probably a side effect of how name changes worked before, but it seems like if you don't change the name in the text field, but you hit the update username submit button then nothing happens. Maybe if you hit update without changing the name it should just cancel.

gabek (commented):
This looks and works great! Thank you so much for working on this.

webroot/js/components/chat/username.js
t1enne: Now it cancels when you update without changing the username!

gabek: Nice! :100:

gabek:
What do you think about reducing the font size a bit? It looks overwhelmingly large compared to the toned down server name on the left.

image

t1enne:
You are def right, sorry for missing that! so font-size now is smaller (text-sm). I also removed a few margins to match the server name

gabek:
Hurray! Let's get this in.

PR #1845, opened 2022-04-21
fixed chat scroll breaking on single col

### LINKED ISSUE #1830: Issue: chat unscrollabe on some resize operations
Hi there, noticed an issue (using Google Chrome, latest version, tested on MacOS and Windows):

If you shrink the browser's width enough to change layouts (from the horizontal layout, to the vertical layout), then expand it a little bit (so it stays in the vertical layout) - the chat becomes unscrollable.

Sometimes it's hard to reproduce. It seems like if the messages get reflowed while widening, everything is fine. But if you can widen just a bit and not have the reflow happen, you can no longer scroll with the mouse scroll wheel.

t1enne:
Some css being applied on smaller screens broke the scroll. Tested on Chromium, Linux 

Fixes  #1830

### DIFF
--- a/webroot/styles/app.css
+++ b/webroot/styles/app.css
@@ -214,10 +214,7 @@ header {
   z-index: 40;
 }
 .single-col #chat-container {
-  position: fixed;
-  bottom: 0;
   width: 100%;
-  height: 100%;
 }
 
 .single-col.chat #video-container,


gabek:
This is working great for me with the two different cases:

- Resizing the window
- Initially narrow window
- Mobile (tested on iOS)

PR #1846, opened 2022-04-22
Fediverse-based authentication

### LINKED ISSUE #1774: Chat auth using the Fediverse
I've never seen a way for people to use their Fediverse identity to create a new account somewhere.  It's just not what it's built for. But I think I came up with an idea.  I'm not sure if it's a good one. I'm usually a proponent of using the right tool for the job, and this isn't really that. But it might be something.

By abusing the remote follow flow, we can have a user verify their identity on a remote Fediverse server.

1. User says they want to authenticate with Owncast chat and types in their remote Fediverse account.
2. Owncast creates a temporary virtual ActivityPub entity with inbox attached to this chat user.
3. Owncast starts the remote follow flow, redirecting them to their Fediverse server.
4. Remote follow flow is successful and the remote Fediverse server sends a "Follow" event to Owncast, for this virtual inbox.
5. Owncast accepts this follow request and makes the connection between the remote Fediverse identity and the local chat identity.

If the user wants to login again, a new virtual inbox would get created and the user would "follow" this completely new virtual entity, and create another connection.

The downside is this Fediverse user would then have fake entities listed as things they follow and that's pretty annoying.  But maybe Owncast can send a "Delete" event to the remote Fediverse server after this flow is complete so this fake entity can be removed.

gabek:
Enable using your Fediverse to authenticate your chat account.


gabek:
Sure, let's give it a spin.

PR #1855, opened 2022-04-22
changed followers tab to 24 per page

### LINKED ISSUE #1823: Make followers tab fit the available space (chat vs no chat)
When chat is shown it reduces the number of columns that fit on the followers tab, meaning the final number is an odd number of followers and looks dumb.

image

We can make the followers section always show the same number, but make it fill the space.

So maybe _always_ show 15, (5 rows of 3 columns) but when chat isn't active make each column wider.

t1enne:
Changed the limit for the followers tab from 16 to 24. Changed to css to responsive grid

# Description

Fixes #1823

### DIFF
--- a/webroot/js/components/federation/followers.js
+++ b/webroot/js/components/federation/followers.js
@@ -24,7 +24,7 @@ export default class FollowerList extends Component {
 
   async getFollowers() {
     const { currentPage } = this.state;
-    const limit = 16;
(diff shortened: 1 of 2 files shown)


gabek:
The empty state looks a little weird now since it's being forced into a single column.

image

gabek:
Looks and works good for me! :+1:

PR #1872, opened 2022-04-23
Add terminations to error repsonses in API.

funkyhippo:
# Description

Certain API responses do not fully terminate when it hits an error. This results in a malformed JSON response that includes both the failure JSON string and the success JSON string.

image

## Changes

This change adds termination on error responses from the API. Without the termination, the error will be written to the response, as well as the success (or further error messages).


gabek:
This is great, thanks for discovering these issues with some of these lesser-used pieces of the puzzle.

funkyhippo:
> This is great, thanks for discovering these issues with some of these lesser-used pieces of the puzzle.

No worries, thanks for your hard work on the software itself. 😄

PR #1909, opened 2022-05-16
Clarify the message for registered usernames

MFTabriz:
I think this explains situation a bit better. Feel free to edit the message.


gabek:
Works for me!

PR #1910, opened 2022-05-16
simpler explanation for registering username

MFTabriz:
I think this should be simpler for new users. [Feel free to edit and improve]

### DIFF
--- a/webroot/js/components/auth-indieauth.js
+++ b/webroot/js/components/auth-indieauth.js
@@ -73,9 +73,9 @@ export default class IndieAuthForm extends Component {
     const loaderStyle = loading ? 'flex' : 'none';
 
     const message = !authenticated
-      ? html`Use your own domain to authenticate ${' '}
+      ? html`Use your own domain to register ${' '}
           <span class="font-bold">${username}</span> or login as a previously
-          ${' '} authenticated chat user using IndieAuth.`
+          ${' '} registered chat user using IndieAuth.`
       : html`<span
           ><b>You are already authenticated</b>. However, you can add other
           domains or log in as a different user.</span


gabek:
While "Authenticate" isn't great, "Register" in this context makes it sound like you're being asked to create an account. Really what you're doing is more of a "verify" of an existing identity or "logging in" to restore it. But I didn't use "verify" because Twitter stole "Verified" and people will think it means the same thing it means on Twitter (it's something you've earned because you're famous). And I didn't say "log in" because that only makes sense if you've already added a level of auth to your chat account and now you're re-authenticating to restore that identity. 

I'm totally fine with clarifying it, but register doesn't sound like the right word here if people are used to register meaning something else.

MFTabriz:
@gabek With the first authentication, aren’t we linking the domain to the currently chosen username, or am I missing something here?

MFTabriz:
What about "Connect your own domain to {username} on this owncast server or login as a previously connected chat user with IndieAuth."?

gabek:
> @gabek With the first authentication, aren’t we linking the domain to the currently chosen username, or am I missing something here?

I wouldn't say linking, since there's nothing about that's used for functionality. It's just a form of authentication to reproducibly identify yourself.

> What about "Connect your own domain to {username} on this owncast server or login as a previously connected chat user with IndieAuth."?

"Connect" makes it sound like there's some long-term functionality that you're signing up for, and not just using it as a tool to identify yourself. I think of "Facebook Connect" personally, where you perform this action once and mysterious actions can happen at any time.

Not sure if it helps, but @aaronpk's wording on IndieAuth.net is:
> You can use IndieAuth to have your users authenticate with their own URL.

Maybe "Log in" is the easiest terminology to understand that doesn't imply additional functionality?

MFTabriz:
I totally agree what what’s happening with IndieAuth is “authentication”. The authentication only verifies that a user is represented by a URL. But on the top of that, [when it's done for the first time] owncast uses this authentication to reserve a chat username. This process of associating a chat username with a URL must be somehow explained and that's what I was trying to explain by "registration".

@aaronpk do you have any suggestion?

gabek:
I just realized you're probably seeing the display name locking as the primary feature here, where I wasn't thinking about that being a feature at all. It's a side effect of being an authenticated user that the name is locked, not the primary purpose. I actually only added that at the last minute and that wasn't originally what I was going for. The primary purpose is to verify and restore your identity in a reproducable way. If you were a person before, you can be that person again. If you were a moderator before, you can still be a moderator on another device, etc.

MFTabriz:
I see your point. But for many users who have no extra permissions, having their username reserved/kept/registered is currently the only feature and IMO a very important one. This is the main reason why jnktn.tv is using a 3rd-party chat app.

gabek:
It's true, for most users I don't think they'll care about auth, and I think that's great. I've always wanted chat to be simple. It's why the functionality is being shoved in a menu away from people. This is mostly useful for moderators or people who want to restore their identity between browsers/devices.

A display name is simply a display name, I never focused on making sure your display name is a distinct identity. Sure, now with authentication there can only be one "Gabe", but there could be a "gabe" and a "Gabe!". The name is a dynamic identifier and I don't want to mistake a chat user's intrinsic identity with a name that can be changed and faked at any point.  To say your name is "registered" makes a claim for something different than what we're actually offering. We're offering restoration of identity. A display name is just one dynamic piece of that identity that is never guaranteed to be static or visually unique.

MFTabriz:
I know the discussion is not related to this PR anymore but then I think with following changes in the owncast we can offer a basic functionality for registering usernames:

- treating the lower and upper cases the same (if "Gabe" is an authenticated username, there can't be any other "gabe"/"GaBe"/...)
- showing a warning to the authenticated users when they try to change their username that the old username will be available for the others to use

What do you think?

gabek:
That kind of goes against my original point that I can't fight every battle and I don't want to make promises I can't keep.  The same thing happened with IP bans. I finally caved and said I would add IP bans, but then somebody immediately reported Owncast for having a security vulnerability because if somebody fakes their IP address the IP ban is no longer valid. So how far do I go?

Display names are no different. Battling trolls trying to come up with similar looking display names just to get around "unique" names is way worse than telling people Owncast doesn't have logins and there are no usernames. Your display name is just an outside representation of what you want to call yourself. It's like a tshirt. Wear whatever shirt you want, it doesn't change who you are.

Right now Owncast display names are similar to IRC nicknames. Even _more_ secure than IRC nicknames, now with auth, since a specific name gets locked down.

I understand this is annoying because everyone wants all the things, but I can't give everyone all the things.

MFTabriz:
I understand the hassle of dealing with people’s expectation. You know that I will help to improve the owncast as much as I can and I don’t expect you to spend more time and effort to make it usable for me. You have already done more than anyone can expect from you.

But for example as the name implies, IP ban is just what it claims: “IP ban”. It just bans IPs. Not devices, and definitely not users (people). Expecting the IP ban to do something else (ban people) is preposterous! 

Display names are the same. People can register similar-looking usernames on any website. Owncast being singled out for letting people use similar names is unfair. IMO if someone really want to control the user authentication, they should for example limit the authentication to a specific domain, or put manual verification by the owncast admin on the top of that.

By the way, in IRC, Nickserv is “registering” the usernames. https://libera.chat/guides/registration 

Finally, for me the owncast is just an excuse to collaborate with great people like you. Stay awesome as you are.


<blockquote><div> Libera Chat</div><div><strong><a href="https://libera.chat/guides/registration">Nickname Registration</a></strong></div><div>Your nick is how people on Libera.Chat know you. If you register it, you’ll be able to use the same nick over and over. If you don’t register, someone else may end up registering the nick you want. If you register and use the same nick, people will begin to know you by reputation.</div></blockquote>

PR #1922, opened 2022-05-23
Add “services” and “metadata” to NodeInfo/2.0

MFTabriz:
“services” and “metadata” are both required by the [NodeInfo schema version 2.0](http://nodeinfo.diaspora.software/ns/schema/2.0).

### DIFF
--- a/activitypub/controllers/nodeinfo.go
+++ b/activitypub/controllers/nodeinfo.go
@@ -59,6 +59,9 @@ func NodeInfoController(w http.ResponseWriter, r *http.Request) {
 
 // NodeInfoV2Controller returns the V2 node info response.
 func NodeInfoV2Controller(w http.ResponseWriter, r *http.Request) {
+	type metadata struct {
+		ChatEnabled bool `json:"chat_enabled"`
+	}
 	type services struct {
 		Outbound []string `json:"outbound"`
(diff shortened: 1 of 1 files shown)


gabek:
I would be hesitant to say that we offer a public chat server because that's not really true. It doesn't use any open protocols and you can't publicly connect to it.  However Owncast video does use open standards and anybody can consume it, so if you feel like we should expose public services maybe `video` is what should be in there?

gabek:
Just curious what you're using this for?

gabek:
I just looked at the Nodeinfo protocol and it looks like `services` is an enum with the following options:

inbound: atom1.0 gnusocial imap pnut pop3 pumpio rss2.0 twitter
outbound: atom1.0 blogger buddycloud diaspora dreamwidth drupal facebook friendica gnusocial google insanejournal libertree linkedin livejournal mediagoblin myspace pinterest pnut posterous pumpio redmatrix rss2.0 smtp tent tumblr twitter wordpress xmpp

So maybe it has to be an empty array for Owncast?

gabek:
Oh nevermind that's exactly what you did, I mixed up `metadata` and `services` in my head. Thanks!  Since `metadata` seems to be open-ended that bool is fine, because why not.

PR #1934, opened 2022-05-27
manifest - Add fullscreen mode/change name

ErAzOr2k:
-Change name to "Owncast"
-Open in fullscreen mode


gabek:
Seems reasonable, thanks!

PR #1944, opened 2022-06-11
Change .editorconfig to reflect current state of the code

128f:
The code appears to be using tabs, but the `.editorconfig` file says `spaces`, this causes tools respecting `.editorconfig` to do the wrong thing.


gabek:
That's a nice catch, thank you!

PR #1945, opened 2022-06-11
Use a custom http client when connecting to aws

### LINKED ISSUE #1940: Update AWS SDK to use a custom HTTP client for S3
The AWS SDK uses the default HTTP client by default. They suggest, to increase performance:

- Create a custom client
- With a timeout
- Increase the re-usable connections from the default of 2 to 10+.

image

128f:
fixes #1940 

~I chose to use only 5 idle connections because as far as I can tell, this would be a single client connecting to aws to write video files. More than happy to be corrected on this.~

Here is some reading that has changed my mind on this, I will change the PR accordingly:

* https://github.com/OJ/gobuster/issues/127
* https://gitlab.com/gitlab-org/gitlab-pages/-/merge_requests/274

I'm going to assume the aws library can make use of these connections.

Additionally, I did not add these params to the `config` module because I don't think it's necessary to bring it out that far.

### DIFF
--- a/core/storageproviders/s3Storage.go
+++ b/core/storageproviders/s3Storage.go
@@ -3,9 +3,11 @@ package storageproviders
 import (
 	"bufio"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/owncast/owncast/core/data"
 	"github.com/owncast/owncast/core/playlist"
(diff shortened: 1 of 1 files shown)


gabek:
Awesome!

Would you mind running the linter on your changes? Seems like maybe it's unhappy about the ordering of imports.

128f:
@gabek looks like it was a spacing issue
`golangci-lint run --fix` fixed it :+1:

gabek:
Sweeeet. Maybe this can squeeze just a little bit of additional performance out of those S3 transfers. I appreciate you looking at it!

Issue #1953, opened 2022-06-15
Implement multiplatform Docker image


hufman:
This project is incredible, thank you so much for building it!
I'd like to add support for building an arm64 Docker image. The Go toolchain makes [cross-compilation](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/) very easy, and your Dockerfile is already set up with multi-stage builds which should keep the build fast.

I'd like your review on my approach before I start:

- I'd like to move the Tailwind CSS compilation out of [docker-nightly.sh](https://github.com/owncast/owncast/blob/34629d0f5da8fd6e82750a49cd7b608a644c320d/build/release/docker-nightly.sh#L8-L15) and [release.sh](https://github.com/owncast/owncast/blob/34629d0f5da8fd6e82750a49cd7b608a644c320d/build/release/build.sh#L43-L50) into another stage within the Dockerfile.
- I'd like to switch the [Docker Nightly Action](https://github.com/owncast/owncast/blob/develop/.github/workflows/docker-nightly.yaml) (and maybe the release action, but I was unable find it) to use the [Docker Buildx build and push actions](https://github.com/docker/build-push-action/blob/master/docs/advanced/push-multi-registries.md).

Admittedly I only started using this project yesterday, so please let me know if there are any other things I should consider or any other thoughts you have!

gabek:
Thanks for the kind words! And thanks for checking in, A multiplatform image would be great.

While the Go toolchain is good with cross-compilation, unfortunately we're using cgo, and that adds a bunch of complexity in building. That's why in `release.sh` you can see it's building with `xgo` instead of `go` so we can get multi-platform binaries.

Because our Dockerfile is assuming only 64bit Linux we don't need to use xgo. But with a multiplatform build we would.

[I've been trying to migrate away from xgo](https://github.com/owncast/owncast/issues/1756) for some time, with hopes of moving to Earthly. I _think_ I got pretty far with that, but it hasn't been a priority: https://github.com/owncast/owncast/blob/gek/earthly-builds/Earthfile

As for Tailwind, this next release (0.0.12) will be the last release that requires the Tailwind CSS compilation, so maybe it's not worth making changes to that.

Sorry for adding complexity to what would generally be a pretty straightforward thing, but I'm open to any ideas you have!

[...]

PR #1962, opened 2022-06-19
Hufman/earthly builds

### LINKED ISSUE #1756: Move away from xgo in building releases
It's been a nice solution for cross-compiling, but it's extra moving parts and the original project has been archived, requiring a fork.

Finding a new cross-compiling solution that supports cgo is needed.

hufman:
Finalizes the initial Earthfile to produce artifacts for all supported platforms.
`earthly +package` should compile and product a distribution artifact for the current platform.
`earthly +package-all` should produce distribution artifacts for all platforms.
`earthly +docker` should produce a docker image for the current platform.
`earthly +docker-all` should produce docker images for all platforms, but requires emulation.
`earthly --push +docker-all` should publish docker images for all platforms, you might consider using a specific --version tag. It doesn't currently change the :latest tag.

Fixes #1756 - Removes xgo from the build tooling
Works towards #1953 but needs CI integration next.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,0 +1,79 @@
+VERSION 0.6
+
+FROM bdwyertech/go-crosscompile
+
+ARG version=dev
+WORKDIR /build
+
+code:
+  # COPY . /build
+  GIT CLONE --branch=develop git@github.com:owncast/owncast.git /build
+
+build:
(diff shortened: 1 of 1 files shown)


gabek:
Sweeeeet!

PR #2026, opened 2022-08-02
Messages table fixes to improve query performance

### LINKED ISSUE #1890: Improve chat history query performance
An issue that has lingered, improved slightly, but still occurs as our queries have become more complex.

When the tables have many rows, the chat history query can get very slow at scale. It doesn't impact most streamers, but one person reported a problem with it.

Query: https://github.com/owncast/owncast/blob/develop/core/chat/persistence.go#L292

I'd love the help of anybody who can shed some light on optimizing SQL queries.

gabek:
- Fixes messages table schema user_id type from `INTEGER` to `TEXT`.
- Add index `idx_messages_hidden_at_timestamp`.
- Fix index name.
- Fix how indexes were being created.
- Add migration that simply drops the `messages` table and recreates it.

Ultimately not a huge improvement, but things are way more correct with these changes.

---

Query plan before changes:
image

Query duration before changes:
Limit 50: 0.2s

Query plan after changes:
image

Query duration after changes:
Limit 50: 0.14s

Fixes #1890

### DIFF
--- a/core/data/data.go
+++ b/core/data/data.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	schemaVersion = 5
+	schemaVersion = 6
 )
 
 var (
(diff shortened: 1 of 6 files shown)


jprjr:
Looks good to me!

PR #2044, opened 2022-08-17
Add Fediverse, Matrix and XMPP social links

### LINKED ISSUE #1981: Add "Fediverse" social link
Right now we offer "Mastodon", but Mastodon is not the Fediverse. While a part of me wants to remove the "Mastodon" option completely, because Mastodon is a piece of software, not a network, many people mistakenly call the Fediverse "Mastodon", so I'll leave it and encourage people to use a new "Fediverse" option instead.

rotmoded:
Fixes #1981 

Removed roundedness from icons in frontend too, they only cut of parts of perfectly square icons.


gabek:
I didn't think of Matrix and XMPP, those are great additions I'm sure people will appreciate. Thank you so much!

PR #2047, opened 2022-08-18
Change validation in social link modal to URI over URL

rotmoded:
This goes along with my last commit (#2044), in order to accept adding of XMPP (and in the future other) URIs in social links instead of just http/https URLs.

### DIFF
--- a/web/utils/uris.ts
+++ b/web/utils/uris.ts
@@ -1,0 +1,3 @@
+export default function isValidURI(uri: string): boolean {
+  return uri.includes(':');
+}
--- a/web/utils/uris.ts
+++ b/web/utils/uris.ts
@@ -1,3 +1,28 @@
 export default function isValidURI(uri: string): boolean {
-  return uri.includes(':');
+  if (uri.includes(':')) {
+    const splittedURI = uri.split(':', 2);
+    if (
+      splittedURI[0] === ''
+    ) {
+      return false;
+    } else if (
+      splittedURI[1] === ''
+    ) {
+      return false;
+    } else if (
+      splittedURI[0] === 'http' ||
+      splittedURI[0] === 'https'
+    ) {
+      if (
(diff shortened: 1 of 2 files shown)


web/utils/uris.ts
```diff
@@ -0,0 +1,3 @@
+export default function isValidURI(uri: string): boolean {
+  return uri.includes(':');
```
gabek: Maybe we can also check for a top level domain. Right now if somebody types `https:` it'll be seen as a completely valid input.

rotmoded:
Ok, I've added a couple more checks in the validator :) hopefully this is good. I don't actually know typescript/javascript so apologies if it's a little bit of a spaghetti, I would usually use pattern matching for something like this.... I've tested the function thoroughly enough I hope.

rotmoded:
Please let me know if you would like me to squash everything to cf56edb

gabek:
Just to let you know I haven't forgotten about this PR, I was just going to look around to see if there's some more concise way we can perform this validation without the explicit string parsing. Seems like this might be in the standard library somewhere but I haven't yet looked for it.

rotmoded:
I looked in the standard library already and I could not find a URI type. I found some libraries but none of them implement validation at all or like this, as there are so many types of URI and possible validation rules for each type. The closest I have been able to find are parsers to make URI types, in which the validation would need to be implemented in a similar way to it has already been done with the string type. This is probably the most efficient way to do it in this case and it seems to fit the requirements. If you want I could make the control flow make it go through a URL specific validator (probably just use the existing isvalidurl lmao) if http(s) is detected instead of the manual validation I've done. Anyways sorry this should be a trivial issue lol

gabek:
Sorry about that. I went ahead and split up xmpp, matrix and http validation just to keep things cleaner. I'm going to go ahead and close this PR.

PR #2082, opened 2022-09-04
reafctor: normalize component formatting

### LINKED ISSUE #2081: Mixed formatting in components

Components have a mix of stylings, import/export patterns, and FunctionalComponent usage.

I propose:

- all components are named exports (as the non-default export)
- all components' props are written as types, not interfaces
- all components' props are named exports
- all components are written with the modern `const Foo = () => ...` syntax
- components without logic or hooks should return with `() => (<>...</>)`, if possible
- components with props should define their props type by typing the `FC` type, not using the JSX PropTypes lib
  - this will result in this syntax: `const Foo:FC<FooProps> = ({ a, b }) => ...`
  - n.b.: the `FC` type wasn't always recommended due to the implicit `{children: React.reactNode}` field, but since React 18, this has been removed and `FC` is now recommended
  - JSX's PropTypes is unnecessary in TSX
- class components should be marked as `needs refactoring` and an issue should be opened to track rewriting them as functional components
- event handlers should be prefixed with `on`

jamescallumyoung:
# Description

Normalize the formatting of components, as described in #2081 . The rationale is layed out in that issue. This PR also updates some eslint rules, to help enforce the new formatting.

Some `export default` statements were not removed, as my I was unable to verify that those exports are unused using my IDE. A full component dependency graph should be able to identify whether they're safe to remove, bit this is out-of-scope of this PR's effort.

Fixes # (issue)
#2081 

---

See: #2081 

~3. Explain how it works and decisions you made.
5. If you're fixing something, what was wrong? How should we stop from having this issue happen again?
6. If this is a new feature or addition to functionality, why should it be added? What are the use cases? Who was asking for this functionality?~

N/A

~If this is an unsolicited change or have no issue associated please do your best to detail the motivations behind this PR.~

N/A


jamescallumyoung:
NOT READY TO REVIEW

gabek:
Saw that the CI stuff isn't running successfully on remote branches so don't worry about those failures. I pushed a change to those jobs to hopefully resolve it.

jamescallumyoung:
@gabek I pushed a change to the eslint config for the `/components` and `/pages` dir so (some of) the style changes suggested in #2081 are enforced. That fixed the linting issue in CI/CD too.

jamescallumyoung:
@gabek This PR is _substantial_. It mostly does the same thing to every file:
- update each components signature to `const Foo: FC<FooProps> = ({ bar }) => ...`.
- update the exports to favor named exports over default exports.
- removes the redundant `index.ts` files in some component's dirs.

Hopefully, the changes are clear. Let me know if not 👍.

gabek:
Wow, even went and cleaned up the old Admin files. Substantial indeed, thank you!

gabek:
Looks like NextJS doesn't like the `web/pages/.eslintrc.js` that was added.

> found page without a React Component as default export in pages/.eslintrc

gabek:
Maybe nextjs has a way to ignore certain file patterns when building. I'll check that out.

gabek:
Just out of personal curiosity, given that the linter default is "prefer default exports", what's the benefit of requiring non-default exports?

I have no preference myself, since I have never thought anything of it either way.

gabek:
FYI I keep doing work in `webv2` where the linter requires default exports and it keeps creating conflicts with your branch, so I've been fixing your branch since I'm the one creating the problems.

gabek:
I went ahead and edited the next config file to only look at `.tsx` files when doing builds in order to ignore the new `.eslint.js` that was added.

gabek:
I'm going to merge this in. The longer it's unmerged the more of a chance I forget what standards you set and I don't use them :)

Thanks again for going through and making all those changes.

jamescallumyoung:
Hey @gabek thanks for taking this over and merging it 👍

> Just out of personal curiosity, given that the linter default is "prefer default exports", what's the benefit of requiring non-default exports?

There's a few reasons:

- Exporting a component as a default export in a single line (`export default function ...`) cannot be auto-fixed by eslint, as it mixes export and const declarations into a single statement.

- We want to export every component's Props along with the component. Using a named export, that becomes `import { Foo, FooProps } from '../foo`, which I think is nicer than mixing named and default.

- When we export a component with a, default export, there's a endency to import the component with a different name. I noticed this a lot in the codebase (e.g. `import Bar from '../foo.tsx`). This can make it difficult to know which component is actually being used. (We can still rename an import using `import { Foo as Bar }` if needed.)

gabek:
Those are great explanations and are all real world benefits for sure, thank you for clarifying!

Issue #2119, opened 2022-09-15
Frontend architecture: Atomic Design


gabek:
### Share your bug report, feature request, or comment.

We're looking to adhere to the [Atomic Design](https://bradfrost.com/blog/post/atomic-web-design/) methodology to build up a library of smaller components that can be used to build current and future features and interfaces for Owncast. It's a little bit of work now to save additional work later.

https://collab.owncast.tv/kanban/#/2/kanban/edit/omLI2N+LcnP+elmdT7qW9GHD/

> Atomic design gives us the ability to traverse from abstract to concrete. Because of this, we can create systems that promote consistency and scalability while simultaneously showing things in their final context. And by assembling rather than deconstructing, we’re crafting a system right out of the gate instead of cherry picking patterns after the fact.

This approach is new to the project, so please bear with us, especially as we begin the process.

- [Atoms](https://bradfrost.com/blog/post/atomic-web-design/#atoms) are our HTML tags, such as a form label, an input or a button.
- [Molecules](https://bradfrost.com/blog/post/atomic-web-design/#molecules) are groups of atoms bonded together and are the smallest fundamental units of a compound. These molecules take on their own properties and serve as the backbone of our design systems.
- [Organisms](https://bradfrost.com/blog/post/atomic-web-design/#organisms) are groups of molecules joined together to form a relatively complex, distinct section of an interface.

Templates and pages fall under larger layout-related things that will be handled separately. 

Some of our components barely need any work at all to convert the component into its atomic structure, and some need more work.

## How to help

[We have a board with a list of components](https://collab.owncast.tv/kanban/#/2/kanban/edit/omLI2N+LcnP+elmdT7qW9GHD/
). If you'd like to help:

Items are tagged as either an atom, molecule or organism based on what we think they should end up being, based on role, size and complexity.

1. Pick a component from the list.
2. Move it to "In progress".
3. If this component has smaller components within it and are listed in the board, move those to "In progress as well".
4. Split the component into small, standalone components such as the text input, button, etc.
5. If existing small components already exist then use those.
6. Create a PR with your changes.

## How to work on the Owncast frontend

Visit the [How to get started with the Owncast frontend](https://owncast.online/development/#how-to-start-with-frontend-development) document for details.

![image](https://user-images.githubusercontent.com/414923/190498322-ac2367ad-bd5f-4e63-9296-8d423d1660d7.png)

jamescallumyoung:
[...]

PR #2125, opened 2022-09-16
refactor(#2119): update button and button row to follow atomic design

jamescallumyoung:
# Description

Refactor the `ActionButton` and `ActionButtonRow` components to be atomic, replacing them with cleaner `Button` and `ButtonRow` components.

NOTE: WIP; This PR does not yet handle the integration of these components.

Fixes # (issue)
Part of the ongoing work towards #2119 

---

The `Button` component no-longer includes a modal. Instead, the modal and button are seperate components. This preserves the actual DOM structure in the components; the modal isn't _inside_ the button in the DOM, so its not in the components.

Also, the button's props have been updated to expose _most_ of the underlying Ant Design button's props. This may not be desired in most components but I though it may be useful for these very low-level "Ant Design wrapper" components. This behaviour can be updated as we continue the Atomic Design updates.

### DIFF
--- a/web/components/atomic/atoms/Button/Button.tsx
+++ b/web/components/atomic/atoms/Button/Button.tsx
@@ -1,0 +1,28 @@
+import { Button as AntButton, ButtonProps as AntButtonProps } from 'antd';
+import { FC } from 'react';
+
+import styles from './Button.module.scss';
+
+export type ButtonProps = Omit<AntButtonProps, 'children' | 'icon' | 'title' | 'target'> & {
+	text: string;
+	iconSrc?: string;
+	iconAltText?: string;
+};
+
+/**
+ * A customised and styled Ant Design button.
+ *
+ * An icon can be set. Both `iconSrc` and `iconAltText` must be set for the icon to be shown.
+ */
+export const Button: FC<ButtonProps> = ({ text, iconSrc, iconAltText, type = "primary", ...props }) => (
+	<AntButton
+		type={type}
+		className={styles.button}
+		{...props}
+	>
+		{iconSrc && iconAltText !== undefined && (
+			<img src={iconSrc} className={`${styles.icon}`} alt={iconAltText}/>
+		)}
+		{text}
+	</AntButton>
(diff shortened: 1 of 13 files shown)


web/utils/storybook-utils.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
jamescallumyoung, line 1: This fn is useful when writing stories for components, as it allows fields to be easily ommitted from the controls panel in Storybook.

jamescallumyoung:
ActionButton is deprecated but ActionButtonRow is not as it is still used by FollowButton et al.

jamescallumyoung:
@gabek Could you test this out locally too? I'm still not 100% on running locally.

gabek:
Looks like the linting and formatting isn't happy even after it changed some stuff via Prettier.

jamescallumyoung:
@gabek fixed the linting issue. The linter failed to fix some issues automatically - sometimes eslint just can't 🤷 

I updated the component .eslint file to change a rule: dev dependencies can be imported in the co-located `.stories.tsx` files. This is done because `@storybook` modules are dev dependencies and are required in stories.

I also moved the `ButtonRow` description to a JSDoc comment, as I suggested you do in #2129 .

gabek:
When testing It looks like the optional color is getting lost.

Before:
image

After:
image

web/components/atomic/atoms/Button/Button.tsx
```diff
@@ -0,0 +1,30 @@
+import { Button as AntButton, ButtonProps as AntButtonProps } from 'antd';
+import { FC } from 'react';
+
+import styles from './Button.module.scss';
+
+export type ButtonProps = Omit<AntButtonProps, 'children' | 'icon' | 'title'> & {
+  text: string;
+  iconSrc?: string;
+  iconAltText?: string;
```
gabek, line 9: I think color is missed here, and then not used, so that's why the background color is getting lost.

PR #2129, opened 2022-09-17
Recategorize Modal as a molecule. For #2119

gabek:
@jamescallumyoung This is a pretty simple one, since it was already standalone, but I figured I'd start somewhere! Let me know if there's something obvious I'm missing.


jamescallumyoung:
Looks good!

If I was writing this new, I might create a basic `Modal` atom, and an `IFrameModal`, but it's not really _needed_ in this case. Just something to consider for the next component. We can always break down components into smaller pieces as they're touched, too.

I'd suggest we also include some adjacent changes when touching components:
1) Move storybook docs to JSDoc comments, so they can be read by IDEs and other tools.
2) Remove defaultProps. It's a _JS_ library that doesn't really make sense when using TS. We can set the default values in the function args.

~I'll add a suggestion for both of the above.~

web/components/atomic/molecules/Modal/Modal.stories.tsx
```diff
@@ -3,12 +3,12 @@ import { ComponentStory, ComponentMeta } from '@storybook/react';
...
+  title: 'molecules/Modal',
   component: Modal,
   parameters: {
     docs: {
       description: {
-        component: `This is the popup modal container that all modal content is rendered inside. It can be passed content nodes to render, or a URL to show an iframe.`,
+        component: `This is the container that all modal content is rendered inside. It can be passed content nodes to render, or a URL to show an iframe.`,
       },
     },
   },
```
jamescallumyoung: (GH won't let me add a suggestion 😐)

Suggestion: Remove these lines and copy the description text as a comment to the component in JSDoc syntax. This will allow IDEs to find the component's description, and enables further tooling in the future.

jamescallumyoung (changes requested):
Can't add suggestions in GH as the code is not line-adjacent to the lines you touched. See earlier comment.

t1enne:
I'm definitely for splitting stuff in Modal and Iframe, and also, since we're at it, I noticed support for the native <dialog> html and it's showModal() method is great (only IE11 and Opera Mini from 2015 have bad support, see [here](https://caniuse.com/mdn-api_htmldialogelement_showmodal)). In 99% of the cases, using native html is a better choice. We could eventually migrate our modal to native html. What do you think?

gabek:
We can file something in the backlog to at some point in the future rewrite it without using the React component, but the Modal component is used a handful of different places and I don't want to start over from scratch. We're already doing a lot of "take one step forwards, take two steps back", so I fear if we keep looking for things to rewrite we'll never get anything completed.

gabek:
I went ahead and filed #2136 to track rebuilding the modals.

jamescallumyoung:
I agree. The modal re-work can wait. The component works fine as-is, so let's get it merged! 😄

There's still a change request: replace default props with default args, and replace Storybook docs string with JSDoc.

gabek:
Sorry for my slacking! I'm out of town for the next couple days and then I'll jump on fixing up this PR with the suggested changes, and then it can go in.

gabek:
@jamescallumyoung Done! Let me know if I'm missing anything.

PR #2155, opened 2022-10-02
Treat fediverse usernames as case-insensitive

### LINKED ISSUE #2048: Fediverse auth treats usernames as case-sensitive
Hi there - I noticed that when performing fediverse-based authentication, the `token` value stored in the database is used in a case-sensitive manner. So if I type `Username@example.com` - that stores a different entry in the `auth` table than `username@example.com`. Meaning I can wind up restoring a different user. It's created some confusion with users - they go to auth, type in their fediverse username in a different case than before, and wind up not restoring the user they intended to.

I can't find a spec that states usernames are considered case-insensitive, but that seems to be the case for most fediverse-based apps. I think converting the token to lowercase could prevent the issue.

brokenintuition:
Fixes https://github.com/owncast/owncast/issues/2048

I believe this should fix the issue, since it's the account from `OTPRegistration` that gets stored. I didn't attempt to handle usernames that were already stored with capital letters, so those users would restore a new user when authenticating. I can take a look at trying to handle existing users if you think that's necessary.

### DIFF
--- a/auth/fediverse/fediverse.go
+++ b/auth/fediverse/fediverse.go
@@ -3,6 +3,7 @@ package fediverse
 import (
 	"crypto/rand"
 	"io"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


gabek:
Thanks so much for looking at this! I just tested it and it works great.

While a migration could be created to normalize the case of all the existing users, I think it's probably ok just to leave them. If I hear about a ton of people running into this issue then maybe it can be revisited in the future.

PR #2168, opened 2022-10-03
Add ability to move social handles up/down

### LINKED ISSUE #1818: Feature Request: Sortable social media links
Hello all,
I think it could be a useful feature to make the social media links sortable with a little up and down function to move them back and forth. I noticed the problem when my friend and I added all our links randomly and found that they were displayed on the front page in the order they were created. So we had to remove all the links again and put them in the order we wanted them. I realize that this would be a pure usability feature but, if by chance there is time, maybe we could implement something like this.

pippo:
This is an attempt to make social handles on the "Configuration > General" page sortable. The 2 buttons added next to each of the social handles - to move the item up and down respectively. The move is followed by a request to API to persist the change on the backend. UI might need some additional polishing, but from the functionality perspective should address the feature request.

Ref: https://github.com/owncast/owncast/issues/1818

### DIFF
--- a/web/components/config/EditSocialLinks.tsx
+++ b/web/components/config/EditSocialLinks.tsx
@@ -1,7 +1,7 @@
 import React, { useState, useContext, useEffect, FC } from 'react';
 import { Typography, Table, Button, Modal, Input } from 'antd';
 import { ColumnsType } from 'antd/lib/table';
-import { DeleteOutlined } from '@ant-design/icons';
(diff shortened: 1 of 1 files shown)


gabek:
Works great! Could you take a look at the build errors that were reported?

pippo:
@gabek Looks better now. PTAL 🙏

gabek:
Awesome, thanks so much! I'm sure users of Owncast will appreciate this.

PR #2171, opened 2022-10-04
Fix slider marks by changing the corresponding labels to non-empty string

### LINKED ISSUE #2159: Admin only allows selecting 24fps or 120fps, nothing inbetween

Something must have changed with the slider component, as it no longer is allowing selection of any framerates aside from the first option and the last option.

image

pippo:
Slider marks seem to be ignored when the corresponding label is an empty string. The workaround of changing it to a single space seems to be doing the trick, while not introducing any visual changes.

Fixes https://github.com/owncast/owncast/issues/2159


gabek:
Awesome! Would you mind commenting on the parent issue so it can be assigned to you so you get the credit? Github won't let me assign issues to somebody unless they comment on it first.

gabek:
Thank you for finding that fix, it works great!

PR #2173, opened 2022-10-05
Add Gitpod config

aprilthepink:
Gitpod is a free only ide which allows you to edit and run code in the browser. This script eases the gitpod setup.
We can discuss here more about it but gitpod [has much already explained themself](https://www.gitpod.io/)
Happy coding :)


gabek:
Sure, why not! I've never used Gitpod myself but if this single file can help people get up and running I don't see a downside.

PR #2175, opened 2022-10-06
clarified in README.md about musl compatible compiler

### LINKED ISSUE #2009: Update source build instructions to not be specific about GCC
Hello,

Request: Update [README.md](https://github.com/owncast/owncast/blob/develop/README.md) so that "Building from Source" section does not imply that Owncast runs only on glibc systems. I thought it was glibc exclusive, but I tried anyway and it works on Musl as well.

**Long story short**: Owncast can be compiled on Void Linux (Musl)

**Long story long**:
I have Void Linux (Musl) on my server. 

in "Building from Source" section in [README.md](https://github.com/owncast/owncast/blob/develop/README.md) it is stated that:
> 1. Ensure you have the gcc compiler installed.

I understand, that You meant that it is required to have A COMPILER installed, but it can just as well be Musl.
[Go](https://voidlinux.org/packages/?arch=x86_64-musl&q=go) is the important piece of software here and it is  available for Musl version of Void Linux.

After installing go with
> xbps-install go

I did

>git clone https://github.com/owncast/owncast
>cd owncast
>go run main.go

After downloading dependencies, Owncast started without a problem.

I then created a binary file with 
>go build

and "owncast" binary appeared, that could be started on a Musl system.

I totally understand, that going with just GCC, saves Your time in terms of support, but it would be nice if you (at least) gave a hint that Musl systems are also capable of running Owncast.
>1. Ensure you have a compiler installed (GCC or Musl)

jahway603:
README.md now includes that a musl compatible compiler can be used to build owncast. This is to resolve issue https://github.com/owncast/owncast/issues/2009

happy #hacktoberfest

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,11 @@ OBS, Streamlabs, Restream and many others have been used with Owncast. [Read mor
 ## Building from Source
 
 1. Ensure you have pre-requisites installed.
-    - [gcc compiler](https://gcc.gnu.org/install/download.html)
+    - C compiler, such as [GCC compiler](https://gcc.gnu.org/install/download.html) or a [Musl-compatible compiler](https://musl.libc.org/)
     - [ffmpeg](https://ffmpeg.org/download.html)
-1. Install the [Go toolchain](https://golang.org/dl/) (1.16 or above).
+1. Install the [Go toolchain](https://golang.org/dl/) (1.16 or above). If your package manager has an older version of go, then you can install and use a newer version with the [Go Version Manager](https://github.com/moovweb/gvm).
 1. Clone the repo. `git clone https://github.com/owncast/owncast`
-1. `go run main.go` will run from source.
+1. `go run main.go` will run from source (or alternatively run `go build main.go` to build a binary from source and then run it)
 1. Visit `http://yourserver:8080` to access the web interface or `http://yourserver:8080/admin` to access the admin.
 1. Point your [broadcasting software](https://owncast.online/docs/broadcasting/) at your new server and start streaming.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,11 +77,11 @@ OBS, Streamlabs, Restream and many others have been used with Owncast. [Read mor
 ## Building from Source
 
 1. Ensure you have pre-requisites installed.
-    - [gcc compiler](https://gcc.gnu.org/install/download.html)
+    - C compiler, such as [GCC compiler](https://gcc.gnu.org/install/download.html) or a [Musl-compatible compiler](https://musl.libc.org/)
     - [ffmpeg](https://ffmpeg.org/download.html)
-1. Install the [Go toolchain](https://golang.org/dl/) (1.16 or above).
+1. Install the [Go toolchain](https://golang.org/dl/) (1.16 or above). If your package manager has an older version of go, then you can install and use a newer version with the [Go Version Manager](https://github.com/moovweb/gvm).
```
gabek: I fear this makes Go Version Manager look like an additional tool that needs to be installed to build Owncast, when in reality you can just download an updated version from the provided link.

README.md
```diff
@@ -77,11 +77,11 @@ OBS, Streamlabs, Restream and many others have been used with Owncast. [Read mor
...
 1. Clone the repo. `git clone https://github.com/owncast/owncast`
-1. `go run main.go` will run from source.
+1. `go run main.go` will run from source (or alternatively run `go build main.go` to build a binary from source and then run it)
```
gabek: Just to keep things simple I think it's best not to have any additional information about creating builds here, just to minimize the choices people have to make to run it from source.

gabek:
Thank you!

PR #2180, opened 2022-10-09
add tests for webhook events

### LINKED ISSUE #1528: Add automated tests for Webhooks
Currently, Webhooks are only tested manually (see e.g. [here](https://github.com/owncast/owncast/pull/1525#issuecomment-967743810)).

To check that they don't break, we should add some manual tests related to this topic. For this, I'd start a separate http server and check if the webhooks are coming through.

mdonoughe:
This PR adds automated tests for both the sending of webhooks and for the format of their payloads.

Is there a way in Go to specify that the `json-patch` dependency is only required for tests?

Is `SetMessageVisibilityEvent` right? It looks like there's another event nested inside that is never populated?

Closes #1528

### DIFF
--- a/core/chat/messages.go
+++ b/core/chat/messages.go
@@ -29,13 +29,7 @@ func SetMessagesVisibility(messageIDs []string, visibility bool) error {
 		return errors.New("error broadcasting message visibility payload " + err.Error())
 	}
 
-	// Send webhook
-	wh := webhooks.WebhookEvent{
-		EventData: payload,
-		Type:      event.GetMessageType(),
-	}
-
-	webhooks.SendEventToWebhooks(wh)
+	webhooks.SendChatEventSetMessageVisibility(event)
 
 	return nil
 }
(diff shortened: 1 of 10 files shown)


gabek:
Awesome, thank you!

I just learned from your PR that the tests running under "1.18.x" aren't actually running under 1.18, because if they were your new tests would have failed because you're using a type that only exists in 1.19. Whoops!

We haven't had any 1.19 requirements until now, so maybe it's time to force that version update for the project. But I feel like I can't just flip this requirement without warning, especially during Hacktoberfest where more people are trying to run the project for the first time.

So I hope if you don't mind that I hold off merging this in until the end of the month to allow a wider variety of Go versions. And I'll prepare people before the change so people aren't surprised.

gabek:
That's very cool of you to change that type, you didn't have to do that, but it's super appreciative!

> Is there a way in Go to specify that the json-patch dependency is only required for tests?

It's fine listing it as a dependency as it won't actually get built unless it's used. I'll double check that the tests aren't getting pulled into the production build. They shouldn't be, but I hadn't actually thought about it until now.

Thanks again so much for this set of tests. When I get around to it in the future I'd like to use this same approach with Fediverse tests. Your contribution will pay dividends down the line for sure :)

Let me know if you'd be interested in contributing to the project more down the road.

Issue #2076, opened 2022-09-02
Outdated frontend dependencies


jamescallumyoung:
### Share your bug report, feature request, or comment.

The dependency graph for the project is quite the mess right now. There's a lot of conflicting dependencies (mostly on react) and a few dependencies that are either unmaintained or deprecated.

I'll add install and audit logs in commens below.

I suggest removing one-off and unnecessary dependencies first, migrating to the latest React + NextJS versions, and then re-reviewing the audit results.

jamescallumyoung:
`npm i` gives:

```
ERESOLVE overriding peer dependency
npm WARN While resolving: react-toggle-component@3.0.8
npm WARN Found: react@17.0.2
npm WARN node_modules/react
npm WARN   react@"17.0.2" from the root project
npm WARN   98 more (@ant-design/icons, @ant-design/react-slick, ...)
npm WARN
npm WARN Could not resolve dependency:
npm WARN peer react@"^16.8.6" from react-toggle-component@3.0.8
npm WARN node_modules/react-toggle-component
npm WARN   react-toggle-component@"^3.0.8" from addon-screen-reader@1.8.6
npm WARN   node_modules/addon-screen-reader
npm WARN
npm WARN Conflicting peer dependency: react@16.14.0
npm WARN node_modules/react
npm WARN   peer react@"^16.8.6" from react-toggle-component@3.0.8
npm WARN   node_modules/react-toggle-component
npm WARN     react-toggle-component@"^3.0.8" from addon-screen-reader@1.8.6
npm WARN     node_modules/addon-screen-reader
npm WARN ERESOLVE overriding peer dependency
npm WARN While resolving: react-toggle-component@3.0.8
npm WARN Found: react-dom@17.0.2
npm WARN node_modules/react-dom
npm WARN   react-dom@"17.0.2" from the root project
npm WARN   80 more (@ant-design/icons, @design-systems/utils, ...)
npm WARN
npm WARN Could not resolve dependency:
npm WARN peer react-dom@"^16.8.6" from react-toggle-component@3.0.8
npm WARN node_modules/react-toggle-component
npm WARN   react-toggle-component@"^3.0.8" from addon-screen-reader@1.8.6
npm WARN   node_modules/addon-screen-reader
npm WARN
npm WARN Conflicting peer dependency: react-dom@16.14.0
npm WARN node_modules/react-dom
npm WARN   peer react-dom@"^16.8.6" from react-toggle-component@3.0.8
[...]

PR #2198, opened 2022-10-12
React 18

heller:
I may have gone overboard and updated everything rather than just React 18, but the stream and admin UIs seem to work at a glance.

Also fixed most issues raised by `npm audit --omit=dev`.


heller:
Relates to #2076, #1761.

gabek:
Let's do it!  Thank you, Matt! :everyemoji:

PR #2211, opened 2022-10-15
fix followers: align the followers to the left

### LINKED ISSUE #2207: Followers should be left aligned, not centered.

image

Pranav2612000:
This PR modifies the logic for placing followers, aligning them to the left.

@gabek 
I've adding a 5px padding to the followers container and removed the justify:space-around attribute from the row. This is how it looks now - 
image

Some more cases - 
More than 1 follower in the same row
image

More than 1 row
image

The more than 1 row looks a bit weired due to the spacing at the left. Fixing it is not straightforward as it seems.  Is that fine for now? If not, what's the expected behaviour? Also, I think we'll need to sprinkle in some JS to make it work, which can be complex.

Fixes #2207


gabek:
On the old version of the web interface (https://nightly.owncast.tv/) we'd use all the available space. Do you think we could do the same thing here?

image
 

<blockquote><div> Owncast Nightly</div><div><strong><a href="https://nightly.owncast.tv/">Owncast Nightly</a></strong></div><div>This is a server running the nightly build of Owncast.</div></blockquote>

Pranav2612000:
Yeah. Based on my explorations on the nightly site here's what I observed:
The width of the cards increases and decreases so that the cards fill up the whole space. And the number of cards in a row is 1 for mobile screens, 2 for ipad screens, 3 for lg screens ( till 1440 ) and 4 thereafter?

Is it fine if I use this logic to build the component?

gabek:
Yeah, that sounds great, thank you for looking at it!

PR #2212, opened 2022-10-15
improv ui: make footer sit at bottom

### LINKED ISSUE #2188: Footer isn't on the bottom of the page

There is a gap between the bottom of the screen and the footer.

image

Pranav2612000:
This commit makes the footer always attach to the bottom of the screen. This is done by setting the minWidth of the main antd layout component to 100vh ( https://stackoverflow.com/questions/52282191/reactjs-ant-design-fix-footer-with-layout ) and moving the footer component outside the spinner, since antd does not allow granular component inside the spin component.

This commit also removes the previous spinner styles which were not being used as antd did not support `className` ( https://ant.design/components/spin/ )

Screenshots:
image

image

image


Pranav2612000:
Fixes #2188

gabek:
Thank you so much!

Issue #2189, opened 2022-10-10
Header does not fill full width on narrow windows/mobile layouts


gabek:
### Share your bug report, feature request, or comment.

![image](https://user-images.githubusercontent.com/414923/194807488-c1b30f1b-c95a-469d-859b-19fc37992888.png)

Pranav2612000:
Oh. Just found out the reason for this. It's an interesting issue. Creating a  PR

gabek:
So unfortunately this issue is still occurring. It certainly seems to depend on what the title of the instance is. The default "Owncast" doesn't cause the issue, but the "Owncast web v2" does.

![image](https://user-images.githubusercontent.com/414923/196607607-ee9505e9-3660-4c70-86e2-ae6ba4df418d.png)

Pranav2612000:
Must be something similar. I'll take a look.

t1enne:
This is fixed with my last commit! Sorry, didn't notice we had an open issue.

gabek:
Yes, fixed! Thanks for the update @t1enne. Closing.

PR #2213, opened 2022-10-15
fix ui: make navbar cover whole width of page

Pranav2612000:
Previously, there was a large url on the page, and the a tag did not have word break property set - which resulted in the browser trying to display the whole page with increased width. Allowing the a tag to have a word break fixes this issue.

Screenshots:
Before:
image

After:
image

### DIFF
--- a/web/styles/globals.scss
+++ b/web/styles/globals.scss
@@ -86,6 +86,7 @@ body {
 
   a {
     color: var(--theme-color-action);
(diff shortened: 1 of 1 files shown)


Pranav2612000:
Fixes #2189

gabek:
Such a good find, thank you for fixing that! I never would have guessed it had to do with a line of text not breaking :)

PR #2223, opened 2022-10-18
Web UI frontend automated browser tests

### LINKED ISSUE #1926: UI Testing
With Web UI v1 we used puppeteer to perform some automated testing of the web page as a whole. Since we're starting fresh we can either use puppeteer again or look at other tools that fit our needs.

Some testing options:

- Storbyook-based testing https://storybook.js.org/blog/storybook-6-5/
- Cypress

### LINKED ISSUE #1929: Mobile testing
Since I'm not in the habit of testing our new layout and components against mobile/narrow windows, we should have a way to automate this in some fashion.  Previously, for web ui v1, I had Puppeteer take automated screenshots of different sizes, but unless I went out of my way to go look at them, they were silently ignored.

Chromatic, the company that makes Storybook, suggests we use them for automated visual testing.  https://storybook.js.org/tutorials/visual-testing-handbook/react/en/automate/

I don't know, maybe we should use it and not think too much of it? We're already using Storybook so maybe it's worth just using the entire ecosystem to make testing easier.

gabek:
WIP while we wait for the frontend to be completed.


gabek:
Still plenty of UI tests to write, especially the desktop vs. mobile, but this is a start.

PR #2224, opened 2022-10-18
fix purple background color for notification reminder pop up & fix failing build

### LINKED ISSUE #2114: Notification reminder popup is incorrectly colored

Previously it [was purple](https://owncast.online/components/static/media/notify-popup.7df1d4bb.png) and looked better, but now is just using the default background color of the page. 

Component: https://owncast.online/components/?path=/story/owncast-components-notify-reminder--active
Colors: https://owncast.online/components/?path=/story/owncast-style-guide-default-theme--page

Currently `theme-color-components-menu-background` is just the background of the page so we could either change that variable, or style this one particular popup to use a different color variable.

unclebinary1001:


gabek:
Awesome, thanks so much for fixing that. I didn't know about about `export` from CSS modules, that's really cool. Thanks for sharing something new!

PR #2256, opened 2022-10-24
Remove follower on delete

### LINKED ISSUE #2200: Actually remove followers if they are deleted remotely

We are currently not handling remote `Delete` ActivityPub activities and we should be. Since our implementation of ActivityPub doesn't deal with threads and account profiles, just the "Followers" it's not the end of the world, but it does end up with an incorrect list of followers if one of them deletes their account and we don't reflect that.

brokenintuition:
Fixes https://github.com/owncast/owncast/issues/2200

Marking this as a draft since I haven't tested this yet (was planning on setting up a test mastodon instance to try this out, but haven't gotten to it yet)

@gabek does it seem sufficient to just remove the follower? I was looking through what's persisted and didn't see anything else that would need to be deleted, but I might be missing something.

### DIFF
--- a/activitypub/inbox/delete.go
+++ b/activitypub/inbox/delete.go
@@ -2,24 +2,32 @@ package inbox
 
 import (
 	"context"
+	"net/url"
 
 	"github.com/go-fed/activity/streams/vocab"
 	"github.com/owncast/owncast/activitypub/persistence"
-	"github.com/owncast/owncast/activitypub/resolvers"
-	log "github.com/sirupsen/logrus"
 )
 
 func handleDeleteRequest(c context.Context, activity vocab.ActivityStreamsDelete) error {
-	// Only need to handle deletes for followers
-	if !activity.GetActivityStreamsObject().At(0).IsActivityStreamsPerson() {
-		return nil
+	obj := activity.GetActivityStreamsObject()
+	if obj != nil && obj.Len() > 0 && obj.At(0) != nil {
+		o := obj.At(0)
+
+		var actorIri *url.URL
+		switch true {
+		case o.IsActivityStreamsPerson():
+			actorIri = o.GetActivityStreamsPerson().GetJSONLDId().Get()
+		case o.IsActivityStreamsService():
+			actorIri = o.GetActivityStreamsService().GetJSONLDId().Get()
+		case o.IsActivityStreamsApplication():
+			actorIri = o.GetActivityStreamsApplication().GetJSONLDId().Get()
+		default:
+			// not a type we care about here
+			return nil
+		}
+
+		return persistence.RemoveFollowIRI(actorIri)
 	}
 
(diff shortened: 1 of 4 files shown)


activitypub/inbox/delete.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/go-fed/activity/streams/vocab"
+	"github.com/owncast/owncast/activitypub/persistence"
+	"github.com/owncast/owncast/activitypub/resolvers"
+	log "github.com/sirupsen/logrus"
+)
+
+func handleDeleteRequest(c context.Context, activity vocab.ActivityStreamsDelete) error {
+	// Only need to handle deletes for followers
+	if !activity.GetActivityStreamsObject().At(0).IsActivityStreamsPerson() {
```
gabek: I've probably used `At()` myself where I probably shouldn't, but if for some reason a `Delete` comes in and doesn't have an associated object attached this would crash.

activitypub/inbox/delete.go
```diff
@@ -0,0 +1,25 @@
...
+		return nil
+	}
+
+	actor, err := resolvers.GetResolvedActorFromActorProperty(activity.GetActivityStreamsActor())
```
gabek: Since all we want to do is remove the follow it seems like doing an additional resolution of the actor's details isn't needed. Maybe we could add a `RemoveFollowIRI()` and just pass in the IRI instead of an entire actor?

gabek:
> @gabek does it seem sufficient to just remove the follower? I was looking through what's persisted and didn't see anything else that would need to be deleted, but I might be missing something.

That should be it!

activitypub/inbox/delete.go
brokenintuition: Yeah makes sense, I'll do that. Thanks!

gabek:
One thing just came to mind when having a conversation about something else:

For `Delete`s, we need to make sure we aren't trying to actively verify that actor the same way we do we all other actions.

If a `Delete` action comes in from a deleted actor, and we go to verify it, the verification fails because that actor no exists, so we'll drop that activity and treat it as invalid.

brokenintuition:
@gabek that makes sense. Thanks for calling that out. I'm hoping to get more done on this either tonight or tomorrow

brokenintuition:
I've made some more progress on this. I ended up attempting to resolve any incoming message as a delete, because I'd rather check an explicit type, but it could be aliased and it looked like the code for pulling that out had a lot of non-exported pieces that would make it difficult.

This still doesn't validate the incoming delete request. If we want to do that by checking against a cached version of the public key, I think this is the list of necessary changes:

- Add column(s) to followers table for last known key
- Add a database migration to add the column for upgrade
- Update the follow and update follower handlers to populate the column
- For deletes, validate with the cached version if there, either accept or reject if not cached

I still need to go through the code to see what exactly would need to be stored for that to work, but does the general approach sound reasonable?

PR #2265, opened 2022-10-26
Fix display of auth modal. Re #2262.

### LINKED ISSUE #2262: Authentication modal doesn't layout properly when mobile

image

heller:
image


heller:
Getting the button to wrap to the next line was really complicated by all the spans and `display:table` styles, etc. We may want to change the search button from the magnifying glass icon to a different icon, or something short like "Go".

gabek:
Thanks for tracking it down!

And yeah, I'll dig through the icons and see if there's anything that makes sense, if not we can go the short text route.

PR #2267, opened 2022-10-26
On narrow screens, bump the footer links to a second row. Re #2260.

### LINKED ISSUE #2260: Mobile footer content doesn't fit properly

Maybe we should hide the "Powered by Owncast" link or something for mobile. I'm still on board not having the footer at all for mobile, but I know people weren't on board with removing it, so maybe we can just find a way to make it work.

image

heller:
There's a little bit of empty space on the right but it's not too bad.

image

image

### DIFF
--- a/web/components/ui/Footer/Footer.module.scss
+++ b/web/components/ui/Footer/Footer.module.scss
@@ -1,6 +1,7 @@
 .footer {
   display: flex;
   align-items: center;
+  flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


heller:
Hmm, I got a build failure from the linter and I'm not sure how to tell GitHub to retry the build.

gabek:
That works great, I didn't think about putting it on two lines. Thank you!

PR #2289, opened 2022-11-01
Sync between clients

SuperKenVery:
- Clients now play at the same time, perfect for an LAN party. 
- Corrected latency measurement: considered how much the current segment has played, rather than using its start time. 
- Refactored code, putting all `getCurrentlyPlayingSegment(tech)` into `/js/utils/helper.js`

### DIFF
--- a/controllers/playbackMetrics.go
+++ b/controllers/playbackMetrics.go
@@ -20,6 +20,7 @@ func ReportPlaybackMetrics(w http.ResponseWriter, r *http.Request) {
 	type reportPlaybackMetricsRequest struct {
 		Bandwidth             float64 `json:"bandwidth"`
 		Latency               float64 `json:"latency"`
+		MaxLatency            float64 `json:"maxLatency"`
 		Errors                float64 `json:"errors"`
 		DownloadDuration      float64 `json:"downloadDuration"`
 		QualityVariantChanges float64 `json:"qualityVariantChanges"`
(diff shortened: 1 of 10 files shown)


gabek:
I appreciate updating the latency calculation, but could you get me up to speed on what the "syncing" is? What problem does it solve? Could you point me to a discussion around it? Is there an issue filed?

SuperKenVery:
> What problem does it solve? 

IMG_0157


Hopefully this image helps explain. 

> Could you point me to a discussion around it? Is there an issue filed?

No discussion, no issue.

gabek:
> Hopefully this image helps explain.

Sorry to ask for more, but you're explaining what, you're not explaining how or why. If you're going to add a rather large change to the primary feature of a piece of software I think it requires some substantial discussion.

1. What happens when people aren't in sync? Does it just seek them to a new location?
2. What happens if one of the players is having network issues and is experiencing buffering? Do you just keep forcing their player to seek, causing more buffering, and causing them to be even less in sync causing an endless cycle of buffering and seeking?
3. How is the seek point decided?
4. How does it work with the latency compensator?

I don't think this should go in as it has more questions than answers, and is not a problem that I'm aware needs fixing or a feature anybody has ever asked for.

I believe a better solution is the latency compensator, where a player can optionally try to stay as close to live as possible. But it is localized to each player to determine how to do that given its own playback conditions.

Issue #2228, opened 2022-10-19
Admin: Fix antd deprecation warnings


gabek:
### Share your bug report, feature request, or comment.

`/admin` in the web app.

```Warning: [antd: Menu] children will be removed in next major version. Please use `items instead.
Warning: [antd: Modal] `visible` will be removed in next major version, please use open instead.
```

PR #2290, opened 2022-11-01
More renames from 'visible' to 'open'.

heller:


gabek:
Thank you! :100: 

For reference this is part of #2228.

PR #2294, opened 2022-11-02
Larger text box font to prevent iOS Safari's auto zoom.

### LINKED ISSUE #2276: When focusing on chat text field the window grows too wide
https://user-images.githubusercontent.com/414923/198814325-9f86a46f-7228-402c-8b82-ada5d142641b.MOV

heller:
Use 16px font size to prevent auto zoom on iOS Safari. Shorten the 'type here' message so it doesn't wrap to a second line when the chat pane becomes a column on the right side of larger screens. Re #2276.


gabek:
Thank you!

Issue #2112, opened 2022-09-14
Manually authored social posts are always set to followers-only


aral:
### Share your bug report, feature request, or comment.

Just upgraded from 0.0.11 to 0.0.12 using the official upgrade instructions at https://owncast.online/releases/owncast-0.0.12/ and fediverse posts are now set to followers only.

The private setting is off (I also toggled it on and off again just to see if that would override an errant setting somewhere.)

![Screenshot from 2022-09-14 16-00-51](https://user-images.githubusercontent.com/41798/190192579-45835257-35d0-4cf9-b29f-312cbd4c51be.png)

## Logs following upgrade

```
info | 3:43:48 PM 09/14/2022 | Web server is listening on IP 0.0.0.0 port 8080.
-- | -- | --
info | 3:43:48 PM 09/14/2022 | The web admin interface is available at /admin.
info | 3:43:48 PM 09/14/2022 | Video transcoder started using x264 with 1 stream variants.
info | 3:43:48 PM 09/14/2022 | RTMP is accepting inbound streams on port 1935.
info | 3:43:48 PM 09/14/2022 | Migrating users. This may take time if you have lots of users...
info | 3:43:48 PM 09/14/2022 | Migrating database from version 3 to 5
info | 3:43:48 PM 09/14/2022 | Owncast v0.0.12-linux-64bit (2d44f8603577bf1eb125447b3894462fa1310960)
```

aral:
I’m getting very odd output for `select * from datastore;` on owncast.db:

```
sqlite> select * from datastore;
stream_key|-
            |2022-02-15 18:38:10
http_port_number|2022-02-15 18:38:10
rtmp_port_number|2022-02-15 18:38:10
logo_path|

          |2022-02-15 18:38:10
server_metadata_tags|,
                      |2022-02-15 18:38:10
server_summary|W
                |2022-02-15 18:38:10
server_welcome_message|)
                        |2022-02-15 18:38:10
server_name|
            |2022-02-15 18:38:10
extra_page_content|��
                     |2022-02-15 18:38:10
����|2022-02-15 18:38:10
HAS_POPULATED_DEFAULTS||2022-02-15 18:38:10
peak_viewers_overall|2022-02-15 18:39:10
peak_viewers_session|2022-02-15 18:39:10
server_url| 
            |2022-02-15 18:41:45
directory_enabled||2022-02-15 18:42:10
����|2022-02-15 18:46:39ants|
directory_registration_key|
[...]

PR #2318, opened 2022-11-08
activitypub: ensure public messages are public

jprjr:
Hi there! I noticed that the "compose a post to your followers" message wasn't working for users on my Pleroma instance. The note was being sent as a non-public announcement, which caused 2 issues:

1. Users can't re-toot/repeat/share etc.
2. Only one user on my instance would receive the update. On Pleroma, all the other posts to users inboxes would be discarded, since there was already an activity with the same ID value.

I'm unsure what the behavior was on Mastodon previously - but with this PR the public messages work as intended.

### DIFF
--- a/activitypub/outbox/outbox.go
+++ b/activitypub/outbox/outbox.go
@@ -173,6 +173,9 @@ func SendPublicMessage(textContent string) error {
 	activity, _, note, noteID := createBaseOutboundMessage(textContent)
 	note.SetActivityStreamsTag(tagProp)
 
+	note = apmodels.MakeNotePublic(note)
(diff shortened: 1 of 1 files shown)


gabek:
Hey there! I believe this should be resolved in #2112 that's in `develop` currently. Let me know if it's not working for you.

jprjr:
Hi @gabek - from looking at the code, I think that would fix it - it just looks like it's not on the `develop` branch yet

gabek:
Sorry about that, I went ahead and moved the commit to develop.

jprjr:
Awesome! Thank you so much!

PR #2347, opened 2022-11-15
Chat name & color modal

### LINKED ISSUE #2205: Design: Chat name+color change modal

This modal needs a design.

image

Component: https://owncast.online/components/?path=%2Fstory%2Fowncast-modals-name-change--basic

xarantolus:
Hi! This is a draft for the new chat name & color modal design discussed in #2205 

Current problems I'm still working on:
* The color selection menu overflows the modal and is hidden behind it, making it impossible to select some colors
* ~~The color is sent to the server, but reopening the modal still uses the initial color (a reload or rename after color changing solves this, as then user info is resent by the websocket). This is the same behavior as before, but I'd like to fix that :) Using useState with the `displayColor` similar to the displayName doesn't seem to persist the color either.~~  The backend now re-sends user info after a color change

I'm also not entirely sure how this dialog should look when logged in e.g. via FediAuth

### DIFF
--- a/core/chat/events.go
+++ b/core/chat/events.go
@@ -114,6 +114,10 @@ func (s *Server) userColorChanged(eventData chatClientEvent) {
 	if err := user.ChangeUserColor(eventData.client.User.ID, receivedEvent.NewColor); err != nil {
 		log.Errorln("error changing user display color", err)
 	}
+
+	// Resend client's user info with new color, otherwise the name change dialog would still show the old color
+	eventData.client.User.DisplayColor = receivedEvent.NewColor
+	eventData.client.sendConnectedClientInfo()
 }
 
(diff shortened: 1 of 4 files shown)


gabek:
Thanks so much for thinking through all this! I went ahead and filed a TODO #2348 to re-send client info after the user's color changed, just like when a name changes. Feel free to make that change yourself if you like, otherwise I can take care of it.

xarantolus:
The backend now resends user info after a color change (ref #2348). I hope `webv2` is the correct branch for this change, the develop branch doesn't have any code for color changes. Will look at the overflowing menu next :)

gabek:
Oh awesome! And yup, `webv2` is the right place. Thanks so much for continuing to work on this.

gabek:
Thanks so much for working on this!

PR #2352, opened 2022-11-17
Update example systemd service file to be far more secure.

captainwasabi:
Added some service options for security purposes.


gabek:
Sounds reasonable!

PR #2353, opened 2022-11-17
improve chat input accessibility

Abbe98:
While using an `<input>` element would be preferable, setting `aria-role="textbox"` and `aria-placeholder` allows one to improve the current situation without making any larger HTML or CSS changes.

### DIFF
--- a/webroot/js/components/chat/chat-input.js
+++ b/webroot/js/components/chat/chat-input.js
@@ -347,8 +347,9 @@ export default class ChatInput extends Component {
         >
           <${ContentEditable}
             id="message-input"
+            aria-role="textbox"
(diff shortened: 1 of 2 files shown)


gabek:
Hi there. This is great feedback! We can't use an input control because it doesn't support rendering HTML elements, something required for custom emoji functionality.

While I'll certainly merge this in for anybody who is puling the source, I wish you would have opened an issue about this because the web applications in `develop` is no longer being worked on and future releases will not be using it. All the work for the web interface is going into the `webv2` branch, so I could have pointed you over there.

While `webv2` is very much a work in progress, I encourage you to check it out and see if you can help with any accessibility issues over there. It'd be super helpful. I need all the help I can get with accessibility!

Abbe98:
@gabek no worries, this change didn't take much time.

I will have a look at `webv2`!

PR #2388, opened 2022-11-29
Stream keys / admin password split

### LINKED ISSUE #2320: Make streaming key and admin password different values
While the initial goal of Owncast is to be single-tenant software (one user per instance) I do think it's reasonable to share an instance with others. But I fear there will be significant downsides to having a streamer without access to the admin. They won't know about stream performance, errors, logs, etc.

My concern is seeing issues like "It's fine when I stream but when my friend streams there's problems" and I ask what the problems are and they say "I don't know, they can't access the admin".

It's just something to think about. But since enough people have mentioned it, I do think it's reasonable to make this change and add a new feature where streaming keys can be added and deleted decoupled from the admin password.

I anticipate being able to get this around version 0.2.0 after the major rebuild of the web interface is launched.

### Discussed in https://github.com/owncast/owncast/discussions/2319

<div type='discussions-op-text'>

<sup>Originally posted by **DidierMalenfant** November 11, 2022</sup>
Hey everyone,

Just starting to dip my toes in the owncast ocean and I have a server up and running on Digital Ocean to replace my Mixcloud account.

I have what may be a noob question (sorry if it is), can you separate the admin passwd from the stream key? If I want to share the stream key with other so they can stream on my server but I don't want them to administrate the server.

Thanks for all the hard work on all this btw, this is really really awesome...</div>

gabek:
- Allows for multiple stream keys
- No longer allows logging into the admin with stream keys

### DIFF
--- a/.github/workflows/hls-tests.yml
+++ b/.github/workflows/hls-tests.yml
@@ -27,7 +27,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
-          stable: 'false'
           go-version: '1.17.2'
       - name: Run HLS tests
         run: cd test/automated/hls && ./run.sh
--- a/config/config.go
+++ b/config/config.go
@@ -40,6 +40,9 @@ var BuildPlatform = "dev"
 // EnableAutoUpdate will explicitly enable in-place auto-updates via the admin.
 var EnableAutoUpdate = false
 
+// A temporary stream key that can be set via the command line.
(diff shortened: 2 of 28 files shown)


MFTabriz:
do we have to manually update `openapi.yaml` file?

gabek:
> do we have to manually update `openapi.yaml` file?

Eventually.

gabek:
For added detail in the checklist of things to do before a release it has update the OpenAPI spec, so I usually do it all at once.

PR #2391, opened 2022-11-29
Edit external actions

### LINKED ISSUE #1884: Ability to edit external actions
I think it would be great to be able to edit an external action instead of having to delete and re-apply it to change information.

An example that would make this useful is having a complicated url that you don't have on hand, but all you need to do is change the background colour because the text has become invisible (set background to white, but text is white and unchangeable).

xarantolus:
This PR fixes #1884 by adding the ability to edit existing external actions from the admin UI.

Before this, one would have to delete an action and add it again. Now one can just use the edit button.

Demo:

https://user-images.githubusercontent.com/32465636/204544312-7007b346-3eed-47a1-93b6-dfe1f6092cff.mp4

For this I edited (and renamed) the `NewActionModal` to also accept an optional old action for filling out default values. Upon saving changes, either the old action is replaced (if we were in edit mode) or a new action is added to the end of the array (as before).

Please feel free to point out any issues with the implementation. I'm not super familiar with React and the use of `useEffect` kind of feels weird to me (however, when using e.g. `const [actionUrl, setActionUrl] = useState(action?.url || '');` instead, the validation logic for the "OK" button didn't seem work correctly)

### DIFF
--- a/web/pages/admin/actions.tsx
+++ b/web/pages/admin/actions.tsx
@@ -34,12 +34,21 @@ interface Props {
 const ActionModal = (props: Props) => {
   const { onOk, onCancel, open, action } = props;
 
-  const [actionUrl, setActionUrl] = useState(action?.url || '');
-  const [actionTitle, setActionTitle] = useState(action?.title || '');
-  const [actionDescription, setActionDescription] = useState(action?.description || '');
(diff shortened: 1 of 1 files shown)


gabek:
This is great, thanks so much for adding this functionality! People will surely appreciate it.

PR #2392, opened 2022-11-29
activate save button on S3 bucket name change in admin

### LINKED ISSUE #2386: Changing S3 bucket name doesn't cause "Save" button to enable

Just noticed this while cloning my Owncast server to make a new one...

- Go into configuration panel
- Change S3 Bucket name
- Save button is disabled and can't be clicked

I'd expect that changing the bucket name would enable the Save button, but it does not.  The result is that I had to delete a character from one of the other fields, save that with the new bucket name, then put it back and save again.

I'm sure this one's an easy fix; probably just didn't consider that field when doing some scripting to auto-activate the Save button?

MFTabriz:
closes #2386


gabek:
Thank you!

PR #2408, opened 2022-12-05
validate response of federation APIs

MFTabriz:
- verify json responses are parsable
- set `Accept` request header and check `Content-Type` in response headers
- refuse empty and invalid webfinger queries more strictly (+tests)
- log the reasons for webfinger query rejections in debug mode
- validate server URL before setting it through API (+test failed validations with `failAdminRequest()`)
- bump supertest minor version
- update `test/automated/api/package-lock.json`

### DIFF
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -416,7 +416,8 @@ func SetServerURL(w http.ResponseWriter, r *http.Request) {
 	}
 
 	rawValue, ok := configValue.Value.(string)
-	if !ok {
+	serverHostString := utils.GetHostnameFromURLString(rawValue)
+	if !ok || serverHostString == "" {
 		controllers.WriteSimpleResponse(w, false, "server url value invalid")
 		return
 	}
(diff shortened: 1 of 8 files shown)


controllers/admin/config.go
```diff
@@ -416,7 +416,8 @@ func SetServerURL(w http.ResponseWriter, r *http.Request) {
 	}
 
 	rawValue, ok := configValue.Value.(string)
-	if !ok {
+	serverHostString := utils.GetHostnameFromURLString(rawValue)
```
gabek, line 424: If `ok` is false then `rawValue` I'm assuming is an empty string. I'm not sure what happens when an empty string is passed to `GetHostnameFromURLString`.

MFTabriz: it calls net/`url.Parse()` which returns an empty string and no error

MFTabriz: I separated the checks just in case.

gabek:
Thank you for adding all these!  :1st_place_medal:

PR #2410, opened 2022-12-06
Fix URL Modal scrolling and size

### LINKED ISSUE #2395: Default modal width is too narrow

image

americobarros:
# Description

Fixes #2395 (Default modal width is too narrow)

Fixes the double scrollbar issue when using the Modal with an `iframe` inside. The double scrollbar was caused because `display` types were not working too well with each other. Setting the display of the `iframe` to `block` instead of the default `inline-block` does the trick.

 Also set a larger default size for the modal when the modal contains an iframe. From `width:520px` and no defined height, to `width:900px` and `height:70vh`. Width still has a `max-width:100vw-16px` as set by Ant Modal.

Changes were made based on [comments on the original issue](https://github.com/owncast/owncast/issues/2395#issuecomment-1338289951).

### DIFF
--- a/web/components/ui/Modal/Modal.tsx
+++ b/web/components/ui/Modal/Modal.tsx
@@ -42,6 +42,7 @@ export const Modal: FC<ModalProps> = ({
       sandbox="allow-same-origin allow-scripts allow-popups allow-forms"
       frameBorder="0"
       allowFullScreen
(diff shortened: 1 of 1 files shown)


gabek:
This works great, thank you for fixing these two things with the modal. Much better!

PR #2411, opened 2022-12-06
Feature: emoji editor

### LINKED ISSUE #2378: Allow uploading custom emoji through the admin

It would be useful for people to have a way to manage custom emoji images through the admin instead of through the filesystem of the machine Owncast is running on.

- Upload new images
- Delete unwanted ones

Note: Starting with`webv2` emoji images are pulled from `data/emoji` and no longer the old `webroot`.

xarantolus:
This PR adds a custom emoji upload page in the admin interface and fixes #2378.

The page lists all emojis from the custom emoji directory, allowing uploads (I reused/refactored some code from the logo upload) and deletion of uploaded emojis. ~~It adds a new configuration value to the database (whether to use custom uploaded emojis or the default ones), as far as I see it's not necessary to migrate this change; it just defaults to `false` if the config value hasn't been set yet.~~
~~I also added an API route that serves the uploaded emojis (as the normal API route wouldn't serve them if our uploaded emojis aren't enabled). While I would like to have admin authentication for this, it would require larger changes to make it work; when adding the admin auth requirement to the `/api/admin/img/emoji` route we can't really use the images in an `` tag due to basic auth missing there.~~

Here is a small demo of the page:

https://user-images.githubusercontent.com/32465636/205880048-4458f13b-a6ca-41e1-a623-f6b439c2faea.mp4

Some thoughts:
* Custom emojis are still in the recents picker for users that have sent them before, even after deletion (maybe removing removed emojis from there could make sense)

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -25,6 +25,6 @@ var (
 	// HLSStoragePath is the directory HLS video is written to.
 	HLSStoragePath = filepath.Join(DataDirectory, "hls")
 
-	// CustomEmojiPath is the optional emoji override directory.
+	// CustomEmojiPath is the emoji directory.
 	CustomEmojiPath = filepath.Join(DataDirectory, "emoji")
 )
--- a/config/constants.go
+++ b/config/constants.go
@@ -9,8 +9,8 @@ const (
 	FfmpegSuggestedVersion = "v4.1.5" // Requires the v
 	// DataDirectory is the directory we save data to.
 	DataDirectory = "data"
-	// EmojiDir is relative to the static directory.
-	EmojiDir = "/img/emoji"
+	// EmojiDir defines the URL route prefix for emoji requests.
+	EmojiDir = "/img/emoji/"
 	// MaxUserColor is the largest color value available to assign to users.
 	// They start at 0 and can be treated as IDs more than colors themselves.
 	MaxUserColor = 7
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -792,3 +792,23 @@ func SetStreamKeys(w http.ResponseWriter, r *http.Request) {
 
 	controllers.WriteSimpleResponse(w, true, "changed")
 }
+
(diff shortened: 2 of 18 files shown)


gabek:
Just in case you missed it, it looks like the linter threw some warnings.

gabek:
This is super cool! Let me know when you want somebody else to give it a spin.

xarantolus:
I'm wasn't entirely sure how to implement using the default emojis, as I'm not super happy with the current approach of either using them or only uploaded ones. I'll try to implement your suggestion with copying on the first launch (or rather: if the emoji directory does not yet exist) if that's fine :)

xarantolus:
I have now implemented the logic for copying the static/embedded emojis on first launch/when the directory doesn't exist.
That way we no longer have the distinction between custom emojis and uploaded custom emojis. Users can now also delete the built-in emojis after they are copied to their emoji directory.

For users that already have emojis set up nothing should change, and for those who don't also nothing should change (before the default emojis would have been used directly, now on the next server start they will just be copied to disk and then be served from there).

It might make sense to add some documentation on where emojis are saved/how to quickly delete them all, as by default more than 100 emojis are copied (and deleting them via the admin UI would require that amount of clicks).

Also another thing to note is that deleted emojis are still in the recents picker of the emoji picker, that might be weird.

gabek:
This is great, thank you for putting so much thought into it! This approach works well and is easier to understand than sometimes there being an emoji directory and sometimes there isn't.

Agreed about documentation. This will be updated: https://owncast.online/docs/website/#custom-emoji

You're right the recent picker is kind of weird if deleted items are in there, but hopefully that scenario is rare enough.

Thanks again for all your help, people are really going to enjoy having this functionality from the admin and is one fewer thing you have to be knowledgeable about the command line or file transfers to accomplish. 

<blockquote><div> Owncast</div><div><strong><a href="/docs/website/">Web Site + Chat</a></strong></div><div>Customize your Owncast web page by adding additional content and links.</div></blockquote>

PR #2416, opened 2022-12-06
refactor chatuser api tests

MFTabriz:
- ban and un-ban both ipv4 and ipv6 of localhost (without explicit un-banning the ipv6 loopback, some tests afterwards may fail) [see #2415]
- cleanup the tests and use `test/automated/api/lib/admin.js` functions for admin API requests


gabek:
This is great, thank you!

PR #2443, opened 2022-12-13
Fix/issue#2441

### LINKED ISSUE #2441: Error when trying to display "More info" moderation menu from chat

image

dorj222:
Currently, there are two rendering/mapping error messages, which can be seen on Console Log. These two error messages are related to the "More info" moderation menu and are caused by React JS Unique key for each item requirement. I solved these error messages by providing unique keys, and the error messages that are related to rendering/mapping disappeared now. Fixing these error messages may fix issue #2441.

The error message: 
Screen Shot 2022-12-13 at 2 59 04 PM


gabek:
I noticed you had to randomly generate a key because the connected client ID wasn't surfaced in the API. So I just went ahead and made the ID get passed along so you can use that instead of randomly generating one.

https://github.com/owncast/owncast/commit/3abc7a3ab8b46973218b8c44599ca0d9729155c9

dorj222:
Oh great, thanks a lot! I will update the client ID key generation.

dorj222:
Hi there! I updated the rendering/mapping methods, and now they use the Connected Client ID as unique keys instead of randomly generated ones.

gabek:
Awesome, works great! Thanks so much for helping with this!

PR #2466, opened 2022-12-18
Add Level-One Header

### LINKED ISSUE #2446: Accessibility: Page should contain a level-one heading
## Page should contain a level-one heading

https://axe.deque.com/issues/a5f7b758-fc67-4ede-8393-7f2efa99e436

> Ensure that the page, or at least one of its frames contains a level-one heading
>
>To solve this problem, you need to...
> Fix the following:
> Page must have a level-one heading

image

Re: #1826

silksow:
Fixes https://github.com/owncast/owncast/issues/2446

### DIFF
--- a/web/components/ui/Header/Header.tsx
+++ b/web/components/ui/Header/Header.tsx
@@ -23,7 +23,9 @@ export const Header: FC<HeaderComponentProps> = ({
       <div id="header-logo">
         <OwncastLogo variant="contrast" />
       </div>
-			<h1><span id="global-header-text">{name}</span></h1>
+      <h1>
+        <span id="global-header-text">{name}</span>
+      </h1>
     </div>
     {chatAvailable && !chatDisabled && <UserDropdown />}
     {!chatAvailable && !chatDisabled && (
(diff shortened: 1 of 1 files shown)


gabek:
Looks like this change alters the appearance of the header text and rises it up, making it look a little goofy.

image

silksow:
This does resolve the semantics issue. Do you want me to open a styling issue for you?

gabek:
Since the change you made wasn't supposed to impact the styling you'll have to make sure it doesn't. It should look the same as it does right now after your change is applied.

web/components/ui/Header/Header.tsx
```diff
@@ -23,7 +23,9 @@ export const Header: FC<HeaderComponentProps> = ({
       <div id="header-logo">
         <OwncastLogo variant="contrast" />
       </div>
-      <span id="global-header-text">{name}</span>
+      <h1>
+        <span id="global-header-text">{name}</span>
```
gabek, line 27: I would venture to guess that for accessibility it might be better for the content to be directly inside the new h1 tags and not again wrapped with unnecessary tags. You might have to change the styling of the h1 tag to be `inline` but I'm just guessing. Edit the corresponding `.module` file to make any necessary styling changes if needed.

gabek:
Oh! And since you're new to this whole PR thing, you can see above that a bot made a commit to your change to format the code and it says "Prettified code!". You'll want to grab this change so do a `git pull` before you continue working on this PR.

gabek:
I went ahead and took care of this when working on some other header items.

PR #2477, opened 2022-12-21
fix platform logo width

### LINKED ISSUE #2472: Fix platform image link image warnings

> Image with src "/img/platformlogos/github.svg" has either width or height modified, but not the other. If you use CSS to change the size of your image, also include the styles 'width: "auto"' or 'height: "auto"' to maintain the aspect ratio.

image

MFTabriz:
closes #2472

### DIFF
--- a/web/components/ui/SocialLinks/SocialLinks.module.scss
+++ b/web/components/ui/SocialLinks/SocialLinks.module.scss
@@ -1,5 +1,5 @@
 .link {
-  width: 1.8em;
+  width: 30px;
   margin-right: 16px;
 }
 


gabek:
I haven't looked into the original error, but it said "If you use CSS to change the size of your image, also include the styles 'width: "auto"' or 'height: "auto"' to maintain the aspect ratio." Did you not have to set a value to `auto` to resolve it?

MFTabriz:
> I haven't looked into the original error, but it said "If you use CSS to change the size of your image, also include the styles 'width: "auto"' or 'height: "auto"' to maintain the aspect ratio." Did you not have to set a value to `auto` to resolve it?

The logos are square and [their size is set to 30*30px]( https://github.com/owncast/owncast/blob/e41f0ba30ec7c6e801f7faf7efe71d6e6e2ca8d7/web/components/ui/SocialLinks/SocialLinks.tsx#L27-L28) using the HTML attributes which is totally fine. [if I have understood it correctly,] The problem was that by CSS their widths were being overridden to 1.8em but not the height. And that was changing their aspect ratio. 
Of course we can set height to auto in CSS to resize them properly but I want to avoid additional resizing if possible.

MFTabriz:
You can also check the measurements in the storybook. they used to be 25.20 x 30 and now they are 30 x 30. Also, the browser is no longer issuing warning!

Issue #2437, opened 2022-12-12
Merge webv2 into develop


gabek:
- Update all the documentation and links that point to `webv2`.
- Set CI jobs to trigger on `develop` instead of `webv2` branch
- Update project Readme to point people to old tags if they're looking for the previous version of the codebase.
- Remove fetching contributors from the old admin repo to populate project home page.
- Update web/README.md to reflect the updated state of branches and "old vs. new" talk of the frontend.

gabek:
Targeting end of January for this, assuming everything gets in before then.

gabek:
Still planning the end of the month to merge `webv2` into `develop`. Not everything will be complete or bug-free by then but it'll at least kick off hopefully more testing and more eyeballs on it. I just hope nobody tries deploying it to production.

gabek:
I have merged `webv2` into `develop`. The `develop` branch is now again the primary place for development. webv2 is dead, long live webv2.

PR #2496, opened 2022-12-24
set a non-root user for containers

### LINKED ISSUE #2488: Dockerfile+Earthfile Docker Container User Not Set

The Dockerfile provided with the Owncast code runs the application with user root

The `Dockerfile` does not specify or create a new user for running the application server, and by default if no user is specified docker will run the application as the root user inside that container.

The application should instead use a user with the minimum required privileges, to ensure that the application can only access to the strictly required functionalities.

Recommendation

Run the application as a user without root capabilities by modifying the `Dockerfile` document.

MFTabriz:
- set container user to "owncast"
- activate container builds on git push and pull requests (push only scheduled builds to container registry)
- update dockerfile based on earthfile (+docker)
- add dockerfile build and misconfiguration test to ci
- consolidate into a single nightly container build workflow and script

**Note:** deployment to container registry has not been tested!


.github/workflows/javascript-formatting.yml
```diff
@@ -61,7 +61,6 @@ jobs:
           ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
           fetch-depth: 0
-          pull: --rebase --autostash
```
gabek, line 64: Why was this removed?

.github/workflows/container.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Checkout repo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Build and push
+        if: ${{ github.event_name == 'schedule' && env.GH_CR_PAT != null }}
+        env:
+          GH_CR_PAT: ${{ secrets.GH_CR_PAT }}
+          EARTHLY_BUILD_TAG: 'webv2'
```
gabek, line 49: If this is going to be merged into `webv2` then the builds in `develop` are going to keep running. Those jobs won't get overwritten until `webv2` get merged into `develop`, and when that happens there will be no `webv2` branch anymore. So we should either plan for when that merge happens, or plan for now when both nightly and webv2 builds are happening. This PR kind of is in a middle ground.

.github/workflows/container-lint.yml
```diff
@@ -0,0 +1,28 @@
...
+  trivy:
+    name: Dockerfile
+    runs-on: ubuntu-latest
+    container:
+      image: aquasec/trivy
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Check critical issues
+        run: trivy config --exit-code 1 --severity "HIGH,CRITICAL" ./Dockerfile
```
gabek, line 25: What's trivy?

build/release/container.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "Building container image ${EARTHLY_IMAGE_NAME}:${BUILD_TAG} ..."
+
+# Change to the root directory of the repository
+cd "$(git rev-parse --show-toplevel)" || exit
+if [ -n "${EARTHLY_BUILD_BRANCH}" ]; then
+    git checkout "${EARTHLY_BUILD_BRANCH}" || exit
+fi
+
+earthly --ci +docker-all --image="ghcr.io/owncast/${EARTHLY_IMAGE_NAME}" --tag="${BUILD_TAG}" --version="${VERSION}"
```
gabek: Since you generalized this script to just `container.sh` it implies it's used to build all the container images, but that's not the case since Github is hard-coded here. Can the script maybe be renamed `nightlies.sh` or something to give a hint to people looking at it that it's not used for production builds and only used for nightly test builds?

.github/workflows/javascript-formatting.yml
MFTabriz: For example [here](https://github.com/owncast/owncast/actions/runs/3791338784/jobs/6446694713#step:3:1):
image

.github/workflows/container-lint.yml
MFTabriz: I don’t have any specific preferences. Which dockerfile checker do you want to use?

.github/workflows/javascript-formatting.yml
gabek: Ah ha! I remember that working somewhere from something before but I must have mixed it up with maybe a different action.

.github/workflows/container-lint.yml
gabek: I was just curious what trivy was. So it's a linter for Dockerfiles? Does it support Earthfiles too?

MFTabriz: it is a security scanner https://aquasecurity.github.io/trivy/v0.35/ Other than dockerfiles, it supports many other formats but unfortunately it does not support Earthfiles yet.
It can detect if we use root as the user inside the container (the main issue addressed by this PR). It also had some recommendations for the dockerfile which I applied. See the report for non-critical issues [here]( https://github.com/owncast/owncast/actions/runs/3776695073/jobs/6420111585)

build/release/container.sh
MFTabriz: I renamed `build/release/container.sh` to `build/develop/container.sh`. I think this should make it clear enough. what do you think?

.github/workflows/container.yaml
MFTabriz: It is in the list for #2437
After the merge, all references to `webv2` in workflow files must be replaced with `develop`

build/release/container.sh
gabek: Maybe a comment on top would help. I just am afraid somebody will want to make changes to the release docker image and start editing this script and I'll have to tell them the script isn't even used, so I'd rather make it super clear.

gabek: I have the same concern with the Dockerfile vs. Earthfile.

MFTabriz: I added a note to the `Dockerfile`. Also made it clear in `build/develop/container.sh`.

.github/workflows/container.yaml
gabek: But before the merge these workflows won't even run, will they? Don't they need to be in the primary branch to run? Or that's what I assumed, I could be wrong I've never tried it. That's why I don't have `webv2` nightlies firing from the `webv2` branch.

MFTabriz: These should get triggered for the PRs to `webv2` too. (as they have ran for this PR) Only the deployment to GH part is skipped.

PS: see the job named "Build development container / Earthly (pull_request) " below.

Issue #2489, opened 2022-12-23
Use a constant time compare for rtmp secret check


gabek:
### Share your bug report, feature request, or comment.

Recommendation

Use a constant time compare, like [ConstantTimeCompare](https://pkg.go.dev/crypto/subtle#ConstantTimeCompare) from `crypto/subtle`.

The Owncast default configuration expose the RTMP port to the same network of the web page, depending on the user configuration that port could be exposed to the internet.

The Owncast server protects the access to the stream input by enforcing that the stream connection URL contains a secret key known only to the server administrator.

In the following snippet it is reported the function that perform the secret key check, taken from file `core/rtmp/utils.go` at line 83:

```go
func secretMatch(configStreamKey string, path string) bool {
	prefix := "/live/"

	if !strings.HasPrefix(path, prefix) {
		log.Debug("RTMP path does not start with " + prefix)
		return false // We need the path to begin with $prefix
	}

	streamingKey := path[len(prefix):] // Remove $prefix
	return streamingKey == configStreamKey
}
```

The last line of function `secretMatch` compare string `streamingKey`, submitted by the user with string `configStreamKey`, read by the system configuration, using the built-in `==` operator. This operator perform the comparison by checking one character at a time, from start to end, returning `false` as soon as a character is not equal.

Impact

An attacker with enough time and resources could implement a time based attack to guess the secret key. The success of time based exploit heavily depends on the server configuration such as resources available and the server processor architecture, as the implementation may vary depending on the byte code.

MFTabriz:
By using `ConstantTimeCompare` on two byte slices which are not equal in length, the size of password can be easily extracted! Just replacing `==` with `ConstantTimeCompare` does not solve the issue.

gabek:
Moving this out of the v0.1.0 milestone as nothing is functionally broken and this doesn't need to be required to ship the release. This will be worked on along with the admin password improvements in #2495.

PR #2511, opened 2022-12-26
Use bcrypt hash comparisons for streaming key (#2495)

### LINKED ISSUE #2495: implement admin password hashing
Using [crypto/bcrypt](https://pkg.go.dev/golang.org/x/crypto/bcrypt) or a similar method should be sufficient.

The implementation must get thoroughly tested and audited.

This issue supersedes #2489

roadriverrail:
The streaming key comparison was previously using the `==` operator between a plaintext version of the streaming key and the one configured and stored.  This comparison exposes the possibility of a time-based attack for discovering the streaming key (see issue #2489).

Project contributor @MFTabriz noted that a better solution would be to use bcrypt and compare hashes.  This commit switches the comparison over to it.

### DIFF
--- a/core/rtmp/rtmp.go
+++ b/core/rtmp/rtmp.go
@@ -9,6 +9,7 @@ import (
 	"github.com/nareix/joy5/format/flv"
 	"github.com/nareix/joy5/format/flv/flvio"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/crypto/bcrypt"
 
 	"github.com/nareix/joy5/format/rtmp"
 	"github.com/owncast/owncast/core/data"
@@ -75,7 +76,14 @@ func HandleConn(c *rtmp.Conn, nc net.Conn) {
 		return
 	}
 
-	if !secretMatch(data.GetStreamKey(), c.URL.Path) {
+	streamKeyHash, err := bcrypt.GenerateFromPassword([]byte(data.GetStreamKey()), bcrypt.DefaultCost)
+
+	if err != nil {
+		log.Errorf("Error hashing streaming key, cannot proceed: %s", err)
+		_ = nc.Close()
+	}
+
+	if !secretMatch(streamKeyHash, c.URL.Path) {
 		log.Errorln("invalid streaming key; rejecting incoming stream")
 		_ = nc.Close()
 		return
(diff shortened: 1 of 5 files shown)


MFTabriz:
Thanks for this PR @roadriverrail ! Can you please rebase this on `webv2` branch?

I prefer to have the hashed password instead of plain text in Datastore. But I know it involves more changes to the code and also we won’t be able to show the password to the admin/streamer afterwards. What do you think @gabek ?

gabek:
Yeah, the real benefit would be not having the plain text values in the database. So a few different things would be required.

1. Update the [stream key and admin password migration ](https://github.com/owncast/owncast/blob/webv2/core/data/datastoreMigrations.go#L57) to store the hashed versions instead.
2. Update the `-streamkey` and `-adminpassword` command line flags to set hashed versions of those values.
3. Remove showing the stream keys in the Stream Keys admin page.
4. Remove showing the admin password in the Server Setup admin page
5. Let anybody doing development on v0.1.0 know that they should delete their `data` directory because they'll have unhashed values there that will fail. Alternatively we could add another migration that runs after step 1 above to hash the values, but since not that many people have touched v0.1.0 it's probably just easier to wipe the directory.

roadriverrail:
Honestly, as I was writing this up, my assumption was that the conversation would end up going in this direction, since it's what's considered standard practice.  However, I wasn't going to do a bunch of work and break a user-facing feature without getting feedback first, and I figured the PR was the best way to do that.

I think I should be able to get to the outlined steps in my week off.  I have practically no web dev experience, but I'm guessing removing some existing fields is easy.

This will also require a documentation change, I think.  It's been a minute, but if memory serves, the documentation mentions the stream key and admin password web UI features and doesn't really mention other ways to access or alter them.  I didn't even know there were command line flags for that!

Also, if we go the recommended route on (5), how will this impact existing users who upgrade?

Also also, I noticed that the `webv2` branch includes support for multiple streaming keys, so rather than rebase existing work, I'll just do a rewrite based on it.

roadriverrail:
Abandoning in favor of a new PR on the correct branch.

MFTabriz:
Just as a note: you could click edit and change the base of this PR on GH. Rebase your branch on `webv2` locally. And, force push it again.

roadriverrail:
> Just as a note: you could click edit and change the base of this PR on GH. Rebase your branch on `webv2` locally. And, force push it again.

Yep, but there really didn't feel like a good reason to do so in this case.  I was changing branches *and* radically altering the work done.

PR #2518, opened 2022-12-27
Migrate to hashed password/key store (#2489)

roadriverrail:
In order to improve Owncast security, the admin password and stream keys are no longer being stored in plaintext, instead using a standard cryptographic hashing process to store them securely.

This commit not only moves the admin password to storage as a hash, but also splits the StreamKey model into StreamKeyPlainText and StreamKeyHashed to make it clear when, where, and how a stream key is being used in its plaintext and hashed forms.  Attempting to store a plaintext key to the database will automatically hash it.

Since it is no longer possible to send the plaintext admin password and streaming keys to the web UI, they have been removed from the relevant HTTP response objects.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,7 +41,7 @@ var BuildPlatform = "dev"
 var EnableAutoUpdate = false
 
 // A temporary stream key that can be set via the command line.
-var TemporaryStreamKey = ""
+var TemporaryStreamKey []byte = nil
 
 // GetCommit will return an identifier used for identifying the point in time this build took place.
 func GetCommit() string {
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/owncast/owncast/models"
+	"golang.org/x/crypto/bcrypt"
 )
 
 // Defaults will hold default configuration values.
@@ -20,8 +21,6 @@ type Defaults struct {
 	WebServerPort    int
 	WebServerIP      string
 	RTMPServerPort   int
-	AdminPassword    string
-	StreamKeys       []models.StreamKey
 
 	YPEnabled bool
 	YPServer  string
@@ -37,16 +36,12 @@ type Defaults struct {
 }
 
 // GetDefaults will return default configuration values.
-func GetDefaults() Defaults {
-	return Defaults{
+func GetDefaults() *Defaults {
+	return &Defaults{
 		Name:                 "New Owncast Server",
 		Summary:              "This is a new live video streaming server powered by Owncast.",
 		ServerWelcomeMessage: "",
 		Logo:                 "logo.svg",
-		AdminPassword:        "abc123",
-		StreamKeys: []models.StreamKey{
-			{Key: "abc123", Comment: "Default stream key"},
-		},
 		Tags: []string{
 			"owncast",
 			"streaming",
@@ -91,4 +86,23 @@ func GetDefaults() Defaults {
 		FederationUsername:      "streamer",
 		FederationGoLiveMessage: "I've gone live!",
 	}
+}
+
+func (d *Defaults) GetAdminPasswordHash() ([]byte, error) {
+	return d.getDefaultPasswordHash()
+}
+
+func (d *Defaults) GetStreamKeysHashed() ([]models.StreamKeyHashed, error) {
+	hash, err := d.getDefaultPasswordHash()
+	if err != nil {
+		return nil, err
+	}
+	return []models.StreamKeyHashed{
+		{Key: hash, Comment: "default stream key"},
+	}, nil
+}
+
+func (d *Defaults) getDefaultPasswordHash() ([]byte, error) {
+	hash, err := bcrypt.GenerateFromPassword([]byte("abc123"), bcrypt.DefaultCost)
+	return hash, err
 }
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -208,7 +208,7 @@ func SetAdminPassword(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := data.SetAdminPassword(configValue.Value.(string)); err != nil {
+	if err := data.SetAdminPasswordPlainText(configValue.Value.(string)); err != nil {
 		controllers.WriteSimpleResponse(w, false, err.Error())
 		return
 	}
@@ -775,7 +775,7 @@ func SetStreamKeys(w http.ResponseWriter, r *http.Request) {
 	}
 
 	type streamKeysRequest struct {
-		Value []models.StreamKey `json:"value"`
+		Value []models.StreamKeyPlainText `json:"value"`
 	}
 
 	decoder := json.NewDecoder(r.Body)
@@ -785,7 +785,7 @@ func SetStreamKeys(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := data.SetStreamKeys(streamKeys.Value); err != nil {
+	if err := data.SetStreamKeysPlainText(streamKeys.Value); err != nil {
 		controllers.WriteSimpleResponse(w, false, err.Error())
 		return
 	}
--- a/controllers/admin/serverConfig.go
+++ b/controllers/admin/serverConfig.go
@@ -49,8 +49,6 @@ func GetServerConfig(w http.ResponseWriter, r *http.Request) {
 			AppearanceVariables: data.GetCustomColorVariableValues(),
 		},
 		FFmpegPath:              ffmpeg,
-		AdminPassword:           data.GetAdminPassword(),
-		StreamKeys:              data.GetStreamKeys(),
 		WebServerPort:           config.WebServerPort,
 		WebServerIP:             config.WebServerIP,
 		RTMPServerPort:          data.GetRTMPPortNumber(),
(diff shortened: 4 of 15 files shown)


roadriverrail:
This replaces PR #2511 

This should cover the data model for storing hashes securely and converting new plaintext passwords from the web UI into hashed passwords for storage, or at least it should be headed in the right direction for that.

I haven't changed anything at the web UI layer because React is way, way outside my base of experience and, even if I did find a way to remove rendering of the password/keys, there are likely UX implications I haven't considered and wouldn't handle elegantly.  I'm happy to take a patch or solid advice on that front, though.

aaronpk:
Sorry, I haven't been following this too closely, but can I just chime in from the UX side for a minute?

It is extremely common in nearly every streaming platform to be able to retrieve the stream key from the admin interface. I would hate to lose that ability here and force everyone to copy the stream key the very first time they set up owncast into some password manager.

If we're making changes at this level, then my ideal scenario would be:

* The admin password should be treated as a password, storing only a hash in the DB, and should not be used as a stream key
* The stream key should be stored encrypted in the DB, but able to be decrypted to show the plain text in the admin interface when needed
* It should be possible to generate a new stream key in the admin interface, which rotates the stream key but doesn't affect the admin password

This also opens up future possibilities for alternative admin authentication options like using a passkey or even integrating with an OAuth provider, but that is a separate discussion.

roadriverrail:
I agree that it seemed pretty weird to me to lose the ability to see the streaming keys through the admin interface, too, so I'm definitely not against that.

That said, it's worth remembering that the original goals here were the following:

1. Prevent a time-based attack based on the fact that `==` has variable return times, allowing a possible attacker to determine the value of the streaming key
2. To prevent a similar time-based attack that allows determining the length of the streaming key via exploiting variability in the `ConstantTimeCompare( )` function.

Part of the point of moving to the hashing and comparison functions in `bcrypt` is that we do get a constant-time comparison in addition to also making the database more secure.  Personally, I never saw storing the streaming keys in plaintext as the central problem; it just was pointed out as worth doing in #2511 

If we do move to encrypted store of the streaming keys, don't we then subsequently need to pose questions about the security of the keys used to encrypt them?  Seems to me we'd then need to generate keys that are protected by a password?  And then changing that password would invalidate all the streaming keys?

roadriverrail:
I'll add that if we do want to just handle the question of time-based attacks against the streaming keys and save their secure storage for a later day, then I'd propose we do something like this: https://go.dev/play/p/NU5uTaB-sp

This plugs up the time-based length discovery as mentioned in (2).

<blockquote><div><strong><a href="https://go.dev/play/p/NU5uTaB-sp">Go Playground - The Go Programming Language</a></strong></div></blockquote>

gabek:
As of v0.1.0 stream keys and the admin password are split up, and you can create/delete new keys as needed, so maybe we're half way there.

If displaying the plain text stream keys is a higher priority than keeping them secure on disk then we can continue to keep them stored in plain text and only worry about the admin password.

However, it does seem like since you can create new streaming keys that there's very little downside to losing an old one. While it might be common to display these keys in the UI maybe it shouldn't be, they are passwords after all.

I don't feel strongly one way or another, but I guess I don't understand why we wouldn't want to take steps to secure all passwords, including stream keys.

aaronpk:
That's great, definitely more than half way there. I mainly wanted to point out that it is extremely common in other platforms to let the user retrieve and copy the existing stream key, without having to make a new one.

Twitch
image

YouTube
image

Restream
image

While this is technically less secure than storing only a non-reversible hash of the stream key, it is by far the most common behavior, and likely what any streamer picking up owncast would expect.

roadriverrail:
I actually want to note that, if we store only the hash of a stream key, then pretty much all acts of stream key management become two operations: (1) create a key and hope you remember it (2) delete the key identified by a given comment.  Right now, this doesn't even have a way for you to use a command line to retrieve your streaming keys.  I haven't used v0.1.0 yet (I was mostly trawling for a 1-3 day project that was mostly golang), so I don't even know what the interface for making new keys is.  Unless there's some sort of purely client-side way of showing the key at the moment of creation (e.g. freeform text entry, random string displayed one time, etc), this is pretty prohibitive.

I use OBS to interface to 3 different streaming servers.  2 of them are my own Owncast instances and 1 of them is JNKTN.TV.  I switch keys in OBS quite often, and I do often look up my own streaming keys via the Owncast web UI.  I could record these somewhere else, of course, and that wouldn't be a big deal if I were simply used to the workflow.  That said, it's so common on Twitch and YouTube that I think most people making the switch would expect it.

This is why I think it's useful to return to what the original issue was describing.  We'd started this with trying to prevent a timing attack whereby the attacker uses the response time of Owncast to try and determine a streaming key so they could then hijack the streaming services of an Owncast instance.  By preventing this, we're pushing the attacker towards having to trying to leak the data store through a web request or to compromise the shell access.

Where I could see a bigger point to having an encrypted store of the streaming keys is the case where one Owncast is managing tens or hundreds of streaming keys, because at that point we're talking about one exploit impacting a large user base.  But, as I mentioned above, I think if we're going down that road, we also need to ask the question about how the encryption keys are protected, because encryption is only as good as the key management.  Just keeping keys laying around on the filesystem is easy but doesn't help much (just changes the shape of the attack).  I could speculate that, at the point you're managing a large number of streaming keys for multiple streamers, you're beginning to veer into a territory of systems design Owncast isn't currently trying to shoot for.

Security is, at best, a tertiary subject for me, so I may be missing something here.

MFTabriz:
Let me summarize my understanding of discussion and important distinctions between them: 

1. Fix current implementation with `crypto/subtle#ConstantTimeCompare` for checking admin password/streamkeys using for example https://go.dev/play/p/NU5uTaB-sp
    - closes #2489 
    - Passwords are saved as plain text in database
1. Generate a hash from the admin password/streamkeys in database using `crypto/bcrypt#GenerateFromPassword` and compare that hash with provided password using `crypto/bcrypt#CompareHashAndPassword`
    - closes #2489 
    - Passwords are saved as plain text in database
1. Use `crypto/bcrypt` but save hash of the admin password/streamkeys in database
    - closes #2489 (similar to option#2) as well as #2495
    - Admin cannot see the streamkeys but can reset them if forgotten
    - Passwords are saved as hashed in database
 1. Implement option#3 for admin password and implement option#1/2 for streamkeys.
    - closes #2489 but only partially addresses #2495
    - Admin can see the streamkeys
    - Passwords are saved as hashed but streamkeys are saved as plain text  in database

For solving just #2489, option#1 seems like a quick fix.
For solving #2495, I don’t see a perfect solution here. From the usability perspective, option#4 seems like an acceptable compromises if we can’t find any better solution.

[PS: updated!]

<blockquote><div><strong><a href="https://go.dev/play/p/NU5uTaB-sp">Go Playground - The Go Programming Language</a></strong></div></blockquote>

gabek:
It does look like option 4 might be the compromise since it sounds like the user experience of having those keys easily available is the most important consideration. And that's fine, it just has to be a conscious trade-off since we're talking about it for the first time and we should be sure about what we're OK with.

MFTabriz:
Personally, I prefer option#3. If I don’t remember a key and have to look it up at the admin panel, I should be able to set a new one in that panel instead with no hassle.

roadriverrail:
If option 3 is pursued, we still need a way to display the streaming keys at least once, yes?

gabek:
Right now it looks like this:

image

roadriverrail:
In all honesty, while I do believe pursuing option 3 may represent a point of friction for streamers making the switch from Twitch and other services, I don't feel so strongly about it to resist it.  One thing I think would greatly improve the UX, if we're going for it, is some kind of button that'll generate and display a random streaming key.  Free-form text entry doesn't make strong passwords and it's tedious.  But,, again, I won't fight it.  These are all things that can be well-overcome with proper documentation and user prompting.

The code in my PR represents pursuit of option 3.  I will be up-front that I don't think I'll have the time to invest in implementing an option 4 type of scheme.  I do think I'd have time to switch up to something like option 2 or even a hybrid (e.g. we hash the admin password but not the keys).  I start losing my bandwidth for recreational coding on January 3.

gabek:
The admin password form has a button that auto-generates a random string and with #2528 I plan on doing the same thing for the new stream keys UI, while adding the enforcement of some kind of basic complexity requirements.

roadriverrail:
Cool!  Like I said, I don't love it, but I also don't hate it, and I think it's something everyone could learn to live with.  Like I said, if all concerned can tolerate option 3, that's what this code should be providing.

gabek:
I'm with you. Since @aaronpk is the streaming workflow expert I know he represents what most people want and expect.

Since I don't feel strongly one way or the other I'd rather go the route where it doesn't upset people and have to change it later :). Though I do wonder how this became the expected practice to be able to streaming access keys in plain text via the UI.

Plus if we end up feeling strongly in the future it'll be easier to encrypt the plain text keys and hide them from the UI than it would be to go the other direction.

aaronpk:
I'm definitely in a weird place with this discussion.

With my day job security hat on, I would say definitely treat the stream key as a password, storing only the hash, make it non-retrievable after it's generated, etc etc. 

But with my streaming hat on, I know that most other sites don't work that way, so you'll almost certainly get users complaining or asking support questions if they aren't able to copy the stream key from the admin interface.

roadriverrail:
So, it seems we've had a good discussion of the pros and cons but I don't feel like I see a settled consensus here.  Should I remove hashing of the keys and switch over to the trick for guaranteeing a constant time compare instead?

gabek:
Yeah, option 4 I think makes sense to go with. If we feel like the opinions about retrievable stream keys change in the future then we can then go forward with encrypting the stream keys on disk at that time. Like Aaron said, being a bit conservative on this front will probably save support requests and comparisons to other interfaces they're used to.

roadriverrail:
That makes reasonable enough sense.  I'm going to be up-front and say that I've never worked on anything involving the creation and management of encryption keys in a programmatic way, so I think the needs for this feature probably surpass my ability to deliver, especially on a quick "one day project" sort of thing.  Perhaps the PR should be closed and a new issue created that encapsulates the needed design for option 4.

gabek:
Good call. I'll go ahead and close this and open a new issue with detailed specifics.

Issue #2480, opened 2022-12-22
Limit logo file size on upload


gabek:
### Share your bug report, feature request, or comment.

Somebody tried to upload an updated logo but it was rejected server-side as being too large and as far as I can tell it didn't even fire the Owncast upload controller because it was rejected somewhere else along the way.

The client should check for a reasonable file size and throw a client-side error rejecting it so an upload doesn't even get attempted for a image that is determined to be too large.

https://owncast.rocket.chat/channel/general?msg=7GSNLKvfWkSr2ZDSd

andrew-secret:
I can take a look at this one 🙂. Just let me know which file size it should have. I think that's a quick one (just adding a size attribute to the input element). But let's see 🫣

gabek:
Thank you! How about 2 megs? I can't see any use case where your logo needs to be any larger than that, even that would be big in my opinion.

PR #2522, opened 2022-12-28
Filesize limit for logoupload

andrew-secret:
This PR closes https://github.com/owncast/owncast/issues/2480

### Changes made:

- limit file size to 2 MB before API call is fired to the backend
- add helper function to display file size in a proper size unit
- adjust wording below logo upload

# Upload failed with:
image

--

I couldn't upload a real image on localhost, maybe someone can assist me to test it properly. But in theory it should work.


gabek:
Super weird, you're right, the actual uploading of a logo is flaky. It's been fine previously as far as I know. It seems like sometimes it just doesn't upload any data:

image

I went ahead and added some error handling so it at least returns something and doesn't just break the backend when testing this PR.

web/components/config/EditLogo.tsx
```diff
@@ -43,10 +48,20 @@ export const EditLogo: FC = () => {
 
   // validate file type and create base64 encoded img
   const beforeUpload = (file: RcFile) => {
+    console.log({ file });
```
gabek: Remove this before merging in.

gabek:
Filed #2524 to look at the issue with uploading the actual image.

web/components/config/EditLogo.tsx
andrew-secret: Ah good catch. Done ✅

gabek:
Thanks for adding this file size check and messaging around it, that'll certainly clear things up for people in the future.

PR #2531, opened 2022-12-29
wrap promise response in set timeout to keep local state in sync

### LINKED ISSUE #2524: Admin logo upload UI sending empty payloads

I found when testing #2522 that the web frontend is sometimes sending empty payloads to the backend when trying to upload an image. It feels random. Sometimes it works, sometimes it doesn't. It may be due to types of file, or name, or permissions or something but it doesn't seem like it's related to the backend.

image

andrew-secret:
This PR closes https://github.com/owncast/owncast/issues/2524

### Changes made
- wrap respone into a `setTimeout` in `beforeUpload` hook to keep local state in sync

It works every time now, but I'm not pleased with this solution. I couldn't figure out a better way to solve this. The root cause of the flakiness is that the local state (`const [logoUrl, setlogoUrl] = useState(null);`) is not updated at that time when the customRequest (`handleLogoUpdate`) function is fired.

Maybe we can find a better solution here. 🥲

### DIFF
--- a/web/components/config/EditLogo.tsx
+++ b/web/components/config/EditLogo.tsx
@@ -71,7 +71,7 @@ export const EditLogo: FC = () => {
 
       getBase64(file, (url: string) => {
         setlogoUrl(url);
-        return res();
+        setTimeout(() => res(), 100);
       });
     });
   };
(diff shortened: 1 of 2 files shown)


gabek:
The only more "proper" way to pull it off would be to set an `useEffect` hook on `logoUrl` and don't fire `res()` until that hook fires. But that would be messier than his is, as we'd have to keep a reference to `res()`.  I think this workaround, while not pretty, can do for now!

Thank you for tracking this down, that was a weird one.

PR #2533, opened 2022-12-30
cleanup automated tests for public dir

MFTabriz:
- fix `randomString()`
- do not suppress unhandled errors
- create public directory if it does not exist
- validate content of server response
- remove the temporary file after the tests


gabek:
Sweet, thank you!

PR #2538, opened 2022-12-31
Fix twitter save button 2534

### LINKED ISSUE #2534: Show Twitter notification save button when enabling/disabling

though the admin panel, you can enable and disable the twitter integration.  If you were to flip it off and later turn it back on, the settings will return and it will appear like it has been enabled, but it doesn't actually save unless you also edit one of the text blocks.  

image

Currently the workaround I've found is editing the title to something else and back will show the save button.

image

I wouldn't think it would be too hard to add the switch changing as a case to make the save button appear.

hsingh124:
- [ ] Closes #2534

This updates the logic of the canSave method in twitter.tsx file to allow saving regardless of enabled's value.

https://user-images.githubusercontent.com/44574494/210133252-abf92113-c8c6-4a40-8c03-8d4fb9a968ae.mov

### DIFF
--- a/web/components/config/notification/twitter.tsx
+++ b/web/components/config/notification/twitter.tsx
@@ -53,18 +53,10 @@ export const ConfigNotify = () => {
   }, [twitter]);
 
   const canSave = (): boolean => {
-    const {
-      enabled,
-      apiKey,
-      apiSecret,
-      accessToken,
-      accessTokenSecret,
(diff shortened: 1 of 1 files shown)


gabek:
Stupid code scanning, ignore that.

Thank you so much for putting in this fix! Works as expected! :100:

Issue #2500, opened 2022-12-25
On new messages chat does not scroll to the very bottom


gabek:
### Share your bug report, feature request, or comment.

When new messages come in the chat container does not scroll to the very bottom.

![image](https://user-images.githubusercontent.com/414923/209457603-2d29ad78-27b5-4b3c-b489-420984ee50c0.png)

ShashwatAgrawal20:
Would like to work on this

gabek:
Thank you!

ShashwatAgrawal20:
@gabek quick question 
how can I get this chat feature in the development environment

gabek:
You'll want to start a test stream. You can point any streaming software you'd like at your local Owncast instance. https://owncast.online/development/#run-a-development-stream

<blockquote><img src="/images/owncast-logo-1000x1000.png" width="48" align="right"><div><img src="/favicon.png" height="14"> Owncast</div><div><strong><a href="/development/">How to work on Owncast</a></strong></div><div>The technical details for those wishing to take part in Owncast development.</div></blockquote>

ShashwatAgrawal20:
@gabek the new messages are scrolling to the bottom by default. It doesn't scroll to the very bottom if the user has scrolled up, I think it's a good option to keep it as it is.

gabek:
I started making some changes to work on this the other day, so I'll go ahead and assign to myself.

gabek:
I don't believe this has actually been improved.

![image](https://github.com/owncast/owncast/assets/414923/6263259e-6faf-4106-937e-94ec48a2c714)

gabek:
Closing this since chat has been rebuilt, and other issues around chat have been filed.

Issue #2342, opened 2022-11-13
"Go to last message" button covers chat messages only on production builds


gabek:
### Share your bug report, feature request, or comment.

We might just need to tweak the tolerance of when this button appears.

![image](https://user-images.githubusercontent.com/414923/201539867-bae62761-b363-43bb-a5bb-d01e07b9672e.png)

gabek:
This is still occurring.

![image](https://user-images.githubusercontent.com/414923/207252668-640a3c8e-e4f8-4ffd-bd06-c3f831e5a0d7.png)

gabek:
So this is happening 100% of the time on production builds, but not at all in local dev builds.

production: (localhost:8080 or https://webv2.owncast.tv for example)

![image](https://user-images.githubusercontent.com/414923/208019539-2a7e733c-aca1-49bb-9078-19f58bf169a2.png)

local dev:

![image](https://user-images.githubusercontent.com/414923/208019177-2c3f723f-1108-4791-b111-b830e912da30.png)


<blockquote><img src="https://webv2.owncast.tv/thumbnail" width="48" align="right"><div><img src="/img/favicon/favicon-16x16.png" height="14"> Owncast Web v2</div><div><strong><a href="https://webv2.owncast.tv">Owncast Web v2</a></strong></div><div>This is the work in progress Webv2 of Owncast's user interface.</div></blockquote>

gabek:
I have not yet been able to figure this out. It works fine in dev, but is broken in production.

gabek:
Addressing this with the production build "bottom" issues in #2500.

gabek:
This still sometimes happens. :(

gabek:
This is still a problem.

gabek:
@t1enne mentioned to me he wanted to look at this, so I went ahead and assigned.

PR #2539, opened 2022-12-31
Add `initiallyMuted` query parameter to embed player

### LINKED ISSUE #2420: Option to mute embed by default

It would be great if there was a way to feed a mute parameter to an embeded player so the player could run muted by default until the user unmutes it.

Maybe I'm missing something and this feature already exists, but I don't see anything in the documentation so I figured I'd request it.

LBBO:
This PR introduces a new `initiallyMuted` query parameter that can be added to the embed URL. If set to exactly `true`, the video will be muted once it is played. It can still be unmuted manually.

This behavior can be observed in the newly added `Player/Embeds` stories. Please note that video.js caches whether a certain video has been muted or not, so the default player may continue to be muted if it was muted manually before a reload. The `initiallyMuted` query parameter overrides this cache.

As a fan of the boy scout rule, I also improved the typings of the `<VideoJS>` component. Please let me know if you would prefer this to happen in another commit or perhaps even not at all. While I was hoping to add some tests for the new behavior, I was only able to find existing cypress tests and I didn't fully understand how the embed tests work so I decided against it for the moment. If you would like me to add tests, I'd appreciate a pointer what type of tests to add, where to add them and where I can find examples and perhaps shared setup code that I could use.

Fixes #2420

### DIFF
--- a/web/components/video/OwncastPlayer/OwncastPlayer.tsx
+++ b/web/components/video/OwncastPlayer/OwncastPlayer.tsx
@@ -24,6 +24,7 @@ let latencyCompensatorEnabled = false;
 export type OwncastPlayerProps = {
   source: string;
   online: boolean;
+  initiallyMuted?: boolean;
 };
 
 async function getVideoSettings() {
@@ -38,7 +39,11 @@ async function getVideoSettings() {
   return qualities;
 }
 
-export const OwncastPlayer: FC<OwncastPlayerProps> = ({ source, online }) => {
+export const OwncastPlayer: FC<OwncastPlayerProps> = ({
+  source,
+  online,
+  initiallyMuted = false,
+}) => {
   const playerRef = React.useRef(null);
   const [videoPlaying, setVideoPlaying] = useRecoilState<boolean>(isVideoPlayingAtom);
   const clockSkew = useRecoilValue<Number>(clockSkewAtom);
@@ -215,6 +220,7 @@ export const OwncastPlayer: FC<OwncastPlayerProps> = ({ source, online }) => {
     playsinline: true,
     liveui: true,
     preload: 'auto',
+    muted: initiallyMuted,
     controlBar: {
       progressControl: {
         seekBar: false,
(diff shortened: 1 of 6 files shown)


gabek:
This is fantastic, thank you! Works as expected and I appreciate the added story and typings, those are great additions.

As for tests we really only have the cypress tests when it comes to the web side of things, and they don't cover everything, especially when it comes to the embeds. Certainly up for any improvements on that front, however!

Thanks so much for your contribution, I really appreciate your additions and improvements!

LBBO:
You're very welcome, I'm glad you like the changes! I'll keep the tests in my mind while working on other issues and might come back to you for a discussion on how tests could be added.

While browsing the code, I noticed that wrapper types (e.g. `Number` instead of `number`) were used in several places. Is there a reason for that? If not, I could open up an issue to fix it. Also, if there's something particular you think I could work on next, let me know (I have several years of web dev experience but no knowledge of go yet). If not, I'll just browse the issues for something interesting :D

Finally, have an early happy new year!  🎉

gabek:
I'm sure I randomly used both number types interchangeably and only put thought into it when some kind of type checking told me I was using the wrong one. So nope, no reason at all :)

Thank you for offering to continue to help! There are a few things still in this next [v0.1.0](https://github.com/owncast/owncast/milestone/18) milestone that I'm not sure if you'd be interested in, but have been hanging around.

One being this annoying issue of #2342 where in development mode (`npm run dev`) things work fine, but in compiled/built/production mode (`./build/web/bundleWeb.sh`) the chat "Go to last message" shows up more than it should. And I can't seem to debug what's the difference between these two scenarios, or why the same code operates differently in these two cases.

And #2500 is similar, and I'm sure they're connected. Where in this one when a new chat message comes in it doesn't fully scroll to the bottom.

If you have any Javascript troubleshooting skills that can be helpful on either or any of these issues, they'd be a huge help!

A big happy new year to you as well!

MFTabriz:
@LBBO normal embed is now generating urls as `/embed/video?undefined"`. Can you have a quick look at it?

LBBO:
@MFTabriz Thanks for pointing that out! PR will be there in a sec.

PR #2548, opened 2023-01-04
Move enable save button logic to a use effect hook - 2546

### LINKED ISSUE #2546: Restrict twitter notification save if the required fields are empty.

When twitter notification is enabled, we can save even if the required fields are empty. This further causes an issue where the save button disappears when the toggle is changed. 

Below is a recording of the replication:

https://user-images.githubusercontent.com/44574494/210255266-d0b7b6f1-0246-4bfc-b987-252cd7fdf9da.mov

hsingh124:
Closes #2546

This moves the enable save button logic for twitter notifications into a use effect hook. This updates the state without delay and therefore fixes the issue.

Below is a recording of the fix.

https://user-images.githubusercontent.com/44574494/210586529-569d3d15-6efd-4016-a654-d072dacca290.mov

### DIFF
--- a/web/components/config/notification/twitter.tsx
+++ b/web/components/config/notification/twitter.tsx
@@ -66,14 +66,16 @@ export const ConfigNotify = () => {
     );
   };
 
+  useEffect(() => {
+    setEnableSaveButton(canSave());
+  }, [formDataValues]);
+
   // update individual values in state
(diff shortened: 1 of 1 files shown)


gabek:
Makes complete sense to me. Thank you for making this change!

PR #2525, opened 2022-12-29
improve stream testing script

MFTabriz:
- merge `test/testContent.sh` into `test/ocTestStream.sh`
- detect ffmpeg (in PATH, current dir, parent dir), fail if it's not found
- log ffmpeg executable path and version
- add `--help`

Also:
- use `test/ocTestStream.sh` for automated api tests
- install font in alpine container used for api testing (for `ocTestStream.sh` overlay text), remove unused ffmpeg package
- remove unused `test/automated/test.mp4`
- remove bitrate check in automated tests (H264 technically does not support CBR)
- fix ffmpeg installation in automated test runner scripts
- remove double spaces from transcoder error logs

Issue #2502, opened 2022-12-25
Create new development test stream option


gabek:
### Share your bug report, feature request, or comment.

A handful of people have had problems running `ocTestStream.sh`, and given it's mentioned in some of the development documentation it probably creates a bad first impression when trying to contribute to Owncast.

This script isn't required to work on Owncast, but I personally find it to be the fastest way to get a test stream running. However when it doesn't work then people are lost.

We need a more robust test stream option that probably combines `ocTestStream.sh` and [`testContent.sh`](https://github.com/owncast/owncast/blob/webv2/test/testContent.sh) to allow people to quickly get a stream up and running.

My initial thought is if you provide the path to a video file it streams it (like ocTestStream does) and if you don't it'll generate test content (like testContent does).

MFTabriz:
@gabek what do you think about using the [test video](https://github.com/owncast/owncast/blob/webv2/test/automated/test.mp4) we already have, if no file is specified?

gabek:
I was actually thinking about removing that test video file since `testContent.sh` can replace it. That test video is purposefully super overly compressed because it was there for automated tests that people wouldn't be visually looking at, I just needed something that was technically a video and something super small that wouldn't increase the size of our repository or hit any kind of max size limits. I kind of fear if people were to stream it and look at it they'd think something was wrong.

But I think we could have a default test video _and_ not rely on something that has to be kept in the repo.

We could have a test video that's not excessively compressed and put it up on a web server and the test script could download that video if required. Then file size wouldn't matter. It could also download ffmpeg if needed, since that seems to be a sticking point for people trying to run the test script.

aaronpk:
One thing that occurs to me is that a lot of development of Owncast has nothing to do with the actual video pipeline, and is just about the UI around the live video, like how the chat only appears when a video is live. What about adding a development flag that simulates a live video without actually using the video streaming part of owncast? Basically something that could toggle the video player playing a short static clip from the web server so that the streaming UI loads. But not actually using any of the transcoding pipeline. Just a thought, feel free to ignore.

gabek:
It's a valid suggestion for sure. Unfortunately there's a lot of things that derive state from the existence of the inbound stream itself. It would requite a lot of `if inTesting` logic to fake state in the backend and changing up the frontend player  to know about test state so it can update its config to point at something else.

[...]

PR #2549, opened 2023-01-04
improve stream testing script

MFTabriz:
patch #2525


test/automated/api/run.sh
```diff
@@ -6,14 +6,13 @@ TEMP_DB=$(mktemp)
...
 # Download a specific version of ffmpeg
-if [ ! -d "ffmpeg" ]; then
-	mkdir ffmpeg
-	pushd ffmpeg >/dev/null || exit
-	curl -sL https://github.com/vot/ffbinaries-prebuilt/releases/download/v4.2.1/ffmpeg-4.2.1-linux-64.zip --output ffmpeg.zip >/dev/null
-	unzip -o ffmpeg.zip >/dev/null
-	PATH=$PATH:$(pwd)
-	popd >/dev/null || exit
-fi
+FFMPEG_PATH=$(mktemp -d)
```
gabek: Does this mean it's now downloading a new copy of ffmpeg to a new place every time the tests run? That's a bit excessive. I often run the tests many times and I'd hate to waste all that time and space downloading the same copy of ffmpeg over and over again.

test/automated/browser/run.sh
```diff
@@ -22,15 +22,13 @@ npm install --quiet --no-progress
...
-if [ ! -d "ffmpeg" ]; then
-	echo "Downloading ffmpeg..."
-	mkdir -p /tmp/ffmpeg
-	pushd /tmp/ffmpeg >/dev/null
-	curl -sL --fail https://github.com/vot/ffbinaries-prebuilt/releases/download/v4.2.1/ffmpeg-4.2.1-linux-64.zip --output ffmpeg.zip
-	unzip -o ffmpeg.zip >/dev/null
-	PATH=$PATH:$(pwd)
-	popd >/dev/null
-fi
+FFMPEG_PATH=$(mktemp -d)
```
gabek: Same here re: downloading ffmpeg.

test/ocTestStream.sh
```diff
@@ -1,55 +1,116 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
-# A recent version of ffmpeg is required for the loop of the provided videos
-# to repeat indefinitely.
-# Example: ./test/ocTestStream.sh ~/Downloads/*.mp4 rtmp://localhost/live/abc123
+# Requirements:
+#   ffmpeg (a recent version with loop video support)
```
gabek, line 4: I didn't catch this before, but I thought one of the primary issues we were trying to resolve was people not having ffmpeg on their machine and not understanding how to install it. Can the script download ffmpeg (if not found) just like the other test scripts do? (Assuming it's run on Linux I guess).

MFTabriz: Our automated test scripts are run in a well-known environments (mostly Linux containers). If we want this script to be used by users who don’t know how to install a package on their own machine, we need to write a fool-proof solution which works everywhere. I don’t know a ready-made solution  for it. Also consider that this script runs on windows too (with cygwin).
What about pointing people to ffmpeg website with a note on how to find the right file instead?

gabek: Isn't a fool-proof solution to help people work on Owncast what the original goal of https://github.com/owncast/owncast/issues/2502 was?

test/automated/api/run.sh
MFTabriz: I moved the ffmpeg to a sub-directory. Now, it is only downloaded once.

test/automated/browser/run.sh
MFTabriz: fixed here too!

test/ocTestStream.sh
MFTabriz: In that case, I guess this PR is not addressing the issue! 😅 I’ll leave installing ffmpeg to people who are experienced in multi-platform scripting.

PR #2572, opened 2023-01-11
Fix HLS automatic test

### LINKED ISSUE #2571: New ocTestStream.sh script does not end stream when asked to from script causing HLS test to fail

When the HLS test script tells ocTestStream.sh to exit it does not, causing the test to fail.

https://github.com/owncast/owncast/blob/webv2/test/automated/hls/run.sh#L60

MFTabriz:
- properly cleanup the test script sub-processes
- check video file before streaming in `test/ocTestStream.sh`

closes  #2571


gabek:
Thank you for fixing this up!

Issue #2540, opened 2022-12-31
Add a share button to the video player


LBBO:
### Share your bug report, feature request, or comment.

While implementing #2539, I had the idea to add a share button to the video player and possibly the admin page. Such a button would make it easier to share a stream to different media. In contrast to the already existing Notifications, this would enable manual control over sharing streams and would enable viewers to share a stream more easily. Also, this could be a good opportunity to add something like an embed configurator where users can have an `iframe` generated for them just by clicking some buttons.

Just as an example, here are some screenshots of the YouTube share button (sorry for the German UI)

![image](https://user-images.githubusercontent.com/8076094/210137864-ac059a88-81e9-485b-b5ee-dcbf406eb524.png)
![image](https://user-images.githubusercontent.com/8076094/210137902-bcce47a9-a529-4a19-be02-d1d9c518fc60.png)

If you think this is a good idea, I'd be interested in implementing it! However, I would probably still need some info on how you would want it to behave (Should it be a popup, a modal or something else? If so, are there any existing components I should use? How would you like it to be styled? What means of sharing should be supported and how should they work?)

gabek:
I think it's an interesting idea!

The biggest question of a share modal would be what services to include, and I think that's a question that would be difficult to answer.

While some would say Facebook and Twitter, others would find that pretty awful to include. I know I wouldn't feel comfortable shipping with "Share to Facebook" and Twitter icons built in front and center. That's kind of why the compromise exists for the customization of the social network handles, so you set the ones you use.

Building something similar to that for a share modal would be far more difficult since you can't just drop a link in, every service has its own flow you need to tie into.

So I'm happy to talk about how this might work, but my expectations of ever finding a middle ground where it could work are low :)

004EK004:
Hey, I created this cool share button. Feel free to edit the files to achieve the desired outcome.[https://github.com/004EK004/GitHub_OpenSource_Project_1](url)

gabek:
Hi there @004EK004. I looked at your repo and I couldn't see how this could work with Owncast. But please file a PR with a change if you're interested in contributing.

PR #2590, opened 2023-01-16
Fix empty query in video embed story

LBBO:
Fixes @MFTabriz's comment in #2539 about the unmuted embed player being loaded with a `?undefined` query.

### DIFF
--- a/web/stories/VideoEmbed.stories.tsx
+++ b/web/stories/VideoEmbed.stories.tsx
@@ -12,17 +12,20 @@ const Template = ({
   title: string;
   width: number;
   height: number;
-}) => (
-  <iframe
-    src={`${origin}/embed/video?${query}`}
-    title={title}
-    height={`${height}px`}
-    width={`${width}px`}
-    referrerPolicy="origin"
-    scrolling="no"
-    allowFullScreen
-  />
-);
+}) => {
+  const queryString = query ? `?${query}` : '';
+  return (
+    <iframe
+      src={`${origin}/embed/video${queryString}`}
+      title={title}
+      height={`${height}px`}
+      width={`${width}px`}
+      referrerPolicy="origin"
+      scrolling="no"
+      allowFullScreen
+    />
+  );
+};
 
 const origins = {
   DemoServer: `https://watch.owncast.online`,


gabek:
I haven't put any thought into the video embed story until now, but they don't make any sense and should probably just be deleted until something better can be written.

Right now the story is pointing at random Owncast servers in an iFrame, not at all testing out the current version of the code because these servers aren't running the current version of the code.

So I wouldn't worry about working the story for video embed stories because they're incorrect and unhelpful right now.

LBBO:
I see your point, but if the server is set to localhost it's testing the implementation, isn't it? So maybe that selection box should be changed / removed, but I think the story still provides value

gabek:
That's only if you're running storybook locally _and_ you're running Owncast locally. I'm more concerned about instances like at https://owncast.online/components/?path=%2Fstory%2Fowncast-player-embeds--default&args=origin%3Alocalhost where selecting `localhost` won't do anything and the other options are incorrect.

LBBO:
Ahh, I see. Maybe there could be a smarter default that depends on storybook's origin? That could set a more useful default for localhost and "the official" online version

web/stories/VideoEmbed.stories.tsx
```diff
@@ -12,17 +12,20 @@ const Template = ({
...
-    referrerPolicy="origin"
-    scrolling="no"
-    allowFullScreen
-  />
-);
+}) => {
+  const queryString = query ? `?${query}` : '';
+  return (
+    <iframe
+      src={`${origin}/embed/video${queryString}`}
```
gabek, line 19: The problem here is the origin isn't directly connected with the state of the Storybook story. Because an instance of Storybook isn't running Owncast it can't point to itself, so I'm unfortunately not sure how to make a story for the embed that's not more than a blank iframe.

gabek:
As a heads up I did go ahead and remove the video embed storybook story mostly because at this point it was being marked as "changed" each time it ran (because it was rendering actual video). It wasn't actually testing a component (just an iframe) and throwing incorrect test results I just removed it until we can find a better option. Maybe the next step would be to have a story just for _offline_ video embed, since there's an actual page component that could be mounted and not just an iframe pointed to a remote server.

LBBO:
Alright, good to know. I guess that removes the need for this PR then, right?

I kind of disagree that it was "just testing an iframe", though. Technically, you're right of course, but it was showcasing how different URL parameters affect the iframe. Perhaps that would be redundant with #2540 though? Anyhow, I wasn't aware of the issues with testing. Sorry about that.

gabek:
I wasn't aware of it until later either, so no big deal :)

gabek:
As for the share button, I still think that's a good idea in theory, but in execution I still don't know specifically what to populate a share modal with. It could be just "share as iframe" and maybe that's enough, but most likely people are looking for something like "Share to <specific service>". That's why that feature hasn't gone anywhere.

gabek:
Closing this PR as the video embed story was removed.

PR #2592, opened 2023-01-16
Add noscript for browsers without JS

### LINKED ISSUE #2543: Add content to the landing page for browsers without JS support
Owncast page in `webv2` branch with browsers without JS support currently looks like [this](https://user-images.githubusercontent.com/39205857/210184282-18e57f63-17a4-41a1-bafd-ee7215495744.png). Its content should get updated based on the [main branch](https://user-images.githubusercontent.com/39205857/210184415-6728935b-a3cc-4a6c-a29f-a710ffff2a7f.png) ([source](https://github.com/owncast/owncast/blob/a5e5037f5c09e9deedad6558ea9a7ea8a0e486de/webroot/index.html#L220-L259)).

LBBO:
Fixes #2543 

Please note that the [styles aren't applied in a dev build](https://nextjs.org/docs/basic-features/built-in-css-support#does-it-work-with-javascript-disabled). In order to test the changes correctly, please run `npm run build && npm start` and then open the page in a browser without JS enabled. In chrome, you can open the dev tools (F12), open the command palette (Cmd / Ctrl + Shift + P) and search for "JavaScript". You should find a command to enable / disable JS. Then reload the page and voilà!

I'm currently still having an issue with the logo just disappearing (like `visibility: hidden;`, not `display: none`) as soon as the screen width is `< 768px`. Does anyone have an idea why that is? I couldn't find any changes in its or its parents' CSS between `768px` and `767px`. It seems likely that this behavior is related to the desktop breakpoint (`>= 768px` in https://github.com/owncast/owncast/blob/029dba0808c6a2dc324371aa81dbf838a977142f/web/components/ui/Content/Content.tsx#L306 or https://github.com/owncast/owncast/blob/029dba0808c6a2dc324371aa81dbf838a977142f/web/styles/mixins.scss#L9) but I couldn't find a way the SCSS mixin could affect the `noscript` or its descendants. I think it's safe to rule out the React state variable as a possible cause because the issue occurs when the screen is resized while JS is disabled.


gabek:
If you're looking to have something rendered without using Javascript I would guess the best bet would be to put plain HTML into `_document` and don't use CSS modules as there's no guarantee React components or CSS modules will get rendered down to plain text and not require Javascript. `_document` is going to be the highest level and most likely going to not turn into a Javascript bundle that needs executing.

LBBO:
Doesn't [this section of the CSS support documentation](https://nextjs.org/docs/basic-features/built-in-css-support#does-it-work-with-javascript-disabled) promise that CSS will work in production with JS disabled? And since everything works right now thanks to SSR (at least it works in Chrome with JS disabled), why exactly would `_document` be preferable over normal JSX?

gabek:
Only because I don't trust it'll _always_ be that way. If the goal is to make something work without Javascript using Javascript it feels like that can easily fall apart. I don't trust Next, it's a black box and it'll do what it's going to do. So I figured the document is most likely to be rendered at build time. Every nested component I feel less confident that it would always be the case. I don't even feel confident about document being rendered into html, to be honest, but it's as close as we'd get.

LBBO:
Alright, I see. The thing is that this solution is way more readable and maintainable, in my opinion. How about I check if it's possible to write a snapshot test that makes an actual HTTP call, ensures there is one single `<noscript>` tag and then checks its contents against a snapshot? Something similar could happen for the CSS. That way, we would have a backup copy of the actual HTML (and CSS) in case the rendering changes and, more importantly, we would be notified if something causes the SSR to break. But until then, we would have more readable code.

gabek:
I think if it's working right now the way that you're doing it, since the noscript case is going to be a pretty niche case, so I don't think too much effort needs to go into verifying it or anything. I don't hold my concerns about Nextjs's rendering decisions strongly or anything. If something changes in the future it's not like this will break everything for everyone, it'll only break things for people who are already broken.

LBBO:
Alright, great! Do you have any idea what the issue with the image and the viewport might be? If not, I'll just have to do some more digging next week 🤷‍♂️

gabek:
I don't have any answers offhand, but since this will impact very few people I think if you're running into issues styling this noscript block then maybe just make it super simple so you don't have to deal with styling as much.

gabek:
Or no styling at all for that matter is fine since it's simply a message and not a part of functionality in the application. As long as it's readable so they know to either turn on Javascript or go somewhere else :)

gabek:
Heads up: I changed the base of this PR from `webv2` to `develop` as all the feature work has been merged into `develop` and that's where changes will take place now.

LBBO:
Alright, after spending way too much time trying to figure out what the hell was going on with the logo (it just bugs me if I can't figure something like that out), I decided to just hide it on smaller screens. I also found and fixed another issue with vertical overflowing. Sorry for the delay, I hope this works for you now. Let me know if there's anything you'd like me to change!

gabek:
Great! Thanks so much for continuing to dig into it. I had totally forgotten about a `noscript` block, so it's good you caught it!

PR #2619, opened 2023-01-24
Add the Client-side Input Validators for Stream Keys and the Admin Password

### LINKED ISSUE #2528: Add minimum stream key + password complexity rules

The server functions that adds new stream keys and modify the admin panel password are in file controllers/admin/config.go: SetAdminPassword at line 201 and SetStreamKeys at line 772.

SetAdminPassword

```go
// SetAdminPassword will handle the web config request to set the server admin password.
func SetAdminPassword(w http.ResponseWriter, r *http.Request) {
	if !requirePOST(w, r) {
		return
	}

	configValue, success := getValueFromRequest(w, r)
	if !success {
		return
	}

	if err := data.SetAdminPassword(configValue.Value.(string)); err != nil {
		controllers.WriteSimpleResponse(w, false, err.Error())
		return
	}

	controllers.WriteSimpleResponse(w, true, "changed")
}
```

SetStreamKeys

```go
// SetStreamKeys will set the valid stream keys.
func SetStreamKeys(w http.ResponseWriter, r *http.Request) {
	if !requirePOST(w, r) {
		return
	}

	type streamKeysRequest struct {
		Value []models.StreamKey `json:"value"`
	}

	decoder := json.NewDecoder(r.Body)
	var streamKeys streamKeysRequest
	if err := decoder.Decode(&streamKeys); err != nil {
		controllers.WriteSimpleResponse(w, false, "unable to update stream keys with provided values")
		return
	}

	if err := data.SetStreamKeys(streamKeys.Value); err != nil {
		controllers.WriteSimpleResponse(w, false, err.Error())
		return
	}

	controllers.WriteSimpleResponse(w, true, "changed")
}
```

These two functions do not enforce any password complexity and will update the current value with any value submitted by the admin.
If no requirements are set it could be likely that weak password are used by the administrator, and this could create an opportunity for a guessing or brute force attack against the admin panel basic HTTP authentication.

## Impact

An attacker could guess the admin panel password or a stream key using a list of common password, obtaining access to the application controls.

This scenario requires that the legitimate admin of the application set a weak password or did not change the default one.

## Recommendation

Enforce the following requirements on the credentials, when they can be applied:

1. Enforcement of a minimum and maximum length
1. Complex passwords requiring mixed character sets (alpha, numeric, special, mixed case)

Add a warning in the admin UI when the admin panel password and stream key are the default one (abc123), suggesting to change them explaining the security risk (e.g. anyone could log in the admin panel and takeover the stream).

dorj222:
Currently, there are no client-side input validators for Stream Keys and the Admin Password. Setting a weak stream key or admin password could result in a scenario where the attacker could obtain access to the application controls. This PR implements the client-side input validators for stream keys and the admin password. These new input validators enforce the user to submit an input (a stream key or admin password) that must be a minimum of 8 characters and a maximum of 192 characters and must include at least 1 uppercase letter, at least one lowercase letter, at least 1 special character, and at least 1 number. [The Ant Design Form Validations](https://ant.design/components/form) have been used to implement these input validators. This PR fixes the issue #2528

1. The Input Validator for the Stream Keys:

Screen Shot 2023-01-24 at 3 04 53 PM

2. The Input Validator for the Admin Password:

Screen Shot 2023-01-24 at 3 04 15 PM


gabek:
Awesome, I'll check it out!

web/components/admin/config/server/StreamKeys.tsx
```diff
@@ -49,16 +53,66 @@ const AddKeyForm = ({ setShowAddKeyForm, setFieldInConfigState, streamKeys, setE
...
+            pattern: /\d/,
+            message: '- at least one digit',
+          },
+          {
+            pattern: /^(?=.*?[#?!@$%^&*-])/,
+            message: '- at least one special character: !@#$%^&*',
+          },
+        ]}
+      >
+        <Input placeholder="def456" onChange={handleInputChange} />
```
gabek: We probably shouldn't set a simple placeholder if that's not an actual valid key.

web/components/admin/TextField.tsx
```diff
@@ -53,14 +54,30 @@ export const TextField: FC<TextFieldProps> = ({
...
   value,
 }) => {
+  const [hasChanged, setHasChanged] = useState(false);
+  const regex = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[!@#$%^&*]).{8,192}$/;
+  const [form] = Form.useForm();
   const handleChange = (e: any) => {
     // if an extra onChange handler was sent in as a prop, let's run that too.
     if (onChange) {
       const val = type === TEXTFIELD_TYPE_NUMBER ? e : e.target.value;
+      // console.log('val: ', val, 'fieldname: ', fieldName);
```
gabek: Nit: Console

web/components/admin/TextField.tsx
```diff
@@ -53,14 +54,30 @@ export const TextField: FC<TextFieldProps> = ({
...
+      } else {
+        setHasChanged(false);
+      }
+
       onChange({ fieldName, value: useTrim ? val.trim() : val });
     }
   };
 
+  useEffect(() => {
+    // console.log('value: ', value);
```
gabek: Nit: Console

web/components/admin/TextFieldWithSubmit.tsx
```diff
@@ -134,15 +135,17 @@ export const TextFieldWithSubmit: FC<TextFieldWithSubmitProps> = ({
...
-            <Button
-              type="primary"
-              size="small"
-              className="submit-button"
-              onClick={handleSubmit}
-              disabled={!hasChanged}
-            >
-              Update
-            </Button>
+            {fieldName !== 'adminPassword' && (
```
gabek: This feels like it shouldn't have a specific use case hard-coded in it.  Maybe if you're conditionally adding a button it can have a new property added to it that's like `hasSubmitButton` or something that can default to `true` but be set to `false` in the adminPassword case.

web/components/admin/TextField.tsx
```diff
@@ -128,25 +145,87 @@ export const TextField: FC<TextFieldProps> = ({
...
-            {...(type !== TEXTFIELD_TYPE_NUMBER && { allowClear: true })}
-            placeholder={placeholder}
-            maxLength={maxLength}
-            onChange={handleChange}
-            onBlur={handleBlur}
-            onPressEnter={handlePressEnter}
-            disabled={disabled}
-            value={value as number | (readonly string[] & number)}
-          />
+      {fieldName !== 'adminPassword' ? (
```
gabek: Same comment about the hardcoded `adminPassword` check here, maybe a property can be added for the two different states.

gabek:
Heads up: I changed the base of this PR from `webv2` to `develop` as all the feature work has been merged into `develop` and that's where changes will take place now.

web/components/admin/TextFieldWithSubmit.tsx
dorj222: I created a new `useState` isPwdInput boolean variable for this button and added a new `useEffect` function to check the `fieldName` and if it is the password input field would be set to true, else false. Please let me know if this change would resolve this code change request.

web/components/admin/TextField.tsx
dorj222: To fix this issue, I tried to create the `useState` boolean variable and `useEffect` function to listen to the variable `type` change for the input fields. However, when they were added the Regex Input Validators stopped working, and I could not find another way to resolve this issue. Currently, `type !== TEXTFIELD_TYPE_PASSWORD ` is being used to check if it is the password input field or not. Please feel free to suggest other possible solutions.

gabek:
I know you're still working on this, so just ping me when you're ready for me to look at it again. Thanks so much for helping with this!

dorj222:
Hi there @gabek! Would you please take a look at it again? :)

dorj222:
Actually, it turns out that changing the admin password endpoint API is causing some issues related to `const { enabled: federationEnabled } = federation;`. So, this error shows up whenever the admin password is successfully changed through the frontend admin. I will investigate this issue, and please feel free to suggest solutions that may fix this issue. 

Here is the error message: 
Screen Shot 2023-02-05 at 4 40 43 PM

gabek:
I just checked out your PR wasn't able to reproduce the error you mentioned above. Could you walk me through how to reproduce it?

dorj222:
Sure! I changed the admin password through the frontend web application, refreshed the page after the password has been changed successfully, and then received the error that I showed. This error does not occur after adding a new stream key and only happens after the admin password change.

gabek:
What's going on is in development mode the password is always hard coded to `abc123`.  https://github.com/owncast/owncast/blob/develop/web/.env.development

In a production build it's fine. You can build production by running `./build/web/bundleWeb.sh` and accessing http://localhost:8080.

I know in this particular case it's super annoying, hopefully at some point in the future it can be improved on. Usually for development a hard coded password makes things faster to get things done, but in this case it's getting in the way.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast">owncast/.env.development at develop · owncast/owncast</a></strong></div><div>Take control over your live stream video by running it yourself.  Streaming + chat out of the box. - owncast/.env.development at develop · owncast/owncast</div></blockquote>
<blockquote><div><strong><a href="http://localhost:8080/">unfurl | built with Probot</a></strong></div></blockquote>

dorj222:
ah, that really makes sense! thanks a lot for your prompt response! I think this PR is in a good shape. Would you please review it again?

web/components/admin/TextField.tsx
```diff
@@ -53,14 +54,29 @@ export const TextField: FC<TextFieldProps> = ({
...
+  const [hasPwdChanged, setHasPwdChanged] = useState(false);
+  const [showPwdButton, setShowPwdButton] = useState(false);
+  const regex = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[!@#$%^&*]).{8,192}$/;
+  const [form] = Form.useForm();
   const handleChange = (e: any) => {
     // if an extra onChange handler was sent in as a prop, let's run that too.
     if (onChange) {
       const val = type === TEXTFIELD_TYPE_NUMBER ? e : e.target.value;
+      setShowPwdButton(true);
+      if (fieldName === 'adminPassword' && regex.test(val)) {
```
gabek: Looks like there's still hardcoded references to an implementation of the text fields in the component itself with `adminPassword`. The generic components shouldn't have to know about specific instances of them.

web/components/admin/TextField.tsx
```diff
@@ -53,14 +54,29 @@ export const TextField: FC<TextFieldProps> = ({
...
+      } else {
+        setHasPwdChanged(false);
+      }
+
       onChange({ fieldName, value: useTrim ? val.trim() : val });
     }
   };
 
+  useEffect(() => {
+    form.setFieldsValue({ adminPassword: value });
```
gabek: Another reference to adminPassword.

web/components/admin/TextField.tsx
```diff
@@ -128,25 +144,88 @@ export const TextField: FC<TextFieldProps> = ({
...
+          <FormStatusIndicator status={status} />
+          <p className="field-tip">{tip}</p>
+        </div>
+      ) : (
+        <div className="input-side">
+          <div className="input-group">
+            <Form
+              name="basic"
+              form={form}
+              initialValues={{ adminPassword: value }}
```
gabek: Another reference to adminPassword.

web/components/admin/TextField.tsx
```diff
@@ -128,25 +144,88 @@ export const TextField: FC<TextFieldProps> = ({
...
+              style={{ width: '100%' }}
+            >
+              <Form.Item
+                name="adminPassword"
```
gabek: Another reference to adminPassword.

web/components/admin/TextFieldWithSubmit.tsx
```diff
@@ -66,6 +67,13 @@ export const TextFieldWithSubmit: FC<TextFieldWithSubmitProps> = ({
     }
   }, [value]);
 
+  useEffect(() => {
+    if (fieldName === 'adminPassword') {
```
gabek: Another reference to adminPassword.

gabek:
Instead of hardcoding a concrete implementation of the component, could a boolean property be added to perform the logic that you're now doing by the string comparison? I know you mentioned above you were having issues removing that string comparison, but a generic component can't have internal logic based on one single use of it.

dorj222:
Hi @gabek! Thank you so much for your feedback and suggestions. As you suggested, I added boolean variables inside the `TEXTFIELD_PROPS_ADMIN_PASSWORD`, `TEXTFIELD_PROPS_FFMPEG`, `TEXTFIELD_PROPS_WEB_PORT`, and `TEXTFIELD_PROPS_RTMP_PORT`, and these boolean variables are used to perform the logic. I rid of all the `fieldName === 'adminPassword'` comparisons. Also, I moved Regex constants into the `config-contstants` in order to make the code more readable and consistent. If it is possible, would please review it one more time again?

gabek:
Absolutely, I'm on it!

web/components/admin/TextField.tsx
```diff
@@ -128,25 +150,66 @@ export const TextField: FC<TextFieldProps> = ({
...
+                  className={`field ${className} ${fieldId}`}
+                  onChange={handleChange}
+                  onBlur={handleBlur}
+                  placeholder={placeholder}
+                  onPressEnter={handlePressEnter}
+                  disabled={disabled}
+                  value={value as number | (readonly string[] & number)}
+                />
+              </Form.Item>
+              {showPwdButton && (
```
gabek, line 194: I see you added a submit button to `TextField`, but there's already a `TextFieldWithSubmit` that has a built-in submit button. It seems like these are now somewhat the same thing with them both showing a submit button on change. Sorry if I'm a little confused about needing to add a button to the one without a button.

web/components/admin/TextField.tsx
```diff
@@ -52,15 +55,30 @@ export const TextField: FC<TextFieldProps> = ({
   type,
   useTrim,
   value,
+  isAdminPwdField,
 }) => {
+  const [hasPwdChanged, setHasPwdChanged] = useState(false);
```
gabek, line 60: Since `TextField` is used for all kinds of text input, not just passwords, could we use generic naming of things? Maybe `hasChanged` or `hasContentChanged` or something so it doesn't confuse somebody who's looking at the code later and wondering what a password has to do with it.

web/components/admin/TextField.tsx
```diff
@@ -128,25 +150,66 @@ export const TextField: FC<TextFieldProps> = ({
...
-            {...(type !== TEXTFIELD_TYPE_NUMBER && { allowClear: true })}
-            placeholder={placeholder}
-            maxLength={maxLength}
-            onChange={handleChange}
-            onBlur={handleBlur}
-            onPressEnter={handlePressEnter}
-            disabled={disabled}
-            value={value as number | (readonly string[] & number)}
-          />
+      {!isAdminPwdField ? (
```
gabek: Similar comment, since these text fields aren't password specific. If what you're splitting up are fields with complexity requirements and fields without complexity requirements maybe the bool should be called `hasComplexityRequirements`.

gabek:
Sorry for having more questions and a few thoughts. Thank you for making those other changes, I think it's on the right path now with not putting implementation specific references in the generic component that's used everywhere in the admin. I just had a couple nits about calling things "password" since it's a generic component and can be used for any kind of input, and a question about the duplication of submit buttons between the two input components.

dorj222:
Hi @gabek! I will work on these code change requests soon and will you know when it's ready. Thanks a lot for your feedback!

gabek:
Sounds great, thanks for the update!

web/components/admin/TextField.tsx
dorj222: Hi, @gabek! The main reason why I add the separate submit button for the Admin Password field is that if we use the generic submit button of `TextFieldWithSubmit`, then the user would be able to submit the input even if their input is not satisfying the password complexity rules. Here in this post, I attached the screenshot example for this case. The separate submit button of `TextField` essentially prevents that problem and enforces the user to type a password that satisfies the password complexity rules. 

The screenshot example:
Screen Shot 2023-02-21 at 3 22 51 PM

web/components/admin/TextField.tsx
dorj222: I named them `hasPwdChanged` and `setHasPwdChanged` because they are only for the Admin Password input field and are password specific. However, I can change them to more generic names like `hasChanged` or `hasContentChanged`.  This `useState` makes the submit button for the Admin Password input field visible from the hidden state whenever the user types something inside the Admin Password input field. Also, we can make the submit button for the Admin Password always visible and remove this specific `useState`.

gabek:
Hey @dorj222, I think we're good to get this in once the merge conflict is resolved.

dorj222:
Hi @gabek! The merge conflict has been resolved. Would you please review it again? 🙂

gabek:
Awesome, will do!

gabek:
Thanks so much for continuing to work on this, let's get it in! :1st_place_medal:

gabek:
Bad news, this PR is breaking the build and it needed to be reverted. I don't see build errors in the test above, but it's broken for me and others.

```
ReferenceError: Cannot access 'TEXTFIELD_TYPE_URL' before initialization
    at Object.xA (/home/gabek/src/webv2-merge/owncast/web/.next/server/chunks/6960.js:57:51)
    at 745 (/home/gabek/src/webv2-merge/owncast/web/.next/server/chunks/6960.js:1379:94)
    at __webpack_require__ (/home/gabek/src/webv2-merge/owncast/web/.next/server/webpack-runtime.js:25:42)
    at 9139 (/home/gabek/src/webv2-merge/owncast/web/.next/server/chunks/6960.js:68:81)
    at __webpack_require__ (/home/gabek/src/webv2-merge/owncast/web/.next/server/webpack-runtime.js:25:42)
    at 7990 (/home/gabek/src/webv2-merge/owncast/web/.next/server/pages/admin/config-federation.js:21:85)
    at __webpack_require__ (/home/gabek/src/webv2-merge/owncast/web/.next/server/webpack-runtime.js:25:42)
    at __webpack_exec__ (/home/gabek/src/webv2-merge/owncast/web/.next/server/pages/admin/config-federation.js:525:39)
    at /home/gabek/src/webv2-merge/owncast/web/.next/server/pages/admin/config-federation.js:526:100
    at __webpack_require__.X (/home/gabek/src/webv2-merge/owncast/web/.next/server/webpack-runtime.js:177:21)

> Build error occurred
Error: Failed to collect page data for /admin/config-federation
    at /home/gabek/src/webv2-merge/owncast/web/node_modules/next/dist/build/utils.js:991:15
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  type: 'Error'
}
```

gabek:
I went ahead and fixed the circular dependency import and the build is now successful. It's merged back in.

PR #2655, opened 2023-01-31
[CONTRIB] Example varnish and hitch config

4censord:
Varnish is a http caching server, hitch is a TLS terminating proxy.

Using this config with 2 varnish instances, I occasionally run internal events with 300+ users.

I intend to write a blogpost about this someday, but I haven't found the time yet.


gabek:
Very cool! I'm surprised I don't hear more people experimenting with putting Varnish in front of Owncast. It's good to hear that it all works together well! Thanks for the example configs.

PR #2657, opened 2023-01-31
Fix/issue 2649/status bar time info

### LINKED ISSUE #2649: Status bar says 1 minutes (plural) instead of minute

Minute should be singular when it's only 1 minute.

image

elonyavist:
#### Reference issues/PRs
Status bar time info: #2649 

#### What does this implementation/fix?
Using a helper function (already present in the codebase), I managed the singular values for the time units.
With this commit [fix: manage singular day visibility](https://github.com/owncast/owncast/commit/564bcaa1411dd78e0ba49c3cd9057f11255614fa), I modified the condition to `>=` otherwise, if the `lastConnectTime` value has only 1 day before _now_, `1 day` was not shown.

Screenshot 2023-01-31 at 21 24 26

#### Any other comments?
I tested the implementation on Storybook and seems to work fine. Let me know if I made some mistakes, thx!


gabek:
Awesome, thanks so much for taking care of this! :100:

Issue #2579, opened 2023-01-13
Add automation around link checking


gabek:
### Share your bug report, feature request, or comment.

I've run into a couple incorrect/dead links, particularly in the admin. We should be able to automate checking of links on all our pages and make sure they're valid. Perhaps with cypress.

PR #2660, opened 2023-02-01
 Fix broken links

xarantolus:
Hey, I was recently clicking around the admin UI and found a few places where links were broken. That gave me the idea of running [a script](https://github.com/ubalklen/Broken-Link-Finder) over both the stream + admin UI to find other links that don't work. This PR fixes all instances I've found


gabek:
That's awesome! I actually filed an issue to find some way to automate link checking in #2579.  In the future maybe I'll build some tests that use this. Thank you for doing this!

PR #2674, opened 2023-02-04
Un-squish logo

### LINKED ISSUE #2667: Logo image is squished

As seen in this example where the moon is no longer a circle, the image should not be squished.

Story: https://owncast.online/components/?path=%2Fstory%2Fowncast-components-page-logo--not-square

image

LBBO:
Fixes #2667 by applying the correct styling for the `img` tag (`background-*` only works if the image is being rendered as a `background` or `background-image`).


gabek:
Thanks for catching that!

Issue #2668, opened 2023-02-02
Page vertical spacing layout issues


gabek:
When scrolling down the footer never stays in view when testing on an iPad using Safari.

https://user-images.githubusercontent.com/414923/216446788-45e6cfe0-ebba-4ff3-af00-63872f9941ef.MOV

gabek:
With the recent layout changes that were merged in the issues with iPad changed. Both portrait and landscape have different issues.

iPad portrait has a gap between the header and the offline message. It also has excessive vertical spacing between all the components.

![image](https://user-images.githubusercontent.com/414923/218225243-f988f255-02cd-4f8b-8483-24a4386a80f0.png)

iPad landscape cuts off the header at the top of the page.

![image](https://user-images.githubusercontent.com/414923/218225254-f90c610c-d8de-492e-ae7c-c6e794eed37f.png)

iPad landscape on chrome is a bit different

![image](https://user-images.githubusercontent.com/414923/218225680-06c3223d-2c98-417e-a4f0-9a762cc60719.png)

gabek:
I've automated some screenshots on different browsers and platforms to hopefully allow catching more of this. Right now it's only taking one set of screenshots per day, but at least it's something. https://owncast.online/components/?path=%2Fdocs%2Fowncast-screenshots-ipad-portrait-safari--online

<blockquote><div><strong><a href="https://owncast.online/components/?path=%2Fdocs%2Fowncast-screenshots-ipad-portrait-safari--online">Webpack App</a></strong></div></blockquote>

gabek:
@LBBO Sorry to bug you, but I was looking into this and the culprit for some of these spacing issues seem to be from the `placeholder` divs that were added previously. I tried removing them and it seems to help, but I'm curious if they're required for some case I'm not aware of, and if so, can they be hidden or be sizeless so they don't add these spaces?

LBBO:
Sorry, I've been a bit busy with exams and work lately.

The idea behind the placeholder divs that I added was to have a constant number of divs under the container at all times so I wouldn't have to have different layouts for when this is hidden or when that is hidden. My plan was to design the heights such that the actual components would get their desired height while the placeholders collapse into nothingness, but I guess something went wrong there.

With the CSS I wrote, removing the placeholders might throw off the grid, but I guess that could be fixed with `grid-area`s. However, I would assume that the actual issue here might be with the heights assigned to the grid rows.

[...]

PR #2676, opened 2023-02-04
Give chat a min-height that other elements yield to on mobile clients

### LINKED ISSUE #2595: Android: Page doesn't scroll when keyboard is displayed

iOS Safari does the correct thing. When you focus on the text field the page scrolls vertically.

With Android everything stays static and the keyboard covers the page, including chat and the text input.

image

This is what iOS is doing for comparison: 

image

LBBO:
WIP, mostly so others can check out the changes easily.

Fixes #2595

### DIFF
--- a/web/components/ui/Content/Content.module.scss
+++ b/web/components/ui/Content/Content.module.scss
@@ -39,6 +39,10 @@
     @include screen(desktop) {
       overflow-y: scroll;
       grid-template-rows: unset;
+
+      &.offline {
+        grid-template-rows: unset;
+      }
     }
   }
 
--- a/web/components/ui/Content/Content.module.scss
+++ b/web/components/ui/Content/Content.module.scss
@@ -15,11 +15,20 @@
 
   .mainSection {
     display: grid;
-    // TODO: make this more adaptive to the currently displayed components!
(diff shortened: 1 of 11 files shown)


gabek:
One thing I noticed is a large empty space between the content header and the tab bar:

image

LBBO:
Good catch, thanks! Should be fixed now.

gabek:
This looks and works great! It's doing everything you'd expect it do with regards to the keyboard. And thanks for adding support for passing in class names to some of those components, it's not something I've thought much about, usually I just wrap custom components in a div and give it the class, not the most correct thing to do.

I don't see any reason not to merge it, things look good and this is a big move forward for the mobile layout. I really appreciate all the work you put into it! I look forward to more testing!

LBBO:
Shoot, I should have marked the PR as a draft again since I'm still working on stories for the layout. Should I just open a new PR for them then?

gabek:
Sure, a new PR for stories sounds good!

gabek:
Just as a head's up, I split up the Mobile and Desktop layouts to new files to make it easier to manage, plus it's a first step to making those components not dependent on global state, so stories could be built around them. I did this before looking at #2668.

PR #2690, opened 2023-02-07
Random Stream Key Generator

dorj222:
The current random stream key generator doesn't always generate a random string that follows the password complexity rules we are implementing. This new random stream key generator always generates a string that follows the password complexity rules.


gabek:
Nice! Thanks for swapping that out!

PR #2693, opened 2023-02-08
Allow embedding HTML for external actions

### LINKED ISSUE #1718: Allow providing HTML instead of a link for external actions
Right now we allow people to point an external action button to any URL that we can open in a modal.  I believe this feature hasn't been fully adopted because people don't have custom HTML hosted anywhere, it's an additional step.

Instead we could allow people to paste in whatever raw HTML they want, and we can render it for them without an additional server being required.

While this wouldn't be totally useful for somebody building some new feature from scratch, if people just wanted to embed a 3rd party "embed code" copy/paste from another site it would work well here.

Examples:

- [Mailchimp signup form](https://mailchimp.com/help/add-a-signup-form-to-your-website/).
- [Paypal donation](https://www.paypal.com/us/smarthelp/article/how-do-i-add-a-paypal-payment-button-to-my-website-faq3629)

xarantolus:
This PR allows admin users to specify HTML that should be embedded into the external actions modal.

Some notes:
- It might make sense to add a **warning** to users to tell them they shouldn't paste untrusted HTML
- The URL field of ExternalActions objects is now optional, but one of URL and HTML must be set
- If HTML is a non-empty string, it is assumed that this action is now an "HTML action". `openExternal` will be ignored if that is the case (similarly when editing an action that is set to use HTML instead of an URL, the open externally checkbox is hidden, as you suggested)
- I changed the editing of actions to use a list index instead of searching by URL
- It would make sense to add documentation on this, especially [here](https://owncast.online/thirdparty/actions/). If you want me to I could add a PR on the website repo as well
- It might also make sense to add some styling for the top element in the HTML; currently it can look weird (e.g. when copying an embed code for Google Forms Surveys) because they set an explicit width that then overflows the container most of the time (embedding as URL works way better in some cases). Or we leave that to admins, as interfering with the HTML sounds frustrating for admins. 
- This works quite well with self-created signup forms that have a `target="_blank"` attribute 
- It is very easy to add style tags that interfere with owncasts design upon opening the modal. Admins need to make sure their CSS doesn't collide with owncast

Here's a screenshot of a really basic/minimal form I put into a modal, just by pasting a few lines of HTML. Setting an `action` for the form and specifying `target="blank"` will submit that form in a new tab without interrupting the stream.
Screenshot

Here's a screenshot of the editing/creation menu, the external checkbox won't be shown when editing the HTML action

| URL Action | HTML Action |
| ------- | ---------- |
| image | image |

Fixes #1718


gabek:
This is super cool! I think there will be a ton of opportunities for people to take advantage of this, thank you!

You make a lot of really good points, things I never thought of with regards to styles and sizing. Hopefully adding some documentation around some best practices can help there. If you'd like to contribute to the documentation with some of those items you mentioned that'd be great! There's a [`v0.1.0` branch](https://github.com/owncast/owncast.github.io/tree/v0.1.0) that will be merged in when this release goes live, so if you want to put up a PR it can be added to that.

I have a couple thoughts that you can let me know what you think about.

We have a component called [CodeMirror](https://codemirror.net/) that we use in the admin for Markdown, Javascript and CSS editors. Maybe we could use it here too as an HTML editor to give that text area some syntax highlighting. You can see it in use for the custom Javascript here (Noticed it says CSS on that page, whoops): https://github.com/owncast/owncast/blob/develop/web/components/admin/EditCustomJavascript.tsx#L98-L105

The second thing that came to mind is instead of a checkbox, maybe a dropdown might make sense so we could have it say something like "Action source: Link" or "Action source: HTML". The checkbox kind of makes the HTML option seem kind of like a secondary option so it'd be nice to make it clear these are two options that both serve the goal, just in different ways.

Thanks again for working on this, it works great and I think people will have fun with it!

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast">owncast/EditCustomJavascript.tsx at develop · owncast/owncast</a></strong></div><div>Take control over your live stream video by running it yourself.  Streaming + chat out of the box. - owncast/EditCustomJavascript.tsx at develop · owncast/owncast</div></blockquote>

xarantolus:
Nice, yes I'll look into implementing it with CodeMirror and will do it with a dropdown, that sounds really nice :)

xarantolus:
I really like the way CodeMirror works here :) I added the dependency for the HTML syntax highlighting package. Nested CSS and JS seem to be included in the default of the HTML language, so that already works.

Signup HTML Example

gabek:
This is awesome, thanks for making those additions! That editor is a nice touch.

Thanks so much for working on this feature, I think it's going to be a really great one for people. And thank you for tweaking the documentation to reflect this new addition. 

Let's get this in!

PR #2700, opened 2023-02-12
 This commit suggests that if there is no name then the username will…

### LINKED ISSUE #2680: Followers: If user doesn't have a name then use the account name

Not all Fediverse accounts set a friendly name, so in that case we can just use the account as the name.

image

Yasir761:
Capture
… come in place of a name

### DIFF
--- a/web/components/ui/followers/SingleFollower/SingleFollower.tsx
+++ b/web/components/ui/followers/SingleFollower/SingleFollower.tsx
@@ -19,8 +19,7 @@ export const SingleFollower: FC<SingleFollowerProps> = ({ follower }) => (
         </Col>
         <Col span={18}>
           <Row className={styles.name}>
-        {/* using ternary operrator for follower name,
-         if name is not present then username will work as a name  */}
+        
             <Typography.Text ellipsis>{follower.name || follower.username}</Typography.Text>
           </Row>
           <Row className={styles.account}>
--- a/web/components/ui/followers/SingleFollower/SingleFollower.tsx
+++ b/web/components/ui/followers/SingleFollower/SingleFollower.tsx
@@ -19,7 +19,6 @@ export const SingleFollower: FC<SingleFollowerProps> = ({ follower }) => (
         </Col>
         <Col span={18}>
           <Row className={styles.name}>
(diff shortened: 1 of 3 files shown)


gabek (commented):
Looks like you accidentally committed a change to `package-lock.json`. Could you revert that change?

web/components/ui/followers/SingleFollower/SingleFollower.tsx
```diff
@@ -19,7 +19,9 @@ export const SingleFollower: FC<SingleFollowerProps> = ({ follower }) => (
         </Col>
         <Col span={18}>
           <Row className={styles.name}>
-            <Typography.Text ellipsis>{follower.name}</Typography.Text>
+            {/* using ternary operrator for follower name,
```
gabek: The double pipe (`||`) isn't the ternary operator, it's the [logical OR operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR). But it's doing the right thing! The comment isn't necessary so you can probably just remove it. See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) for the ternary operator.

Yasir761: Sorry, I mistakenly write ternary operator instead of OR operator, sure I'll make the suggested changes.

gabek:
Looks like there's a new unnecessary change as `FollowerCollection.tsx` has a random whitespace change. Could you revert that and make sure only that what you expected to change gets pushed? You can look at the "Files changed" tab in this PR to see your changes after you push them to double check it's exactly what you intended.

gabek:
Hey there, looks like there's still two files changed. The easiest fix is just to restore the file you don't mean to change to the original version:

```
git checkout origin/develop web/components/ui/followers/FollowerCollection/FollowerCollection.tsx
```

Yasir761:
@gabek, Is still there is need to change something, if there let me know.

gabek:
Looks great! Thanks for taking care of this! :100:

PR #2709, opened 2023-02-14
fix placeholder styling (fixes #2677)

### LINKED ISSUE #2677: Followers: Avatar image should use placeholder

We should use a placeholder image for the avatar (`/logo`) if an image doesn't exist or it errors out when loading.

image

JacobWrenn:


gabek:
Thanks so much for looking at this!

Issue #1522, opened 2021-11-12
Architecture: Project file structure


gabek:
It's something I often think about, but never have put it down as a TODO.  The project could use some new structure and cleanup.  

Bradley did the first pass at giving the project some structure early on, but lack of definition and ease of just throwing it in the root directory made lack of organization commonplace.

It seems like a lot of projects have taken on the _[Hexagonal Architecture](https://github.com/katzien/go-structure-examples/tree/master/domain-hex)_ approach to project structure while splitting up the binary and the packages forcing definition and clearing up what goes where.

Now is probably not the time to do this with a bunch of in-flight feature branches going on, but I'm filing it so I actually get to it when it makes sense.

Gophers, feel free to chime in with any suggestions or opinions.

gabek:
This is being addressed with the backend architecture work going on.

PR #2712, opened 2023-02-17
Draft: refactor to common golang project layout

gravestench:
Hi, there! Cool project! Some of the nicest code I've seen!

Anyways, in this PR i have moved things around to adhere to a more [standard golang project layout](https://github.com/golang-standards/project-layout). This is mainly to reduce the noise in the repo root and to provide clarity.  For instance, moving the Owncast server `main.go` into `cmd/server` sets an example for including other build targets that would go along with Owncast that we may want in the repo (like a binary for streaming a test signal) but maybe not  in an actual release. Almost everything else has been moved into `internal/` as it is all intended to be used by Owncast and not imported to other projects.

________________________
* moved `main.go` build target into `cmd/server/`
* moved all owncast internal stuff into `internal/`
* moved build/docker scripts into `scripts/`
  * updated paths where necessary
* added installation script to `scripts/`
*  gone through all of the imports and cleaned up and organized them


gabek:
Hi there, great to meet you! Thanks for wanting to be involved!

I must admit I'm a little overwhelmed by such a large change. And while it's been a long filed todo (#1522) it would be nice to discuss some goals and specifics before making pulling the trigger on such a big refactor.

While I don't doubt moving files around for a better layout be more correct, would you instead be interested in helping on some higher-level architecture and work toward some specifics in smaller chunks? I'd love to work with somebody who cares about this kind of thing, and it's clear you do!

gravestench:
Yeah, totally. This only took about an hour or so and I was able to get everything compiling again. The embedded frontend build is also working.  I'm not sure why this PR shows conflicts, i rebased off of `development`.

gravestench:
I can keep this PR open and we can discuss what we want to make sure works. I know it's a big refactor, but it's mainly because of all of the static frontend files.

gravestench:
I'm wondering if all of the stuff inside of `web/_next/static` can be removed from the git cache and a rule be added to `.gitignore`

PR #2714, opened 2023-02-18
Draft: Change to service instance approach instead of global module state

gravestench:
At a high level, this branch serves to change several of the modules from a module-state strategy to a service-instance strategy. One of the main changes is in the ActivityPub layer, which includes the persistence layer.

To facilitate this change, it was necessary to create a high level abstraction which contains the instances of the sub-services. This high-level abstraction is located in `~/app` and is what the `main()` function now instantiates. This `App` struct contains the initialization logic for the sub-services, as well as the route handler initialization logic.

Another change was splitting up the route handler initialization logic, as many of the routes are coupled to the app sub-services and controllers. They are coupled in that their initializer methods must be passed a refernce to one of the services. This is due to the fact that the exported module-level functions have now been converted to methods of the service instances.

EDIT: I understand that this is a huge PR, but this really is a first step for  implementing the ActivityPub as service and getting rid of global persistence layer. I'll get the pipeline stuff passing again soon.


gravestench:
there's actually a lot wrong with this still, so will be continuing to push to this branch

gravestench:
this is compiling again, but getting deadlock with the database when serving.

PR #2722, opened 2023-02-21
Create stories for layout testing

### LINKED ISSUE #2668: Page vertical spacing layout issues

LBBO:
As discussed in #2676, this PR adds some stories that should showcase the different layouts. Since the header bar is an important part of the layout, the stories are created for the `Main` component. This also means that a bunch of services have to be mocked (they would otherwise try to make fetch requests to endpoints that don't exist in Storybook), so the PR also adds infrastructure for that. More specifically, it creates contexts for the services so they can be injected wherever needed and they can easily be substituted for mocks.

This is still in the draft stage because
- [x] The tests are failing on my GitHub branch
- [x] I'm interested in feedback for the service injection stuff
- [x] I'm unsure if the stories are complete (i.e. no relevant layouts are missing)
- [x] I'm not quite 100% sure if the stories are consistent (i.e. if they stay the same on every single render). All the mocking should™ guarantee that, but I'd like a second opinion on it
- [x] The stories aren't perfect, e.g. no actual video can be viewed in the "Online" stories. I don't think this is relevant for testing the layout and it might even help with consistency (see corresponding point above), but I'd also like some feedback on that.


gabek:
This is really fantastic, you put a lot of thought and effort into this. Thank you!

As a reference for testing, it got auto-deployed to https://629132c6e23893003a9e89c5-qrnjaoujrr.chromatic.com/?path=%2Fdocs%2Fowncast-layout-main--offline-desktop so that's where I'm checking it out.

Looks like the specific build/test error is this, does that sound accurate?

> Type error: Property 'getVideoQualities' is missing in type 'typeof VideoSettingsServiceMock' but required in type 'VideoSettingsStaticService'.

> I'm not quite 100% sure if the stories are consistent (i.e. if they stay the same on every single render). All the mocking should™ guarantee that, but I'd like a second opinion on it

> no actual video can be viewed in the "Online" stories. I don't think this is relevant for testing the layout and it might even help with consistency

It certainly looks like they should be consistent thanks to your mocks, and not having the video has more positives than negatives. No need to test that video is playing in layout validation, plus the consistency would be a mess like you said.

I noticed you built some context providers around Recoil. I could be wrong, but that feels like a bit of duplication of concerns, since Recoil is already providing a global state that exists when wrapping with `RecoilRoot`. So I'm a little confused why we would want both React Context and Recoil state at the same time.

<blockquote><div><strong><a href="https://629132c6e23893003a9e89c5-qrnjaoujrr.chromatic.com/?path=%2Fdocs%2Fowncast-layout-main--offline-desktop">Webpack App</a></strong></div></blockquote>

LBBO:
> Type error: Property 'getVideoQualities' is missing in type 'typeof VideoSettingsServiceMock' but required in type 'VideoSettingsStaticService'.

Ahh, I see. I didn't have time to look into the error yesterday, but this was an easy fix. VS Code just messed up a renaming because there's no proper inheritance going on here since inheritance doesn't cover static methods in TS.

> I noticed you built some context providers around Recoil. I could be wrong, but that feels like a bit of duplication of concerns, since Recoil is already providing a global state that exists when wrapping with RecoilRoot. So I'm a little confused why we would want both React Context and Recoil state at the same time.

I'm not quite sure what you're referring to here. I'm wrapping the `Main` component in a `RecoilRoot` in order to have control over the recoil state. That provider is exported by recoil and I'm just using it because `Main` depends on having a recoil state available.

The contexts that I added are just to "inject" services. The services are now "private" (not exported) and can only be retrieved via their corresponding context (e.g. `const ChatService = useContext(ChatServiceContext)`). I chose this because

1) context providers aren't strictly necessary as the context has a default value (the original service)
2) it allows me to override the services in the story. The original services would make HTTP requests that would fail, so I want to mock them.

Since I wouldn't consider these services to be state (they should never really be changed once created), I don't think recoil would be the right state for them. This context solution is kind of hacky, but it's the best thing I could come up with. Do you have a better solution? Or would you prefer the services to be integrated with the recoil state? If so, why?

gabek:
Got it! thanks for the additional details and clarifications around the use of context. I think there are places where I'm passing around the websocket using Recoil state, so that's probably something I should look at in the future and migrate to exposing them via Context instead.

I'm not sure if it's helpful, but there are a couple places in Storybook stories where I'm using `storybook-addon-fetch-mock` to return mock responses to stories. Here's an example: https://github.com/owncast/owncast/blob/develop/web/components/chat/ChatModerationActionMenu/ChatModerationActionMenu.stories.tsx#L6-L65

I don't know if that simplifies anything with regards to any of the mocks, since you had to mock the entire service, not just the response, but I thought I'd mention it.

And lastly, one question. Is there a way to surface both landscape and portrait versions of mobile and tablet stories?



<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast">owncast/ChatModerationActionMenu.stories.tsx at develop · owncast/owncast</a></strong></div><div>Take control over your live stream video by running it yourself.  Streaming + chat out of the box. - owncast/ChatModerationActionMenu.stories.tsx at develop · owncast/owncast</div></blockquote>

LBBO:
> 'm not sure if it's helpful, but there are a couple places in Storybook stories where I'm using storybook-addon-fetch-mock to return mock responses to stories. Here's an example:

Huh. Well that definitely eliminates the direct need for all the mocking stuff. If you want, I could undo all the context stuff I did and use these mock responses instead. The effect should be the same in this case. I personally like the idea of easily being able to switch out the service, but it's completely your choice.

> Is there a way to surface both landscape and portrait versions of mobile and tablet stories?

Yeah, [looks like it's possible via custom viewport definitions](https://storybook.js.org/docs/react/essentials/viewport). I can take a look at it later this week / this weekend.

gabek:
Doh! Sorry about that!

You're not wrong, being able to swap out the implementations of services has real benefits. I'm just not sure if they are benefits that we'd need or be able to take advantage of.

I think it's fine to leave things the way you put it together, since you put the work into it. If in the future it makes sense to standardize, or found that added abstraction makes it harder for people to understand then it can always be refactored into the simpler story mock responses.

LBBO:
Alright, there are now a ton of new viewports to choose from (imported from `@storybook/addon-viewport`). I convert them to landscape and portrait in order to be able to specify stories to start in landscape by default for Chromatic. For manual use, the `<->` button in the control bar between the dimension sizes should be more convenient. I hope this is what you were imagining!

LBBO:
After taking a quick look at the error messages and other PRs, it looks like this is not an issue with my changes. Please let me know if there's anything I can do to fix them (or if I'm wrong, of course!)

gabek:
Are you referring to the test failure? If so, yeah, that's not related here. It showed up the other day and I filed it as #2739. Hopefully I can track down what's wrong and fix it today so all these PRs don't look broken!

LBBO:
Yeah, that's what I was talking about. Just making sure this PR is complete :-)

gabek:
Thanks so much for expanding upon those Storybook stories! Those additional orientations will be super helpful, not just now but as new features get built too.

Also I went ahead and merged `develop` into this PR just for peace of mind to see all the tests pass.

PR #2723, opened 2023-02-22
Remove usage of the PropTypes dependency

### LINKED ISSUE #2710: Remove usage of the PropTypes dependency

All our components should be using the `FC` Functional Components props without the use of an added dependency.  I think there's probably only about five components that need to be changed so this can be removed. Search for `prop-types` to see where it's imported in the web app.

See https://owncast.online/components/?path=%2Fdocs%2Fowncast-documentation-building-frontend-components--page

dhanusaputra:
close #2710


gabek:
This is great, thanks so much @dhanusaputra!

PR #2741, opened 2023-02-26
Chore: update gzipping library to use CAFxX/httpcompression

### LINKED ISSUE #2697: Replace github.com/nanmu42/gzip + support brotli encoding

We use github.com/nanmu42/gzip as gzip middleware, but it's importing the entire Gin web framework. This dependency should replaced with something thinner.

https://github.com/owncast/owncast/blob/develop/router/router.go#L8

Pranav2612000:
Instead of using nanmu42/gzip which imports the whole gin framework, we replace it with CAFxX/httpcompression which is more lightweight.

Fixes #2697


Pranav2612000:
@gabek Can you take a look at this? I've tested that the project builds and works correctly

Pranav2612000:
Also I am not exactly sure how to test this, but the http://localhost:8080/hls/stream.m3u8 now has `Content-Encoding: br` header so I'm guessing it's using brotli

<blockquote><div><strong><a href="http://localhost:8080/hls/stream.m3u8">Error</a></strong></div></blockquote>

gabek:
I'll check it out, thank you!

gabek:
Nice, all encoding types are working as expected! Thank you for doing this! :100: 

For reference I tested with the following:

### No encoding

`curl http://localhost:8080 --silent --write-out "%{size_download}\n" --output  /dev/null`
631912


### deflate

`curl http://localhost:8080 --silent -H "Accept-Encoding: deflate" --write-out "%{size_download}\n" --output /dev/null`
81774

### gzip

`curl http://localhost:8080 --silent -H "Accept-Encoding: gzip" --write-out "%{size_download}\n" --output /dev/null`
81788

### brotli

`curl http://localhost:8080 --silent -H "Accept-Encoding: br" --write-out "%{size_download}\n" --output /dev/null`
70891

PR #2748, opened 2023-02-27
Fix Owncast Avatar URL for discord webhooks

SleepyLili:
Currently, discord webhook posts don't display any avatar for the bot user, only the default discord logo in a blue circle.
image

When I was browsing the webhooks code recently, I realized that the default discord logo appears because Owncast links a nonexistant logo .png to discord. (Which discord can't display, so it falls back to a default discord logo.)

With this change, the discord webhook will display the /logo file instead.
image


gabek:
Great, thanks!

PR #2769, opened 2023-03-03
removed timers from component in favor of css transition

### LINKED ISSUE #2342: "Go to last message" button covers chat messages only on production builds

t1enne:
Should fix #2342
@gabek Would you mind giving it a test run?


gabek:
This works in my testing! Replacing the timer with CSS transitions look to have one fewer variable and piece of state, so that's great. Let's get it in!

PR #2774, opened 2023-03-04
Chat join message for users with moderator flag is misaligned

### LINKED ISSUE #2763: Chat join message for users with moderator flag is misaligned

The flag should be after the username, all on one line. "joined the chat" should be on the same line.
image

Story: https://owncast.online/components/?path=%2Fstory%2Fowncast-chat-messages-chat-join--moderator

dhanusaputra:
close #2763 

image


gabek:
Awesome! Thanks so much for taking care of this! :100:

PR #2776, opened 2023-03-04
fixes #1946

### LINKED ISSUE #1946: Video variant admin interface should start with the default setting values
The default video variant that installs with Owncast is 1.2Mbps@24fs

Screenshot from 2022-06-13 11-46-01

However, if you go to add a new variant, it defaults to 800kbps.

image

It would be more consistent if it defaulted to 1200kbps @ 24fps. This would also make it easier for people to revert to the default should they have deleted/changed it.

And/or the interface could have a link to revert to the default settings.

## Use case

I had to do a fresh install to find out what the default video setting was when trying to revert to it.

JacobWrenn:
Creating a new video variant will now default to the same settings as the one initially installed.


gabek:
Nice, thanks so much for making that change! :+1:

PR #2783, opened 2023-03-08
fix: Prevent color selector from closing on choosing a color

### LINKED ISSUE #2782: Dragging color selector automatically closes selector

Each time you move the color selector it closes the color picker entirely and doesn't let you continue dragging to pick your desired color.

This can be found under General -> Appearance

image

Pranav2612000:
The bug was because on each click the state of the parent component was changing causing it to re-render. These re-renders then cascaded to its children recursively, resulting in the Color Picker component re-rendering which caused it to close. This PR uses useMemo & useCallback to ensure that there are no unnecessary re-renders which results in the color selector component staying open.

To make it easier to review, I've broken down the changes into smaller commits.

Kapture 2023-03-08 at 20 44 32

Fixes #2782

### DIFF
--- a/web/components/admin/config/general/AppearanceConfig.tsx
+++ b/web/components/admin/config/general/AppearanceConfig.tsx
@@ -70,18 +70,18 @@ const allAvailableValues = [...componentColorVariables, ...chatColorVariables, .
 );
 
 // eslint-disable-next-line react/function-component-definition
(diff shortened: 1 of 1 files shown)


gabek:
This is fantastic @Pranav2612000, thank you so much! :1st_place_medal:

PR #2786, opened 2023-03-09
feat: show a friendlier error msg in admin panel when unable to connect to Owncast Service

### LINKED ISSUE #2698: Give friendlier error in the admin when it cannot connect to Owncast service

Right now the admin throws ugly errors if it tries to access the Owncast service and it cannot. This can be improved by catching this somewhere and showing something friendlier. 

Current example:
image

Pranav2612000:
Modified the code to correctly handle the errors that were thrown when the frontend wasn't able to connect with the backend. Then, added an error object to keep track of the errors thrown and finally updated the Main Component to show the error if it exists.

image

Let me know what you think about this approach and what code / UI / Copy changes are required.

Fixes #2698

### DIFF
--- a/web/utils/server-status-context.tsx
+++ b/web/utils/server-status-context.tsx
@@ -91,6 +91,10 @@ const initialServerStatusState = {
     message: '',
     representation: 0,
   },
+  error: {
+    type: null,
+    msg: null,
+  },
 };
 
 export const ServerStatusContext = React.createContext({
@@ -112,17 +116,34 @@ const ServerStatusProvider: FC<ServerStatusProviderProps> = ({ children }) => {
   const getStatus = async () => {
     try {
       const result = await fetchData(STATUS);
-      setStatus({ ...result });
+
+      if (result instanceof Error) {
+        throw result;
+      }
+
+      setStatus({ ...result, error: { type: null, msg: null } });
     } catch (error) {
+      setStatus(initialStatus => ({
+        ...initialStatus,
+        error: {
+          type: 'OWNCAST_SERVICE_UNREACHABLE',
+          msg: 'Cannot connect to our servers. Some functionality may not work correctly',
+        },
+      }));
       // todo
     }
   };
   const getConfig = async () => {
     try {
       const result = await fetchData(SERVER_CONFIG);
+
+      if (result instanceof Error) {
+        throw result;
+      }
+
       setConfig(result);
     } catch (error) {
-      // todo
+      console.error(error);
     }
   };
 
--- a/web/utils/server-status-context.tsx
+++ b/web/utils/server-status-context.tsx
@@ -127,7 +127,7 @@ const ServerStatusProvider: FC<ServerStatusProviderProps> = ({ children }) => {
         ...initialStatus,
         error: {
           type: 'OWNCAST_SERVICE_UNREACHABLE',
-          msg: 'Cannot connect to our servers. Some functionality may not work correctly',
+          msg: 'Cannot connect to the Owncast service. Please check you are connected to the internet and the Owncast server is running.',
         },
       }));
       // todo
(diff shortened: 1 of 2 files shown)


Pranav2612000:
Also had a question about how Owncast is deployed. ( Will determine who frequently this `unable to connect to Owncast` error is shown )

- As a user, will I need to deploy the NextJS frontend and the go backend seperately? Or just running the go backend suffice? [ Because I've observed running the go backend also serves the contents in public folder which has the website ]
- How do we handle the multi-tenancy? Is every user's deployment completely separate with different DBs, backend & frontend servers? Or are there some common components that Owncast provides?

Apologies if these are silly questions.

gabek:
This is fantastic, thank you for looking at this!

> * As a user, will I need to deploy the NextJS frontend and the go backend seperately? Or just running the go backend suffice? [ Because I've observed running the go backend also serves the contents in public folder which has the website ]

Nope, just running the backend is all that is needed! When web (main frontend or admin) changes take place they automatically get built into static HTML and bundled into the Owncast binary. No web toolchain or runtime is required to run Owncast in production, only required when developing.

> * How do we handle the multi-tenancy? Is every user's deployment completely separate with different DBs, backend & frontend servers? Or are there some common components that Owncast provides?

There is no multi-tenancy support. One stream, one instance. That's why it's so quick and easy to get up and running in production, there's no external dependencies just to run it. Scaling is accomplished by using the S3 integration or using a CDN, but that's generally not required for most people.

web/utils/server-status-context.tsx
```diff
@@ -112,17 +116,34 @@ const ServerStatusProvider: FC<ServerStatusProviderProps> = ({ children }) => {
...
+        throw result;
+      }
+
+      setStatus({ ...result, error: { type: null, msg: null } });
     } catch (error) {
+      setStatus(initialStatus => ({
+        ...initialStatus,
+        error: {
+          type: 'OWNCAST_SERVICE_UNREACHABLE',
+          msg: 'Cannot connect to our servers. Some functionality may not work correctly',
```
gabek: Adding some specificity around "our servers" may make sense, especially to somebody not completely familiar with what's going on.  Plus the plural of "server" implies there's multiple servers.

[This](https://github.com/owncast/owncast/blob/develop/web/components/stores/ClientConfigStore.tsx#L39) is the error for the frontend. While not perfect, perhaps it might be a guide for what the admin error might look like.

gabek:
What are your thoughts about using something more blocking than the little message on top? If the Owncast service is offline, or somebody's internet connection is down and cannot reach the Owncast instance everything will fail, so perhaps a completely blocking state might be best. For the frontend I use a blocking modal.

Pranav2612000:
> What are your thoughts about using something more blocking than the little message on top? If the Owncast service is offline, or somebody's internet connection is down and cannot reach the Owncast instance everything will fail, so perhaps a completely blocking state might be best. For the frontend I use a blocking modal.

Yes. I think that's better. Have we used a similar Modal somewhere else?

gabek:
Yup! Nothing special in it though: https://github.com/owncast/owncast/blob/develop/web/components/modals/FatalErrorStateModal/FatalErrorStateModal.tsx

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast">owncast/FatalErrorStateModal.tsx at develop · owncast/owncast</a></strong></div><div>Take control over your live stream video by running it yourself.  Streaming + chat out of the box. - owncast/FatalErrorStateModal.tsx at develop · owncast/owncast</div></blockquote>

web/utils/server-status-context.tsx
Pranav2612000: Sure. Updating the message to 
`Cannot connect to the Owncast service. Please check your internet connection or if needed, double check this Owncast server is running.`.
Does this work?

Pranav2612000: On second thought, the message when viewed on the FatalModal seems a little odd in my opinion.
image

Another alternative is?
`Cannot connect to the Owncast service. Please check you are connected to the internet and the Owncast server is running.`
image

Which do you like more?

gabek: That works for me!

gabek:
Thanks for the adding this! Hopefully most people will never see this, but if they do it'll be a much nicer experience. Much better!

PR #2796, opened 2023-03-11
fix(issue: #2772) change text color white to black

### LINKED ISSUE #2772: Chat user details modal in the admin has incorrect white text

It should be black text like everything else.

image

jackgris:


gabek:
Thanks so much for the fix!

PR #2809, opened 2023-03-12
fix: #2755 clicking more details on un-registered chat user throws ex…

### LINKED ISSUE #2755: User moderator chat details modal throws exception

1. Become a moderator
2. Select moderator menu of a chat message from the primary chat web interface
3. Choose "More info"

image

thisprojects:
Fixes #2755 

Un-registered users have a `scope` of `undefined` - not sure if this should be the case or not. Fix checks validity of `scope` variable before applying map method.


gabek:
Makes total sense, thanks for tracking this down! :1st_place_medal:

PR #2829, opened 2023-03-16
Menu state in admin not saved after page refresh

### LINKED ISSUE #2807: Menu state in admin not saved after page refresh

Open a page that's behind a drop down menu. For example Hardware info and refresh the page in your browser. All the menu's are collapsed and you current place in Owncast is no longer highlighted.

owncast-menu

dhanusaputra:
close #2807 

for highlight state, update menu keys to use route format so we can use it as reference
for collapse state, put the state in openKeys

[Screencast from 16-03-23 14:14:47.webm](https://user-images.githubusercontent.com/35093673/225542305-4ac0933b-69ea-455c-beea-bb238d1ad14d.webm)


gabek:
Works great! I had forgotten I previously was using the key for highlight state and then the root path of the admin changed. Thanks for looking at this!

PR #2830, opened 2023-03-16
S3 headers

### LINKED ISSUE #2771: Set correct content-type and cache-control for S3 uploads

Bumped into this when figuring out why using Cloudflare R2 isn't working as it should as S3 storage.

For `.m3u8` playlists locally served, the headers are set like this:

```
cache-control: no-cache, no-store, must-revalidate
content-type: application/x-mpegURL
```

For `.m3u8` playlists served over S3, the headers are set like this:
```
cache-control: max-age=0
```
And `content-type` is not set at all. Some S3 providers like Backblaze B2 seem to guess the correct `content-type` based on the file extension, but not all do. For example Cloudflare R2 doesn't.

The `s3manager` for Go does support setting these values. I suggest making them universal for playlists and chunks served locally or over S3.

JacobWrenn:
Cache control and content type headers are now set for m3u8 playlists in s3 as specified in #2771


gabek:
I fixed the HLS + S3 test to run again in CI so I merged `develop` into your branch so it would run as a sanity check.

gabek:
Thanks for adding the headers!

PR #2860, opened 2023-03-22
feat: add support to dev with nix

enieber:
add support develop with nix-shell


gabek:
Could you explain what this is?

enieber:
the [nix](https://nixos.org/) is package manager to reproducible builds with very help to use in developer install all dependencies isolate in local machine.

gabek:
If this is for a package manager wouldn't this file exist in the package manager's repository?

PR #2892, opened 2023-04-01
FIX : Admin custom CSS area forces entire page to scroll #2886

### LINKED ISSUE #2886: Admin custom CSS area forces entire page to scroll

I think the code editor itself should be allowed to scroll horizontally, but not force the entire page to overflow and force horizontal scrolling.

image

dev265545:
Fixes #2886 
Making codemirror component vertically scrollable


gabek:
Hi there! Thanks for looking at this!

It looks like the entire page still scrolls horizontally if there's a long line in the editor, such as a comment. I wonder if a specific width needs to be added to the editor or something so it doesn't grow horizontally and instead just overflows internally and allows only the editor to scroll.

image

gabek:
Looks great! Thank you!

PR #2896, opened 2023-04-03
fix: Prevent storybook from making websocket connections

### LINKED ISSUE #2862: Storybook layout stories try to make websocket connections

Obviously there's no server to connect to.

image

Pranav2612000:
I've added another websocket config variable to the client config and then using that to determine whether to start the websocket connections or not. Let me know what you feel of this approach.
Fixes #2862

### DIFF
--- a/web/components/layouts/Main/Main.stories.tsx
+++ b/web/components/layouts/Main/Main.stories.tsx
@@ -38,7 +38,7 @@ export default {
 
 // mock the Websocket to prevent ani webhook calls from being made in storybook
 // @ts-ignore
-window.WebSocket = () => {}
+window.WebSocket = () => {};
 
 type StateInitializer = (mutableState: MutableSnapshot) => void;
 
--- a/web/components/layouts/Main/Main.stories.tsx
+++ b/web/components/layouts/Main/Main.stories.tsx
@@ -36,6 +36,10 @@ export default {
   },
 } satisfies ComponentMeta<typeof Main>;
 
+// mock the Websocket to prevent ani webhook calls from being made in storybook
+// @ts-ignore
+window.WebSocket = () => {}
+
 type StateInitializer = (mutableState: MutableSnapshot) => void;
 
 const composeStateInitializers =


Pranav2612000:
Hey @gabek! Did you have time to review this?

web/components/layouts/Main/Main.stories.tsx
```diff
@@ -49,6 +49,7 @@ const defaultClientConfig = {
   name: "Spiderman's super serious stream",
   summary: 'Strong Spidey stops supervillains! Streamed Saturdays & Sundays.',
   extraPageContent: 'Spiderman is **cool**',
+  socketPathOverride: '',
```
gabek: I feel like if we had an actual use case for needing to override the path for the websocket connection that using adding a new property and using for two things could make sense, but in this case it's mostly just being added as a way to disable the websocket. So I'm not sure it's the right answer in this particular case when we're just trying to stop the websocket from connecting in Storybook.

gabek:
Just like we mock API requests and responses in Storybook I wonder if there's a way to do the same with Websockets.

web/components/layouts/Main/Main.stories.tsx
Pranav2612000: Makes sense. I guess we are already using the `storybook-addon-fetch-mock` mock for fetch requests, I can try using that to see if it's able to mock Websocket requests. If not, are you open to using another library to mock Websockets? ( Or if we don't want to burden ourselves with another library, we can also try to mock the `window.Websocket` func )


Regardless, this won't **stop** storybook from making websocket connections, but atleast the connections would be properly handled this time.
What do you think?

gabek: Good point! That's probably the easiest solution, by wrapping the Storybook story with a component that overwrites the websocket's constructor to just be a no-op. Then it should just silently do nothing. That way it also doesn't change the real code, and only impacts the Story.

web/components/layouts/Main/Main.stories.tsx
```diff
@@ -36,6 +36,10 @@ export default {
   },
 } satisfies ComponentMeta<typeof Main>;
 
+// mock the Websocket to prevent ani webhook calls from being made in storybook
+// @ts-ignore
+window.WebSocket = () => {};
```
Pranav2612000, line 41: Does this work? @gabek 
I've not wrapped the component inside, but I've mocked the WebSocket func to not do anything

gabek: That works!

gabek:
Thanks for looking at this and coming up with a simple solution!

PR #2901, opened 2023-04-03
FIX #2897 Admin: "Copy" button for stream keys copy asterisks if key is hidden 

### LINKED ISSUE #2897: Admin: "Copy" button for stream keys copy asterisks if key is hidden
In the admin web ui for stream keys (/admin/config/server/ > "Stream keys"), if you click the "copy" button while the key is hidden ("********"), the placeholder asterisks are copied to the clipboard instead of the key.

Tested on Firefox 111 (Linux), Ungoogled Chromium 111 (Linux)

dev265545:
Fixes  #2897 
Adding a copyText function to solve this issue


gabek:
Awesome, thanks Dev. Works great! :100:

PR #2903, opened 2023-04-03
FIX : Chat: words containing one or more dots are turned into links #2898

### LINKED ISSUE #2898: Chat: words containing one or more dots are turned into links

In chat messages, words containing one or more dots (e.g. "A.B", "a.b.c") are turned into clickable links.
Numbers don't seem affected (eg. "123.45").

dev265545:
Fixes #2898 
Making validateTLD to true and also adding a parameter of customTLDs.
Using customTLDs we can add  domain names which will also be matched as a link


gabek:
Perfect! Thanks for looking up that additional `customTLDs` property.

gabek:
Tests seem to be failing after merging this in . I'm going to revert this change.

gabek:
Fixed!

PR #2928, opened 2023-04-16
Fix : autogenerated streaming key contains bad characters #2919

### LINKED ISSUE #2919: autogenerated streaming key  contains bad characters

on nightly version v20230410-nightly

while testing new feature of multiple stream keys I had some trouble.

it appears some characters are not compatible, for example this key `OFDIZcapUVR!0IL8ovX8O2WG3%XjjU`

I suppose it is related to `%`. I've tested other generated keys without any issue.

dev265545:
Fixes #2919


gabek:
Thanks, Dev!

I just wired up Javascript unit tests for the first time and it seems like this would be a good candidate for that. Could you add a unit test in `web/tests` to this PR to verify the keys that are being generated?

dev265545:
Yes, I will definitely do that

gabek:
Yay for tests! I should have set that up a long time ago and it would have changed how frontend code was written to be a bit more modular. Better late than never, I suppose!

PR #2929, opened 2023-04-16
feat: add support for robots.txt disabling search indexing

### LINKED ISSUE #2684: Option to prevent Google indexing of Owncast pages
### Feature request

An experimental instance of mine got indexed by Google. Currently the web interface doesn't prevent indexing by search engines. 

I propose a checkbox in the admin interface that reads "Prevent search engines from indexing this Owncast instance". It could be unchecked by default. When it's checked, the following could be placed in the `<head>` of the public web app.

`<meta name="robots" content="noindex, nofollow">`

If the checkbox is not checked, the following could be placed:

`<meta name="robots" content="index, nofollow">`

#### Making the case for nofollow
Since most Owncast interfaces provide a chat interface that let's the public enter anything, it's good practice to tell search engines to not follow external links on the page. Because nofollow links do not pass PageRank they don't impact search engine rankings and this prevents spammers for trying to boost their SEO juice via popular Owncast instances.

gabek:
Can toggle disabling search engine indexing via `robot.txt` directives. Closes #2684

### DIFF
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -751,6 +751,26 @@ func SetHideViewerCount(w http.ResponseWriter, r *http.Request) {
 	controllers.WriteSimpleResponse(w, true, "hide viewer count setting updated")
 }
 
+// SetDisableSearchIndexing will set search indexing support.
+func SetDisableSearchIndexing(w http.ResponseWriter, r *http.Request) {
+	if !requirePOST(w, r) {
+		return
+	}
+
+	configValue, success := getValueFromRequest(w, r)
+	if !success {
+		controllers.WriteSimpleResponse(w, false, "unable to update search indexing")
+		return
+	}
+
+	if err := data.SetDisableSearchIndexing(configValue.Value.(bool)); err != nil {
+		controllers.WriteSimpleResponse(w, false, err.Error())
+		return
+	}
+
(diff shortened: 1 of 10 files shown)


rmens:
Hi Gabe, any reason you chose the robots.txt route and not the metatags? This covers the 'don't index this instance' scenario, but doesn't add any metatags that encourage search engines to index the site.

gabek:
The number of things that require server side rendered page content the better. There's currently no guarantee that the meta tags would be rendered for a search bot without Javascript. `robots.txt` is server rendered, and therefore always available because it doesn't require javascript. If somebody flips the switch to disable search engine indexing I don't want it to be a "maybe, depending on the search engine".

As far as I've been able to tell Owncast instances have been indexed without problems so far, so the real change here is to tell search engines not to index it like. If explicit `Allow` directives for some reason are needed they can be added, but I don't believe they have to be.

rmens:
Makes sense. I wasn't aware that the metatags were client side rendered (being a simple PHP/MySQL guy 😌).

gabek:
Such is the drawback of completely client-side rendered pages! Technically a part of the build step performs an optimization where it will render static HTML instead of Javascript components for items it believes should be static and in-lined, but it's up to the compiler to make that call so there's no guarantee that HTML will exist before Javascript kicks in. It usually doesn't matter much, but in this case it would.

PR #2932, opened 2023-04-17
Removing % from streamkey generator, and adding unit test for same.

### LINKED ISSUE #2919: autogenerated streaming key  contains bad characters

dev265545:
Fixes autogenerated streaming key contains bad characters #2919 
Also changed the transform function on jest.config.js to 
``` 
transform: {   '^.+\\.(js|jsx|ts|tsx)$': ['babel-jest', { presets: ['next/babel'] }], },
```


gabek:
Thanks so much, Dev!

PR #2958, opened 2023-04-27
Feat/autoscroll on homepage tabs

### LINKED ISSUE #2936: Tabs: Jump to content when selecting tab
I'm not 100% this is a thing that should be done, but it _feels_ like it would be nice.

Currently on smaller screens there's a scenario (such as the iPad example) where you see the tab bar, so you select the tab you want to see, and then you have to scroll around to see that content.

It feels like it would be nice when you select a tab, because it's obvious you want to view that content, to auto-scroll down so that content is in view.

What would others think?

https://user-images.githubusercontent.com/414923/233722977-ea1f4415-d975-43a6-8bf1-e7f3795213c7.MOV

Pranav2612000:
Fixes #2936

### DIFF
--- a/web/components/ui/Content/DesktopContent.tsx
+++ b/web/components/ui/Content/DesktopContent.tsx
@@ -34,7 +34,7 @@ const FollowerCollection = dynamic(
   },
 );
 
-const autoScrollToSection = (key) => {
+const autoScrollToSection = key => {
   let elementToScroll = null;
   if (key === '2') {
     elementToScroll = document.getElementById('custom-page-content');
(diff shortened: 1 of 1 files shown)


gabek:
What do you think? Do you think we should have something like this? I'm not sure myself. It scrolls a bit far, cutting off the content, but that could be fixed I'm sure.

Pranav2612000:
I agree. It's not the best experience

gabek:
Thanks for willing to experiment with me and trying it out! It was worth a shot :)

Issue #2962, opened 2023-04-27
Mobile layout: Footer wraps


gabek:
![image](https://user-images.githubusercontent.com/414923/235003961-31621a0a-822e-4418-8752-f30db4990d20.png)

gabek:
Previously the footer was hidden on mobile to save space, so I'm just going to hide it again.

PR #2966, opened 2023-04-28
fix: #2956 Mobile Landscape Fix

### LINKED ISSUE #2956: Mobile layout: Landscape + mobile breaks page content
When turning the iPhone horizontal to landscape the tabbed content goes away. 

image

thisprojects:
Mobile landscape should now work properly. The user should be able to scroll down to the chat section and other tabs.

### DIFF
--- a/web/components/video/OwncastPlayer/OwncastPlayer.module.scss
+++ b/web/components/video/OwncastPlayer/OwncastPlayer.module.scss
@@ -12,21 +12,14 @@
     max-height: 75vh;
   }
 
-  @include screen(desktop) {
-    // prevent sidebar from overlapping stream
-    // padding-right: 320px;
-  }
-
   //set height of player for tablet
   @include screen(tablet) {
     height: 400px;
-    max-height: 400px;
   }
   
   //set height of player for mobile
   @include screen(mobile) {
     height: 250px;
-    max-height: 250px;
   }
 
   .player,
--- a/web/components/layouts/Main/Main.module.scss
+++ b/web/components/layouts/Main/Main.module.scss
@@ -10,15 +10,5 @@
     left: 0;
     bottom: 0;
     right: 0;
-    // this one is for fixed footer
-    margin-bottom: 30px;
-  }
-
-  @include screen(mobile) {
-    margin-bottom: 0px;
-
-    footer {
-      display: none;
-    }
   }
 }
--- a/web/components/ui/Content/ActionButtons.tsx
+++ b/web/components/ui/Content/ActionButtons.tsx
@@ -10,6 +10,7 @@ import styles from './Content.module.scss';
 import { ActionButton } from '../../action-buttons/ActionButton/ActionButton';
 
 interface ActionButtonProps {
+  online: boolean;
   supportFediverseFeatures: boolean;
   externalActions: ExternalAction[];
   supportsBrowserNotifications: boolean;
@@ -30,6 +31,7 @@ const NotifyReminderPopup = dynamic(
 );
 
 const ActionButtons: FC<ActionButtonProps> = ({
+  online,
   supportFediverseFeatures,
   supportsBrowserNotifications,
   showNotifyReminder,
@@ -67,7 +69,7 @@ const ActionButtons: FC<ActionButtonProps> = ({
           )}
         </ActionButtonRow>
       </div>
-      <div className={styles.mobileActionButtonMenu}>
+      <div className={styles.mobileActionButtonMenu} style={!online ? { right: '10px' } : {}}>
         <ActionButtonMenu
           className={styles.actionButtonMenu}
           showFollowItem={supportFediverseFeatures}
--- a/web/components/ui/Content/Content.module.scss
+++ b/web/components/ui/Content/Content.module.scss
@@ -4,24 +4,52 @@
   padding: var(--content-padding);
 }
 
-.lowerSectionMobile {
+.contentHeaderMobile {
+  @include screen (tablet) {
+    display: flex;
+  }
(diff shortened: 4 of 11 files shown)


gabek:
Looks like something isn't working as expected, all the mobile tests are failing on mobile. You can see what's going on here: https://cloud.cypress.io/projects/wwi3xe/runs/5592/overview?reviewViewBy=FAILED&roarHideRunsWithDiffGroupsAndTags=1

thisprojects:
Is the footer supposed to be hidden on mobile?

gabek:
I hid the footer on mobile yesterday since it had previously been that way, and its contents was wrapping weird, so I figured I'd just reclaim the space on the small screens: #2962

thisprojects:
Cool 👍 will make a few more tweaks and then I think its done.

gabek:
It's usable now!

https://user-images.githubusercontent.com/414923/235229993-853200ca-940d-4696-b4ec-ccea95054a8c.MOV

It's odd that testing in Chrome under dev tools looks completely different however. It uses the side by side layout where on a real device it stayed vertical. What do you think it should be?

image

thisprojects:
Mobile UI/X is difficult 🤔. From a users motivational view, the main reason they have visited the page is to watch the stream and chat. So maybe the side by side "Desktop" experience is best for landscape, as these two features are the most prominent.  

I think maybe the best thing to do is to refactor Mobile Content and Desktop Content into a single component and have the experience be governed by CSS media queries, rather than the "isMobile" boolean. I'll have a go at this over the weekend.

web/components/video/OwncastPlayer/OwncastPlayer.module.scss
```diff
@@ -29,6 +29,11 @@
     max-height: 250px;
   }
 
+  @include screen(mobileLandscape) {
+    height: 250px;
+    max-height: 250px;
```
Pranav2612000: Since we are setting the height to fixed 250px. Do we also need the `max-height` here?

gabek:
> Mobile UI/X is difficult thinking. From a users motivational view, the main reason they have visited the page is to watch the stream and chat. So maybe the side by side "Desktop" experience is best for landscape, as these two features are the most prominent.

I agree that the side by side is a better experience on landscape mobile.

> I think maybe the best thing to do is to refactor Mobile Content and Desktop Content into a single component and have the experience be governed by CSS media queries, rather than the "isMobile" boolean. I'll have a go at this over the weekend.

That makes sense that even on landscape because of that mobile bool it's still using the vertical layout. Originally the desktop and mobile layouts were the same layout component, but I ended up splitting them out because mobile started becoming so different than desktop, and it was becoming harder to maintain with so many different conditionals and sub-components. Happy to have them combined again if that makes more sense, since that would be more "correct" anyway.

thisprojects:
It's a tough one... media queries are the best option for presentational logic, but they can become complex and less visible due to abstraction. On the other hand, managing mobile presentation via javascsript is more visible (in the component) but you rely on re-renders catching screen size changes. Ideally you don't want your components to be re-rendering that often.

I'm going to try and find a happy balance between the two. We will still need the isMobile boolean for some things.

thisprojects:
Weird test fail. Output says "New Owncast Server" text is repeated, but doesn't appear in the rendered page or locally (in dev tools or browser). Will continue to dig into it.

gabek:
I wonder if it's because of the merger of the two layouts. Is there more than one instance of the `ContentHeader`, one for mobile and one for desktop but one is hidden? Just a guess.

thisprojects:
Yeah, was thinking the same 🤔 doesn't show up in the DOM locally though. Might have to govern that particular section with conditional logic again.

thisprojects:
Seems like that fixed it. Any idea what the causing the Lint / prettier fail? Does it just need to be re-run?

gabek:
Looks like Prettier was trying to commit some formatted files but they were rejected due to permissions on your branch. I'm guessing this has to do with a change I made to the prettier workflow job, let me revert that see.

thisprojects:
Any issues with the changes in storybook?

gabek:
The first screenshot in the UI tests show the tabbed content no longer spanning the width that it used to. You can see them here: https://www.chromatic.com/build?appId=629132c6e23893003a9e89c5&number=1316

image

gabek:
It looks like the margin around the whole bottom half (content header + page content) is missing, since the content header moved to the left and the page content (Spidey is cool) is now hard left aligned.

gabek:
It seems like this might be taking too many steps backwards as far as breaking more things. I wonder if it's acceptable just to have a lousy landscape phone layout right now. I fear as we keep making huge layout changes we're going to be working on this indefinitely. The layout changes you made previously got us so much closer, and I hate that we're moving backwards.

There's a couple other issues that I haven't started work on because I was afraid they would conflict with the changes you're doing, and one of them is a pretty big issue, that the player is cut off, so that needs to be worked on asap, but it's layout related as well.

What do you think? Is what is in `develop` good enough to move forward with?

thisprojects:
family-guy-css

Yeah, we're kind of in this situation 😆. If you're happy with develop let's roll with it, I'll still be on hand to help with any problems that occur during acceptance testing / post release.

gabek:
We can always get back to it. I don't feel too badly with landscape mobile being the thing that's pushed off, I doubt many people will use it. I lock my device not to ever rotate, personally :)

[There's some other, smaller issues](https://github.com/owncast/owncast/milestone/18) if you're interested in tackling some more defined layout issues!

thisprojects:
Will do - closing this PR for now.

PR #2974, opened 2023-04-30
FIX #2933 Make admin stream title input more visible

### LINKED ISSUE #2933: Make admin stream title input more visible

In the admin, the input used to change the stream title is barely noticeable compared to the one in v0.0.13 because:
- it's tucked against the top of the page
- its unfocused state colors are very muted against the background
- the placeholder text is very light against the white input background
- ~~it doesn't have a visible "Update" button like the old one, making it less clear it's an input~~ (0.0.13 was the same for this)

I could argue that the placeholder color is too light for all inputs, but it's less of a problem for the other ones which all(?) have a label/description.

Below, and for illustration purposes only, is a proposition of how small quick style changes could help a little:

```css
.app-container .layout-header {
    padding-top: 1em;
}
.app-container .layout-header #field-streamTitle::placeholder {
    color: #252525;
}
.app-container .layout-header .field-streamTitle {
    border: 1px solid rgba(120,113,255,0.5);
}
```

image
image

dev265545:
FIXES #2933 
image

### DIFF
--- a/web/jest.config.js
+++ b/web/jest.config.js
@@ -1,5 +1,5 @@
 module.exports = {
-  transform: {'^.+\\.(js|jsx|ts|tsx)$': ['babel-jest', { presets: ['next/babel']}]},
+  transform: { '^.+\\.(js|jsx|ts|tsx)$': ['babel-jest', { presets: ['next/babel'] }] },
   testEnvironment: 'node',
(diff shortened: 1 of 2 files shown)


gabek:
Awesome! Certainly improved, thanks so much!

Issue #2071, opened 2022-08-30
Support ffmpeg 6


dumbasPL:
### Share your bug report, feature request, or comment.

In ffmpeg version 5 the three distinct pixel formats seems to have been merged into one named just `vaapi`

[The code](https://github.com/owncast/owncast/blob/develop/core/transcoder/codecs.go#L178) currently forces the `vaapi_vld` pixel format that is no longer present in the latest version of ffmpeg

Here is a comparison of available pixel formats on ffmpeg 5.0.1 (from alpine 3.16) and 4.4.1 (from alpine 3.15).
```
$ docker run --rm -it alpine:latest sh -c "apk add ffmpeg > /dev/null && ffmpeg -pix_fmts | grep vaapi" 
ffmpeg version 5.0.1 Copyright (c) 2000-2022 the FFmpeg developers
  built with gcc 11.2.1 (Alpine 11.2.1_git20220219) 20220219
  configuration: --prefix=/usr --enable-avfilter --enable-gnutls --enable-gpl --enable-libass --enable-libmp3lame --enable-libpulse --enable-libvorbis --enable-libvpx --enable-libxvid --enable-libx264 --enable-libx265 --enable-libtheora --enable-libv4l2 --enable-libdav1d --enable-lto --enable-postproc --enable-pic --enable-pthreads --enable-shared --enable-libxcb --enable-librist --enable-libsrt --enable-libssh --enable-libvidstab --disable-stripping --disable-static --disable-librtmp --enable-libaom --enable-libopus --enable-libsoxr --enable-libwebp --enable-vaapi --enable-vdpau --enable-vulkan --enable-libdrm --disable-debug
  libavutil      57. 17.100 / 57. 17.100
  libavcodec     59. 18.100 / 59. 18.100
  libavformat    59. 16.100 / 59. 16.100
  libavdevice    59.  4.100 / 59.  4.100
  libavfilter     8. 24.100 /  8. 24.100
  libswscale      6.  4.100 /  6.  4.100
  libswresample   4.  3.100 /  4.  3.100
  libpostproc    56.  3.100 / 56.  3.100
..H.. vaapi                  0              0      0

$ docker run --rm -it alpine:3.15 sh -c "apk add ffmpeg > /dev/null && ffmpeg -pix_fmts | grep vaapi"
ffmpeg version 4.4.1 Copyright (c) 2000-2021 the FFmpeg developers
  built with gcc 10.3.1 (Alpine 10.3.1_git20211027) 20211027
[...]

PR #2976, opened 2023-05-01
Use VA-API hardware decoding and scaling

hagn:
The VA-API encoding currently still uses a lot of CPU resources for the video decoding step.
I don't know if this previously was a specific use case for systems with only an H.264 encoder but no decoder (if something like this even exists) or just an oversight. Otherwise a separate entry in the "Video Codec" drop-down is needed.

These changes switch to the VA-API hardware decoder. But I still had performance issues (frame drops) if video resizing was enabled. For that I also added support for the hardware scaler ("scale_vaapi").
Currently ffmpeg still receives the "-sws_flags bilinear" option. I guess this does nothing while using "scale_vaapi" but it's not the nicest thing. According to the help for "scale_vaapi" the default scaling mode should be high quality. So I didn't see any reason to change it.

The hardware scaler should probably also be enabled for NVENC but I can't test it with my setup as I only have an Intel UHD 630.
I only saw that there are two different scalers available for NVENC. "scale_npp" and "scale_cuda". But this should probably looked at separately.

I also wasn't sure about the coding style in the "GlobalFlags" method. The NVENC implementation filled the array differently compared to VA-API. I adjusted it to the same style but maybe that is wrong.

(This is my first pull request. So I don't know exactly if I should have created an issue first.)


gabek:
Thanks so much for looking into adding additional hardware features! It's not necessarily that things were overlooked, but after the first set of hardware transcoding features went in they weren't taken advantage of much, so future investment just wasn't worth it. As far as I'm aware, most everyone who runs Owncast run on servers without dedicated video hardware. Adding hardware support didn't benefit many people so I never came back to it.

Right now the focus is to get v0.1.0 out the door, and a rather large change such as touching the transcoding pipeline would require a lot of testing across different pieces of hardware. So unfortunately I can't get this tested and merge in right now. But hopefully it can be tested and brought in for the v0.1.1 release.

For future things, like you mentioned, it's best to open an issue so it can be discussed ahead of time. This is fine though, I'll add it to the v0.1.1 milestone and hopefully it can get tested for that release.

hagn:
Ah okay. I was already wondering why some of the hardware encoders looked a bit unfinished. With my locally hosted server I'm probably in the minority.

I was already expecting that it won't fit into the upcoming release, so don't worry about it.
But if desired I could take a look at some of the existing hardware transcoder implementations to make sure they run as efficient as possible. Maybe with an issue attached this time ;)

gabek:
That would be awesome! It would certainly be useful to some people for sure! Even I run an Owncast instance at home on a server with an Nvidia card, but I just use the software encoding because I feel like I should be running it like everyone else is.

gabek:
There's a related issue that's been sitting around (for all the same reasons as mentioned above), for supporting the newer versions of ffmpeg. If you're interested in this side of things you might be interested in helping out with that issue since it's dealing with the same stuff! https://github.com/owncast/owncast/issues/2071

gabek:
This PR for some reason won't let me push fixes to the conflicts, it must be locked to just you. Would you mind looking at the conflicts? It's in the test.

hagn:
I currently don't have time to test it, but it should work like this. I also added write permissions if anything else comes up.

gabek:
Works great for me in testing. Thanks again for the improvement!

PR #2978, opened 2023-05-02
Added margin under the page content #2963

### LINKED ISSUE #2963: Add margin under the page content

So it doesn't sit right up against the footer.

image

VaibhavKambar7:


web/styles/ant-overrides.scss
```diff
@@ -160,6 +160,7 @@ DROPDOWN
   height: 100%;
   .ant-tabs-content-holder {
```
gabek, line 161: This looks like it would add 8px of margin under every use of tabs, anywhere. Instead, it would be better to only add spacing to just this one layout.

PR #2987, opened 2023-05-05
fix:#2954-player-gets-cut-off-vertically

### LINKED ISSUE #2954: Layout: Player gets cut off vertically on some screens

There seems to be one layout issue with regards to the player where the height seems too short. The `develop` branch correctly sizes it as a 16:9 aspect ratio. This was tested on a desktop browser.

After PR #2863: 

image

Previous to PR in the `develop` branch for comparison:

image

thisprojects:
Should fix the player cut off issue.

### DIFF
--- a/web/components/layouts/Main/Main.module.scss
+++ b/web/components/layouts/Main/Main.module.scss
@@ -2,7 +2,7 @@
 
 .layout {
   // this margin is for fixed header
-  margin-top: 70px;
+  margin-top: var(--header-height);
   background-color: var(--theme-color-main-background);
   @include screen(tablet) {
     position: absolute;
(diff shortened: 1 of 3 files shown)


gabek:
Thank you! A height was already set [above](https://github.com/owncast/owncast/pull/2987/files#diff-48d6e737ac04c7897bbd2ba46f035472e5a234749dd1ceac2fac6141b9e5037dR4) but using `var(--header-height)` and is set as `4.3rem` by default.  I changed the variable value to the `70px` you set in this PR, and then used that var in the `margin-top` you added.

When the viewport is smaller it decreases the height of the footer, so I changed that to update that var's value globally, and it all seems to be working. 

Do you see any downsides to these changes? [Here's my tweaks](https://github.com/owncast/owncast/pull/2987/commits/822a351b797327962298ad47af5bd04dfa513245).

thisprojects:
Looks good to me Gabe 👍

gabek:
Wohoo, thanks!

PR #2993, opened 2023-05-06
Fixes : Admin: Custom emoji page layout 

### LINKED ISSUE #2979: Admin: Custom emoji page layout

Right now it's one very long list that takes a lot to scroll through. Changing this to a grid would make it easier to browse through your emoji collection.

image

dev265545:
Fixes : #2979 
I have changed the layout to grid, this is a basic layout if you want to change any specifics from design point of you I can make  updates.
Also, I have changed the position of Upload button at the top as it make more sense I think.


gabek:
This is awesome, thank you for working on this! It's so much better.  One thought I have now that I see it like this, is we could probably save even more room by not dedicating so much space to the delete button. 

What do you think about adding a little delete icon in the corner of the item instead?

image

dev265545:
I have made few changes to the design, I have removed the emoji name and  put the name as tooltip.

gabek:
Very nice! What are your thoughts about putting the delete button in the corner and making it smaller? It seems like having a large button isn't necessary.

dev265545:
If you want to try different icons I can do that

gabek:
It's not so much the icon itself, it's just that there's this big button using up so much room, when in reality most people won't need to click it very often. So reducing the size and repositioning it seems to make sense.

dev265545:
Hey, I have updated the design

gabek:
I added a couple small style tweaks to the button, but we're good to go! This page layout change makes it a lot easier to browse these images. Adding the upload button to the top is going to be much easier for people, too. Thank you for working on this!

PR #3007, opened 2023-05-12
fix(mobile): toggle tablet content layout based on online state.

### LINKED ISSUE #3003: Tablet layout problems

I thought we were in good shape. I was wrong. The web layout problems continue. I'm not really sure how to fix all these web problems that continue to surface.

image

gabek:
Closes #3003

@thisprojects Curious if you could take a look and see if this is a good solution. The issue that I saw was the hard-coded 430px was being set even on tablet if the offline banner was not 430px. So instead I set it to relative positioning when offline, and then moved the hard coded 430px absolute positioning when online.

Offline:

image

Online:

image

### DIFF
--- a/web/components/ui/Content/Content.module.scss
+++ b/web/components/ui/Content/Content.module.scss
@@ -14,8 +14,7 @@
   width: 100%;
 
   @include screen(tablet) {
-    //sets the position of tabbed content for online mode
-    top: 430px;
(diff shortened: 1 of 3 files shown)


thisprojects:
Looks good to me Gabe! 👍

gabek:
Thanks for looking!

Issue #3012, opened 2023-05-13
The layout problems will never end. The web needs to burn.


gabek:
### Share your bug report, feature request, or comment.

Is it too late to build Owncast's frontend as a native iOS app? These layout problems are really taking a toll on my sanity and they're impossible to fix without breaking everything else.

![image](https://github.com/owncast/owncast/assets/414923/2017bbc9-f893-4fb9-9fb8-a6fc19a64b8f)

gabek:
Obviously I'm joking about the iOS app thing. But I'm not joking about all the web layout problems. It feels like this release will never be release ready, all the mobile crap is so buggy.

amrinder1189:
Hi, I am working on this

thisprojects:
I think this may have been introduced when we changed the offline banner from absolute to relative positioning. We might be able to fix the offline-banner using flex basis and flex grow, if not it will prob have to be absolute positioned again.

Regarding the app idea - we could port some of the mobile components into a React-Native app, would prob be quicker than starting from scratch with pure iOS or Android (and would work on both). I've no experience with publishing react-native apps though.

gabek:
> Regarding the app idea - we could port some of the mobile components into a React-Native app, would prob be quicker than starting from scratch with pure iOS or Android (and would work on both). I've no experience with publishing react-native apps though.

I was just joking about an app, since I'm really frustrated and sad right now with working on the web.

But since you mentioned it, I do have experience releasing React Native apps and they're just absolutely the worst. It's all the worst parts of the web and none of the great things about native mobile development. It tries to do cross platform UI development, and that's the wrong thing to try and do. Every mobile environment is unique and should use it's native toolkit. Shared business logic makes a ton of sense though, just not UI.

gingervitis:
I'll try to address these issues that I've noticed:
- odd offline spacing issues (too much space at the top)
- mobile chat spacing issues (too much space at the top)
- odd footer issues (doesn't show up if window is tall enough)

gabek:
Closing since #3022 was merged in, and I'm filing smaller issues that can be looked at individually.

PR #3022, opened 2023-05-17
Address some layout issues with odd content spacing on mobile, and footer position

gingervitis:
Addresses some issues in https://github.com/owncast/owncast/issues/3012

- odd spacing issues on mobile
  - added css vars for top section (video container height + status bar height) to more consistently calculate the `top` positioning of the tabbed content div

- odd footer issues 
  - gave entire content `<section>` a min height of 100vh, so we can place the `<footer>` absolutely at the bottom. 
  - removed all the fixed positioning of footer
  - transferred some logic into `<Footer>` so that it wouldn't need to accept other props and can be reused. It'll still take `dynamicPadding` prop optionally. If prop exists, set its padding right to a calculated value of various vars.
  -  I ended up putting Footer in 2 areas. One on the main layout, only when `!mobile || !online`. Then put the other one under the About tab when mobile. The one in the tab is a bit awkward and could sometimes look like it floats in some cases. 
    - I'm not too thrilled about this solution to not make the footer show up when the chat panel is active. There wasn't a good way to detect when chat panel is on. Maybe we just dont show a footer at all when online? 

- created some consts for "320px" used in many areas, 

---

I feel like there's a bit of extraneous logic in places that could make style and layout logic a bit brittle. I didn't really address that logic too much, but here are some of the cases I had to test with.

 ####  mobile layout test cases 
  - **no tabs offline** - flow, Footer is part of Main
  - **no tabs online** - flow; but Footer is part of `Content_lowerSectionMobile` About
  - **with tabs offline** - pos relative; top: none, Footer is part of Main
  - **with tabs online** - top = video+status , pos abs; Footer is part of `Content_lowerSectionMobile` About
  - **with tabs online** - but chat hidden  = no tabs online case
 
At some point maybe there could be some consolidation somewhere.


gabek:
> I'm not too thrilled about this solution to not make the footer show up when the chat panel is active. 

For a while we weren't showing a footer for mobile at all, and I think that's perfectly fine. It's not necessary online or offline for mobile since you probably won't scroll far enough down to see it most of the time anyway on a phone screen.

gingervitis:
> > I'm not too thrilled about this solution to not make the footer show up when the chat panel is active.
> 
> For a while we weren't showing a footer for mobile at all, and I think that's perfectly fine. It's not necessary online or offline for mobile since you probably won't scroll far enough down to see it most of the time anyway on a phone screen.

Ok I'll just remove that footer instance to keep things simple.

gabek:
Just tested on iOS Safari. Looks like some spacing issues still exist.  image

---
Ignore. I was running old code.

gabek:
This is interesting. It seems to be a double scrolling issue where the bottom section can't scroll completely to the bottom, but then the entire page will, pushing up the video. When scrolling back down the video stays cut off unless you scroll again. 

I guess this kind of thing was my original hope of getting rid of fancy scrolling and letting the entire page scroll. But that requires removing chat from the tabs. 

This probably isn't the worst thing in the world, I would hope somebody would figure out they need to scroll twice to fix the page layout, but it's still annoying. Not sure what you think.

https://github.com/owncast/owncast/assets/414923/0cc80d55-0fe8-4b2b-bbc9-7ab346de6294

gingervitis:
Oh hm, yeah I 've noticed that too and thought it was already like that, and that it was annoying as well.
hmm, I'll take a quick look.

gingervitis:
Pushed a fix. Lemme know if that works better for you.

gabek:
I say we merge this in and get it deployed on more servers and configurations to test!

gabek:
After it hits `nightly` tonight I'll deploy it to the demo server.

PR #3023, opened 2023-05-17
fix : icon-text alignment issue

### LINKED ISSUE #3009: Different fonts make images align to text differently in chat

Notice the left people image and right star image are not vertically centered with the text with this font. These should be center aligned regardless of the font used.

image

dev265545:
fixes : #3009 Different fonts make images align to text differently in chat

### DIFF
--- a/web/components/chat/ChatJoinMessage/ChatJoinMessage.module.scss
+++ b/web/components/chat/ChatJoinMessage/ChatJoinMessage.module.scss
@@ -1,5 +1,8 @@
 .root {
   padding: 10px 10px;
+  display: flex;
+  align-items: center;
+  justify-items: center;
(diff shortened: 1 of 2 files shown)


gabek:
Looks like there's a spacing issue where the display name and the word "joined" don't have a space between them. [See this in Storybook](https://629132c6e23893003a9e89c5-qedzxqxiiu.chromatic.com/?path=%2Fdocs%2Fowncast-chat-messages-chat-join--regular).

image

gabek:
It still doesn't look centered.  For example, here's the same icon vertically centered with a display name in a chat message. See how the star is vertically centered with the text:

image


However in the join message, the star is not vertically centered with the text, and looks more baseline aligned:

image

PR #3038, opened 2023-05-22
Refactor mobile chat into modal

### LINKED ISSUE #3037: Rework mobile web layout

- Convert chat to be a modal, removing it from the scrolling requirements of the page.
- Make the mobile page completely scrollable, no longer forcing locking of scrolling to specific areas. This simplifies the tabs and content layouts.

gabek:
- Make page always entirely scrollable, even when live
- Move mobile chat into a standalone modal

https://github.com/owncast/owncast/assets/414923/ada11501-76e4-4ee6-842e-ae51e4b4d8f2

image

### DIFF
--- a/web/components/layouts/Main/Main.tsx
+++ b/web/components/layouts/Main/Main.tsx
@@ -171,11 +171,9 @@ export const Main: FC = () => {
           <FatalErrorStateModal title={fatalError.title} message={fatalError.message} />
         )}
 
-        {(!isMobile || !online) && (
-          <div className={styles.footerContainer}>
-            <Footer dynamicPaddingValue={dynamicFooterPadding} />
-          </div>
-        )}
+        <div className={styles.footerContainer}>
+          <Footer dynamicPaddingValue={dynamicFooterPadding} />
+        </div>
       </Layout>
       <Noscript />
     </>
--- a/web/components/modals/ChatModal/ChatModal.tsx
+++ b/web/components/modals/ChatModal/ChatModal.tsx
@@ -46,8 +46,6 @@ export const ChatModal: FC<ChatModalProps> = ({ messages, currentUser, handleClo
         title={<UserDropdown />}
         maskStyle={{
           zIndex: 700,
-          background:
-            'linear-gradient(180deg, rgba(0,0,0,0) 40%, rgba(0,0,0,0.45) 50%, rgba(0,0,0,0.5) 100%)',
         }}
         className={styles.root}
         bodyStyle={modalBodyStyle}
--- a/web/components/modals/ChatModal/ChatModal.tsx
+++ b/web/components/modals/ChatModal/ChatModal.tsx
@@ -1,0 +1,68 @@
+import { Modal } from 'antd';
+import { FC } from 'react';
+import { ErrorBoundary } from 'react-error-boundary';
+
+import styles from './ChatModal.module.scss';
+
+import { ComponentError } from '../../ui/ComponentError/ComponentError';
+import { ChatContainer } from '../../chat/ChatContainer/ChatContainer';
+import { ChatMessage } from '../../../interfaces/chat-message.model';
+import { CurrentUser } from '../../../interfaces/current-user';
+import { UserDropdown } from '../../common/UserDropdown/UserDropdown';
+
+export type ChatModalProps = {
+  messages: ChatMessage[];
+  currentUser: CurrentUser;
+  handleClose: () => void;
+};
+
+export const ChatModal: FC<ChatModalProps> = ({ messages, currentUser, handleClose }) => {
+  if (!currentUser) {
+    return null;
+  }
+  const { id, displayName, isModerator } = currentUser;
+
+  const modalBodyStyle = {
+    padding: '0px',
+    height: '55vh',
+  };
+
+  return (
+    <ErrorBoundary
+      // eslint-disable-next-line react/no-unstable-nested-components
+      fallbackRender={({ error, resetErrorBoundary }) => (
+        <ComponentError
+          componentName="ChatModal"
+          message={error.message}
+          retryFunction={resetErrorBoundary}
+        />
+      )}
+    >
+      <Modal
+        open
+        centered
+        maskClosable={false}
+        footer={null}
+        title={<UserDropdown />}
+        maskStyle={{
+          zIndex: 700,
+          background:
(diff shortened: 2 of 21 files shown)


gabek:
TODOs: 

- Add automated browser tests around chat modal
- Add Story around chat modal for UI tests

gingervitis:
Some observations:
- Did you want to be able to close the chat modal if you click outside of it?

- I wonder if the button should float above the footer?  OR, just move the button up.
image


- If you start off on small screen with chat, then expand to a larger screen, I think the modal should go away?
  image

gabek:
> Did you want to be able to close the chat modal if you click outside of it?

I specifically made it that you didn't because I was afraid people would accidentally close it. But that can be easily changed.

> If you start off on small screen with chat, then expand to a larger screen, I think the modal should go away?

I noticed that too, but I figured that was pretty rare so I ignored it. To close it on window size we'd need to tie into window size changes and then set the `showChatModal` to false. Sounds like more flakiness waiting to happen.

gingervitis:
- I wonder if the title bar should say "Chat" on the left, and have the menu on the right.  Having the menu on the left without context feels a bit weird? But then that doesn't leave room for the X button. 🤔 
image


- Tablet, at 768 portrait, Hide Chat is still there
image

- At 768 Hide chat also shows up in  the dropdown on the chat modal. If you hide it, the floating button does not appear.
image


- iOS mobile , icon gets cut off.
image

web/components/layouts/Main/Main.tsx
gingervitis, line 1: I think we'll need to revert the min-height back to `100vh` in  `Main.module.scss`. 
To see the footer you may need to do extra bit scrolling on iOS/android browser, but  I think that's ok.

Currently on Android tablet that does not support the `100dvh`.
image

web/components/ui/Content/MobileContent.tsx
```diff
@@ -111,20 +71,7 @@ export const MobileContent: FC<MobileContentProps> = ({
...
-          showChat={showChat}
-          chatEnabled={chatEnabled}
-          messages={messages}
-          currentUser={currentUser}
-        />
-      ),
-    });
-  }
+
   items.push({ label: 'About', key: '2', children: aboutTabContent });
```
gingervitis: should this be key:1 now?

web/components/ui/Content/Content.module.scss
```diff
@@ -109,3 +104,12 @@
 .offlineBanner {
   color: var(--theme-color-background-main);
 }
+
+.floatingMobileChatModalButton {
+  position: fixed;
+  width: 100px;
+  height: 40px;
+  bottom: 20px;
```
gingervitis: changing this to `bottom: 50px;` would be an easy workaround for footer overlap

web/components/modals/ChatModal/ChatModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Modal
+        open
+        centered
+        maskClosable={false}
+        footer={null}
+        title={<UserDropdown />}
+        maskStyle={{
+          zIndex: 700,
+          background:
+            'linear-gradient(180deg, rgba(0,0,0,0) 40%, rgba(0,0,0,0.45) 50%, rgba(0,0,0,0.5) 100%)',
```
gingervitis: If we don't want to be able to click outside the modal to close it, then I think we should still have the backdrop cover the entire background. Otherwise buttons on the top half of the page look like they're clickable. 

I know we do want to prioritize seeing the video while chatting, but at least you can still see it! ..and know that the other buttons around it might not be clickable.

So if we remove that line it might feel better?

web/components/ui/Content/Content.module.scss
```diff
@@ -13,20 +13,9 @@
   bottom: 0;
   width: 100%;
 
-	@include screen(tablet) {
-		top: 0;
+  @include screen(tablet) {
+    top: 0;
     position: relative;
-
-		&.online {
```
gingervitis, line 20: wait do we still need to do this for other tabs?

web/components/ui/Content/Content.module.scss
gabek: I increased the z-index of the button so it renders above the footer, so hopefully that works. But I'll bump the position up anyway.

web/components/layouts/Main/Main.tsx
gabek: Done!

web/components/ui/Content/Content.module.scss
gabek: `&.online` exists for tablet, and I didn't want to touch anything tablet related because it seems to be working. Is there something I should look at?

gabek:
> I wonder if the title bar should say "Chat" on the left, and have the menu on the right. Having the menu on the left without context feels a bit weird? But then that doesn't leave room for the X button.

I was wondering if the close button should be on the left and the menu on the right, but then there would be one modal that has a reverse close button and that feels more wrong than the menu being in a slight odd place. This can be iterated on if there's a better solution in the future.

gabek:
> At 768 Hide chat also shows up in the dropdown on the chat modal. If you hide it, the floating button does not appear.

What device is that?

---

Oh I see Phone + Landscape can do this.

gabek:
> iOS mobile , icon gets cut off.

I haven't seen that on any devices I've tested with, how do I reproduce that?

gingervitis:
> > iOS mobile , icon gets cut off.
> 
> I haven't seen that on any devices I've tested with, how do I reproduce that?

I see it happening on an iPhone 14 pro, as well as in the iPhone Simulator (for the same phone)

gabek:
Weird!  I'm on a 14 Pro as well. Hardware. It doesn't look like this for you?

gabek:
Oh crap you specifically said the ICON is cut off, not the button. Nevermind.

web/components/ui/Content/Content.tsx
```diff
@@ -305,6 +303,24 @@ export const Content: FC = () => {
...
+      )}
+      {isChatVisible && (
+        <Button
+          id="mobile-chat-button"
+          type="primary"
+          onClick={() => setShowChatModal(true)}
+          className={styles.floatingMobileChatModalButton}
+          style={{ zIndex: 99 }}
+        >
+          Chat <MessageFilled style={{ overflow: 'auto', position: 'relative', top: '2px' }} />
```
gingervitis: Not sure how `top: 2px` is working here with `relative`, but using transform translate is preferred. `transform: translateY(2px);`

gabek:
Let's see how this goes!

PR #3071, opened 2023-06-05
Alexsat/fix pix_fmt for multiple ffmpeg versions

AlexSat:


config/constants.go
```diff
@@ -5,8 +5,8 @@ import "path/filepath"
 const (
 	// StaticVersionNumber is the version of Owncast that is used when it's not overwritten via build-time settings.
 	StaticVersionNumber = "0.1.1" // Shown when you build from develop
-	// FfmpegSuggestedVersion is the version of ffmpeg we suggest.
-	FfmpegSuggestedVersion = "v4.1.5" // Requires the v
+	// FfmpegWithGeneralPixFmtVersion is the version of ffmpeg user for 'vaapi' fix_fmt value, instead of 'vaapi_vld'.
+	FfmpegWithGeneralPixFmtVersion = "v5.0" // Requires the v
```
gabek: I know we had that ffmpeg version in constants before. The thought was it would be changed over time, but that never really happened. The ffmpeg version pixel format version will never need to be changed, so maybe it should just live local to the single place it's being used.

gabek:
Seems to have two linter errors:

```
 Error: string `vaapi` has 3 occurrences, make it a constant (goconst)
  Error: commentFormatting: put a space between `//` and comment text (gocritic)
```

gabek:
And a test failure:

```
--- FAIL: TestOldFFmpegVaapiCommand (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
```

config/constants.go
AlexSat: I check it and found usage in several places: codecs and tests, so i refactored usage

gabek:
How come you're doing version comparison instead of relying on the actual list of pixel formats? ffmpeg version testing has not been very reliable in the past. Not all builds of ffmpeg report versions the same way and it resulted in not knowing the actual version for many ffmpeg builds.

AlexSat:
> How come you're doing version comparison instead of relying on the actual list of pixel formats? ffmpeg version testing has not been very reliable in the past. Not all builds of ffmpeg report versions the same way and it resulted in not knowing the actual version for many ffmpeg builds.

Oh, You are right, I should use list of pixels format. I fix it by next commits (for a few days, i think)

AlexSat:
@gabek fixed

core/transcoder/codecs.go
```diff
@@ -23,6 +29,16 @@ type Codec interface {
 	GetPresetForLevel(l int) string
 }
 
+// MockedCodec is mock for each codec to easy testing. Respect to https://github.com/anpavlov .
+type MockedCodec struct {
```
gabek: What's the mock wrapper for? Additionally it feels wrong to have test-specific code in application code.

core/transcoder/transcoder.go
```diff
@@ -447,8 +447,3 @@ func (t *Transcoder) SetIdentifier(output string) {
 func (t *Transcoder) SetInternalHTTPPort(port string) {
 	t.internalListenerPort = port
 }
-
-// SetCodec will set the codec to be used for the transocder.
-func (t *Transcoder) SetCodec(codecName string) {
```
gabek, line 452: Why was the setter removed? If anything I think we should standardize on setters instead of setting properties directly.

core/transcoder/codecs.go
```diff
@@ -31,6 +47,14 @@ var supportedCodecs = map[string]string{
 	(&VideoToolboxCodec{}).Name(): "videotoolbox",
 }
 
+var prefferedPixelFormats = map[string][]string{
```
gabek, line 40: Think we could have each of these formats that are used for each codec as a part of the implementation of the codec itself? Much like how you can ask a codec for its name or flags.

gabek (commented):
Awesome, thank you! I left a couple questions and thoughts.

core/transcoder/codecs.go
AlexSat: You are right, i should use earthfile instead host system to run tests. Fixed it

core/transcoder/transcoder.go
AlexSat: Yes, I agree.

core/transcoder/codecs.go
AlexSat: But I have not idea how to choose pixel format... Ffmpeg returns so many pixel formats and I dont know how to make best choice. Could you help me with it, please? Could you explain choose-pixel-fmt-way to me, please?

gabek:
Looks like tests are failing.

PR #3100, opened 2023-06-24
fix TitleNotifier crash

janWilejan:
rare chance of crashing when using ReactRefresh with an empty chat


gabek:
Great catch, thanks for fixing this up!

PR #3101, opened 2023-06-24
username of repeat chat messages is display:none

### LINKED ISSUE #3077: Add username for each chat message in html

Hi!
I have written a CSS theme for Owncast which I've seen a few people use over the months. I wanted to port it to the new Owncast 0.1.0 but the HTML for the chat is not suitable for it: Right now, the username is only included once for every "block" of messages. 

I'd like to be able to display the name of the author before each message. 

That's why I would request that the `.ChatUserMessage_user__` div be added to each `.ChatUserMessage_root`. 
If you can set it to `display:none` to hide it for subsequent messages (as to mimick the current look), it would enable us to set it to `display:block`if we want to.

Not sure if I'm expressing this right, feel free to ask for clarification!

janWilejan:
users who want to show repeat usernames can inject the following CSS: `[class^="ChatUserMessage_repeatUser"]{display:flex !important;}`

closes #3077

### DIFF
--- a/web/components/chat/ChatUserMessage/ChatUserMessage.module.scss
+++ b/web/components/chat/ChatUserMessage/ChatUserMessage.module.scss
@@ -20,6 +20,11 @@ $p-v-size: 2px;
     font-weight: 600;
   }
 
+  .repeatUser {
+    @extend .user;
+    display: none;
+  }
+
   .userBadges {
     margin-left: 3px;
   }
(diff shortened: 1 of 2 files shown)


gabek:
Would you mind commenting on the original issue so it can be assigned to you? Thanks!

gabek:
Would you mind commenting on the original issue so it can be assigned to you? Thanks!

gabek:
Nice, thank you!

PR #3113, opened 2023-06-27
change chat from a sidebar to a column

janWilejan:
Using a 2-column layout prevents the chat scrollbar from overlapping the page scrollbar. Also, it no longer needs to calculate extra padding for elements.


gabek:
For future reference, it would be useful if you could file an issue before putting up a PR so changes can be discussed before they're worked on. This is a rather large change that will require a lot of testing. There was a lot of issues with layout that took a long time to get resolved previously, so I'm hesitant to throw in changes now, after it's been working.

It might be a while before I can focus on testing this, so I hope that's ok.

janWilejan:
Yeah, sorry.

When I noticed the overlapping scrollbars I thought that fixing it would be simple, so I just jumped in.



Test this at your own pace; there's no rush :+1:

gabek:
One piece of helpful testing, if you look through the checks you can look at the UI review to see how the changes impact things. This most recent change is at: https://www.chromatic.com/review?appId=629132c6e23893003a9e89c5&number=3113&type=linked&view=changes

janWilejan:
(there were merge conflicts so i rebased this branch)

I've fixed the remaining differences i see in Chromatic. the differences Cromatic is still showing are due to the way it captures the screenshot and are not visible in browser:

Footer: Example -- The footer now uses flexbox rather than position:absolute. The footer has the same size and position on the page as before, but Chromatic adds a margin to one and not the other.

Main: Offline Mobile & Main: Offline Tablet -- Cromatic sees some spurious differences in font aliasing, i guess?

Main: Online & Main: Online Mobile & Main: Online Tablet -- The page is slightly taller than the viewport. In baseline, that means the screenshot includes a few pixels below the page; in this branch, extra height goes into the mainColumn, which has its own scroll box.

gabek:
Thank you for all your hard work on the improvements! Functionally it all looks to work as expected. 

Previously we had hid the scrollbars completely to make it one less thing to worry about, it seems like your PR brings them back.  
image

There's also a new scrollbar visible under the footer:
image

janWilejan:
Oops! Those scrollbars weren't visible in Firefox or the Chromatic preview. I guess FF and Chromium handle scrollbars differently, but I'd expect to see them in Chromatic's screenshots

gabek:
It all looks great! Removing the dynamic padding is a huge improvement. Let's get this in and I'll deploy it tonight to servers running nightly to get some real world testing. Thank you!

PR #3119, opened 2023-06-28
FIX Remove animation of chat sidebar #3089

### LINKED ISSUE #3089: Remove animation of chat sidebar

The Ant Design sidebar component is animating in when the page first loads and chat is active. This shows a flash of empty space and a broken feeling. We should completely remove this transition so it doesn't slowly move into place.

In my initial attempts I have not figured out how to remove this.

1. Start stream
2. Load page
3. Watch sidebar slide in from the right

dev265545:
FIXES Remove animation of chat sidebar #3089 
hi, 
just put the same width of loading-sider  with the chat one.

### DIFF
--- a/web/components/ui/Sidebar/Sidebar.tsx
+++ b/web/components/ui/Sidebar/Sidebar.tsx
@@ -27,7 +27,7 @@ export const Sidebar: FC = () => {
 
   if (!currentUser) {
     return (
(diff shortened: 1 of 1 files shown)


gabek:
Fantastic! Thanks for tracking it down, Dev!

Issue #1803, opened 2022-03-29
Chat - Emote Wire Protocol


jprjr:
Hi there! I noticed when emotes are sent, the HTML for an `<img>` tag is sent via websocket.

Thought I'd suggest sending the code/title (`:partyparrot:`) instead and have it updated/rendered server-side? As far as I can tell (and I'm just getting started, so apologies in advance if I'm mistaken) - for a bot to use a custom emote, it needs to also generate and send the HTML.

Being able to send the tag would simplify bot code a bit. Plus, for webhooks - `rawBody` could also just contain the tag, instead of the HTML. Would make it easier for bots to do things like, count how often emotes are used - right now they have to parse the HTML and extract the alt or title attribute.

gabek:
I think that's fair, I could see how that would be helpful.

gabek:
Since this will require frontend changes I'll wrap this up with the future frontend changes. It probably won't be soon, but it makes sense to hold off until then since it'll all have to be touched anyway.

jprjr:
Hi there, now that the new UI is out, I wanted to volunteer to take a crack at implementing this. Just wanted to check-in and ensure 1) it's still desired and 2) nobody else has already started.

Here's my rough idea:

* change front-end to send emoji as text like `:partyparrot:` rather than HTML.
* sever-side, validate sent emoji and convert to HTML.
* keep plain-text version in `rawBody` and HTML-rendered version in `body` attributes of message object.

gabek:
This sounds great, thank you for offering to work on it! Since the output will still end up being HTML it won't impact anybody using chat webhooks or the rendering of chat messages, so there's no downside here.

jprjr:
Took a look and there's one behavior change I'd like to propose - basically for the emoji picker to just always put emoji at the end of the chatbox input, rather than where the caret position is. This is similar to how twitch chat works - it ignores your caret position.

There's a fundamental mis-match, at least in Firefox, between the caret position and the innerHTML content.

Right now, the caret position is being saved by taking the length of the innerHTML up to the caret. So if you have an emoji you're taking the length of `<img src=".... alt="emote">` - etc, and coming up with a high value for the caret position, say 30.

As far as the browser is concerned, if you have a single custom emote, the caret position is really just `1`. So when you go to re-set the caret position it's some invalid value.

[...]

PR #3124, opened 2023-06-30
chat: simplify input handling

### LINKED ISSUE #3121: ChatBox Error -  SelectAll + Delete does not work if input is text + emoji

ChatBox Error -  SelectAll + Delete does not work if input is text + emoji
If we have an emoji and some text or all emojis  then doing select all and backspace/delete does not work.
Only one char/emoji gets deleted per click.
Works fine if input is only text

jprjr:
Removes the cursor save/restore functionality.

Removes most key handling.

Allows message to go over limit.

Moves the message length check into SendMessage.

Changes the chat input to change to the max style only when over the limit, rather than at the limit. Makes it apparent that something is wrong.

Fixes #3121

One thing I wasn't sure about in the change handler is this line: 
https://github.com/owncast/owncast/blob/99aabc64b0a107cdbf5e6c0743dd04dc103541d4/web/components/chat/ChatTextField/ChatTextField.tsx#L172C5-L172C70

I suspect the regex is removing some kind of tag, but then there seems to be a disconnect - now the character count doesn't match what's sent (since SendMessage doesn't apply that same regex). I figure just keeping the character count at or above what's actually sent makes things a bit easier to understand.

An alternative may be changing getCharacterCount to also discard count the opening/closing `<p>` tags since sendMessage discards them before sending. That way you get a true "this is the number of characters that will send" number.

One issue I'm still seeing is if I select all, eventually the selection gets cleared. But it's not immediate like it was before.

### DIFF
--- a/web/components/chat/ChatTextField/ChatTextField.tsx
+++ b/web/components/chat/ChatTextField/ChatTextField.tsx
@@ -33,44 +33,10 @@ export type ChatTextFieldProps = {
 
 const characterLimit = 300;
 
-function getCaretPosition(node) {
-  const selection = window.getSelection();
-
-  if (selection.rangeCount === 0) {
-    return 0;
-  }
-
-  const range = selection.getRangeAt(0);
-  const preCaretRange = range.cloneRange();
-  const tempElement = document.createElement('div');
-
-  preCaretRange.selectNodeContents(node);
-  preCaretRange.setEnd(range.endContainer, range.endOffset);
-  tempElement.appendChild(preCaretRange.cloneContents());
-
-  return tempElement.innerHTML.length;
-}
-
-function setCaretPosition(editableDiv, position) {
-  try {
-    const range = document.createRange();
-    const sel = window.getSelection();
-    range.selectNode(editableDiv);
-    range.setStart(editableDiv.childNodes[0], position);
-    range.collapse(true);
-
-    sel.removeAllRanges();
-    sel.addRange(range);
-  } catch (e) {
-    console.debug(e);
-  }
-}
-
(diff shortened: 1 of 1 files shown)


gabek:
> I suspect the regex is removing some kind of tag

Yeah, since emoji are img tags I figured it simplified things just to remove the tags and not count them as anything at all and only count non-html in the character count. Alternatively we could replace all html tags with an `x` or something to count them as one character. Since we're not displaying the actual character count I'm not currently too worried about accuracy in that regard.

web/components/chat/ChatTextField/ChatTextField.tsx
```diff
@@ -167,31 +107,10 @@ export const ChatTextField: FC<ChatTextFieldProps> = ({ defaultText, enabled, fo
         h3: 'p',
       },
     });
-    text.current = sanitized;
-
-    const charCountString = sanitized.replace(/<\/?[^>]+(>|$)/g, '');
```
gabek, line 172: Without this it hits the character limit at four custom emoji, because it's counting the characters in the HTML tags.

jprjr: My thought process was in #1803 I'd be converting the custom emoji into their shortcodes (and the shortcode length is then counted in the message length), I can try changing this PR to count emoji similarly. It wouldn't result in emoji counting as a single character but it's still way less characters than the entire tag.

gabek: Oh yeah, if you're planning on changing in it with that using shortcodes then no problem!

gabek:
Simpler is better, thanks!

PR #3125, opened 2023-07-01
Chat wire protocol

### LINKED ISSUE #1803: Chat - Emote Wire Protocol

jprjr:
This PR changes how emoji are sent.

Currently, the chat client sends emoji as images. This extends to third-party integrations/webhooks - say if a bot wants to send an emoji, it has to embed the HTML.

Now, the chat client send the text shortcode (ie, `:some-emoji:`), the server validates that it's an emoji and renders it to HTML. This also works for the access-token based chat messages. It does not work for system messages, I think it's just an issue of where to insert the injectEmoji call.

A few things to note:

* assumption is that all emoji have unique names. This isn't currently enforced in `core/data/emoji.go` as far as I can tell, but in the renderer I create a `map[string]string` to map emoji names to HTML snippets, which means any duplicates get overwritten. Unsure if `core/data/emoji.go` should be changed to use a `map[string]models.CustomEmoji` as well and warn on duplicates?
* web client also updated to count native emoji as a single-character each by using [grapheme-splitter](https://github.com/orling/grapheme-splitter).
* for custom emoji, the length is the shortcode length, I'd probably have to come up with some kind of custom length method.

Closes #1803


jprjr:
Going to take a look at implementing a goldmark extension (or possibly just use the goldmark emoji extension), that would likely allow custom emotes to render anywhere and still respect rules of say, code blocks.

core/chat/events/events.go
```diff
@@ -5,6 +5,8 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"sync"
+	"text/template"
```
gabek, line 8: This error just means you should run the go formatter to to re-order the imports. I've never understood why it just doesn't say it's the order that's the problem.

gabek:
I'm sure you've thought about this more than I have, but I'm just curious. I was thinking at first the approach might be to keep raw version (with the shortcode) and then render that as a HTML version with the img tags. But you did the reverse, converting the HTML string into a raw/shortcoded string. Seems like it's more effort to parse the HTML than it would be to convert the shortcodes into img tags. The result is the same, but I'm just curious what you ran into.

jprjr:
You mean for the chat input field? I guess there's two things -

I'm not really parsing html, I'm walking the DOM. The browser has to parse the HTML anyway, right?

But also - the main benefit of this approach is somebody could paste some pre-formatted text into the field and this keeps some of the formatting. Like if I copy a snippet from some web page with bold or italic text, what I send is what I get back.

Basically, the contents of the editable div is really HTML (or even more accurately, the DOM), so I figured it makes the most sense to walk that and come up with a markdown representation.

gabek:
Let me know if you have any questions fixing up those CI issues.

jprjr:
Will do! Had to put this on pause the past day but I should be able to get back on this

jprjr:
Pushed up some changes to integrate the emoji parsing/rendering into the Markdown renderer using [goldmark-emoji](https://github.com/yuin/goldmark-emoji).

It adds another assumption to emoji - in addition to assuming they all have unique names, this also assumes the all names are unique in a case-insensitive manner. I'm unsure if anybody's running owncast on systems with a case-insensitive filesystem, but I figure most people want their emoji names to be unique without regards to capitalization. And if a user say, manually types out a shortcode, they'd probably be frustrated to not see their emote because of wrong captialization.

Since it's now part of the markdown renderer and not being injected as raw HTML via regex, things like system messages will support the shortcode format, and they'll be ignored in cases where they should be ignored like in code blocks.

So now, a user can send:

* `:ablobattention:`
* `:ABLOBATTENTION:`
* etc

and get the same emote.

Another change is a few more mutex locks - I figure while rendering the markdown you want to ensure nothing else can change the emoji data, so there's a lock added there. Also, `GetEmojiList()` returns a copy of the data, I'm not 100% familiar how go works with regard to returning slices and where the actual, underlying data is stored. I figure making a copy and returning that is safer - again, in case another cache update occurs while say, the markdown renderer is trying to update its own list.

Currently waiting on latest CI checks to pass, I think I got them all figured out.

jprjr:
Oh and one other assumption - when I say "unique names" I'm referring to the filename without extension.

So all together, if a user uploaded emotes named say:

* `MyEmote.webp`
* `myemote.webp`
* `myemote.gif`

This may cause some confusion in the current state - since the emoji data used by the picker is basically an array of all walked files, I think all 3 emotes would show there. But the server-side renderer would only use 1 of them, basically whatever the last emote in the list is. Unsure if the code in `core/data/emoji.go` should be updated to also enforce that? I doubt that there's a lot of owncast users out there uploading emotes with conflicting names - but maybe logging a warning or something would be useful.

gabek:
Maybe we could just add an extra blurb on the emoji upload page that specifies that names should be unique for best results. That's probably where that information is most useful.

jprjr:
Done - added some text to the admin page about how to get best results.

gabek:
Looks like there's a conflict with `package-lock.json`. If you rebase off `develop`, delete `package-lock.json`, and `npm install` it should regenerate.

jprjr:
Rebased and regenerated the package-lock, pushed

core/chat/events/events.go
```diff
@@ -66,6 +73,105 @@ func (e *UserMessageEvent) SetDefaults() {
...
+	emojiDefs         = newEmojis()
+	emojiHTML         = make(map[string]string)
+	emojiModTime      time.Time
+	emojiHTMLFormat   = `<img src="{{ .URL }}" class="emoji" alt=":{{ .Name }}:" title=":{{ .Name }}:">`
+	emojiHTMLTemplate = template.Must(template.New("emojiHTML").Parse(emojiHTMLFormat))
+)
+
+func loadEmoji() {
+	modTime, err := data.UpdateEmojiList(false)
+	if err != nil {
```
gabek, line 146: This probably wasn't handled great before, but if `os.Stat` returns an error it won't get to the error log in `UpdateEmojiList` and will silently fail, so maybe it should be logged here.

gabek:
I pasted the text from https://loremipsum.io/ into the chat box that starts with some italic. It renders as expected in the textbox, but it gets stripped when sending it:

image

image

However, when pasting example text from https://www.w3schools.com/html/html_formatting.asp the italic worked as expected. 

image

I'm personally not overly concerned about keeping the rich text formatting, but I know one of your goals was to support that. I guess the biggest issue is the italic text that was completely stripped in the first example.


<blockquote><div> Lorem Ipsum</div><div><strong><a href="https://loremipsum.io/">Lorem Ipsum – Generator, Origins and Meaning</a></strong></div><div>Generate Lorem Ipsum placeholder text for use in your graphic, print and web layouts, and discover plugins for your favorite writing, design and blogging tools. Explore the origins, history and meaning of the famous passage, and learn how Lorem Ipsum went from scrambled Latin passage to ubiqitous dummy text.</div></blockquote>
<blockquote><div><strong><a href="https://www.w3schools.com/html/html_formatting.asp">HTML Text Formatting</a></strong></div><div>W3Schools offers free online tutorials, references and exercises in all the major languages of the web. Covering popular subjects like HTML, CSS, JavaScript, Python, SQL, Java, and many, many more.</div></blockquote>

<blockquote><div> Lorem Ipsum</div><div><strong><a href="https://loremipsum.io/">Lorem Ipsum – Generator, Origins and Meaning</a></strong></div><div>Generate Lorem Ipsum placeholder text for use in your graphic, print and web layouts, and discover plugins for your favorite writing, design and blogging tools. Explore the origins, history and meaning of the famous passage, and learn how Lorem Ipsum went from scrambled Latin passage to ubiqitous dummy text.</div></blockquote>
<blockquote><div><strong><a href="https://www.w3schools.com/html/html_formatting.asp">HTML Text Formatting</a></strong></div><div>W3Schools offers free online tutorials, references and exercises in all the major languages of the web. Covering popular subjects like HTML, CSS, JavaScript, Python, SQL, Java, and many, many more.</div></blockquote>

jprjr:
oh good catch - I managed to use `emph` instead of `em` as a tag name  :facepalm:

Corrected the tagname, and also added a fallback for the default type to return the `.textContent` field for any un-accounted for tags.

gabek:
All is looking great! Thank you so much for the numerous improvements that this change includes. Let's get it in.

PR #3137, opened 2023-07-05
highlighting: make case-insensitive, support unicode

jprjr:
Small change to make highlighting case-insensitive and to be unicode-aware for say, users with emoji or accented characters in their display name. Also allows matching any whitespace - I found it was pretty easy to say, accidentally insert a non-breaking-space when copying an emoji to use in my username. (alternatively it may be worth looking into normalizing usernames such that non-breaking-spaces are collapsed into a single regular space, I'll check that out, too).

Relies on the browser supprting unicode-aware RegExp, which is supported in all current browser versions: https://caniuse.com/mdn-javascript_builtins_regexp_unicode

### DIFF
--- a/web/components/chat/ChatUserMessage/customMatcher.ts
+++ b/web/components/chat/ChatUserMessage/customMatcher.ts
@@ -19,9 +19,18 @@ export class ChatMessageHighlightMatcher extends Matcher {
       return null;
     }
 
-    const highlightRegex = new RegExp(highlightString.replace(/\s/g, '\\s'), 'ui');
+    const escapedString = highlightString
+      .replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&')
+      .replace(/\s/g, '\\s');
 
-    const result = str.match(highlightRegex);
+    const normalizedString = escapedString.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
+
+    let highlightRegex = escapedString;
+    if (escapedString !== normalizedString) {
+      highlightRegex = `(?:${escapedString})|(?:${normalizedString})`;
+    }
+
+    const result = str.match(new RegExp(highlightRegex, 'ui'));
 
     if (!result) {
       return null;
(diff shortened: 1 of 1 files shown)


jprjr:
Pushed up another commit that allows matching with unaccented forms of usernames. This allows for things like, if I had the username "Pokémon Master" - a user could type "pokemon master" and the message would still get the highlight.

Also escapes RegExp special characters. Previously, if I made my username something like "john|away", the highlight would match both "john" and "away".

Admittedly, that's a case where it's not so bad if both match, but it seems like the safe thing to do is ensure any username characters aren't interpreted as regex.

gabek:
This is a great addition, thank you! It does look like there may be some an issue with the escaping, however, as I saw this error:

https://www.chromatic.com/pullrequest?appId=629132c6e23893003a9e89c5&number=3137&activeElementId=comparison-test-key-631d62bfa0c4f0c1eb500a7b-1200&view=changes

jprjr:
Oh wow, and like all the default usernames use dashes!

I replaced the escape pattern with one recommended on MDN - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping - If we're already escaping `[` and `]` which prevents forming character classes, there's not really a need to escape the dash.

<blockquote><div><strong><a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions">Regular expressions - JavaScript | MDN</a></strong></div><div>Regular expressions are patterns used to match character combinations in strings.
  In JavaScript, regular expressions are also objects. These patterns are used with the exec() and test() methods of RegExp, and with the match(), matchAll(), replace(), replaceAll(), search(), and split() methods of String.
  This chapter describes JavaScript regular expressions.</div></blockquote>

gabek:
Awesome, thank you!

jprjr:
A note to maybe explore in the future:

Changing the normalization form from `NFD` to `NFKD` will also make changes like, convert fullwidth text to the halfwidth version, bubble letters to regular letters, etc, and would allow those to match as well.

PR #3146, opened 2023-07-06
simpler chatbox

### LINKED ISSUE #3131: Chat: current edit messes with the caret (and more)

Hi there, so while implementing chat-related updates, I've noticed a few things about react-contenteditable:

* It seems to override when it should render and does so at strange times
* it messes with the caret position
* it gets updated by overriding innerHTML
* it doesn't seem to fire change events when expected

For example, setting the current character count for the message length is somewhat strange. What would make the most sense to me is to have an onchange callback, and make that the sole place to recalculate the length of the current message. So when the emoji picker picks an emoji - append it and let the onchange handler update the count.

But - it seems like the onchange handler is delayed, so the character count is instead updated in the emoji picker callback. So now there's a few places where the character count is updated.

I'm not a react expert, but my gut says a more ideal implementation would do less blocking of rendering, and maybe avoid overwriting innerHTML and instead use DOM-based methods to append images and emoji. I'd imagine that's more efficient since you're no longer parsing the html, rendering to virtual DOM, then updating the real DOM, but instead updating the virtual DOM and (probably) applying changes to the real DOM immediately

This issue is to just explore the idea replacing the dependency, either with another module or writing a custom one. I think it may be possible to write a simpler implementation that allows the browser to behave more naturally.

jprjr:
Removes 'react-contenteditable' and adds a new contenteditable that generally tries to leave things alone.

See #3131

### DIFF
--- a/web/components/chat/ChatTextField/ChatTextField.tsx
+++ b/web/components/chat/ChatTextField/ChatTextField.tsx
@@ -1,12 +1,12 @@
 import { Popover } from 'antd';
-import React, { FC, useEffect, useReducer, useRef, useState } from 'react';
+import React, { FC, useEffect, useState } from 'react';
 import { useRecoilValue } from 'recoil';
-import ContentEditable from 'react-contenteditable';
 import sanitizeHtml from 'sanitize-html';
 import GraphemeSplitter from 'grapheme-splitter';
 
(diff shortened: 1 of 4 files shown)


gabek:
This is awesome! It's certainly cleaner and simpler and works great. Thank you!

Issue #1883, opened 2022-04-26
Request: large emotes


jprjr:
Hi there - I want to stress that this feature request is very silly.

So, users on my Owncast instance discovered they can edit the HTML elements of emotes before sending - and by removing the `emote` class, the emote is no longer limited by width. A lot of our source emote images are something like 3x or 4x the displayed size, so the result is they send a large emote.

It's actually really fun, so I figured I'd request it as a formal feature - some method of doubling the width, maybe. I figure if the on-the-wire chat protocol is changed to be Not HTML (see #1803) this could go away.

PR #3147, opened 2023-07-06
emoji: display native emoji simiarly to custom

jprjr:
Hi there - this is a PR to allow displaying native/system emoji similarly to custom emotes by increasing their font size.

It matches emoji in user messages using a regular expression and inserts them into `<span>` elements with the `emoji` class, which sets the font-size and line-height to 30px. It also sets the display attribute to `inline-block`, which shouldn't affect images with the `emoji` class (images are technically `inline` but have special treatment and really behave more like `inline-block` anyway). The other change is setting the vertical-alignment which matches the image behavior.

The regular expression relies on a pretty new feature [that's still rolling out](https://caniuse.com/mdn-javascript_builtins_regexp_unicodesets), so this adds a dependency on [regexpu-core](https://github.com/mathiasbynens/regexpu-core). Matching on unicodeSets is required to match things like, complicated emoji made up of multiple codepoints/zero-width joiners/etc. At runtime, the unicodeSets feature is checked and if needed, `regexpu-core` will rewrite the regex to one that's compatible - otherwise it just uses the native support.

Here's a screenshot showing a native and custom emote side-by-side.

Screenshot_2023-07-06_15-29-47

### DIFF
--- a/web/components/chat/ChatUserMessage/ChatUserMessage.tsx
+++ b/web/components/chat/ChatUserMessage/ChatUserMessage.tsx
@@ -6,6 +6,7 @@ import dynamic from 'next/dynamic';
 import { Interweave } from 'interweave';
 import { UrlMatcher } from 'interweave-autolink';
 import { ChatMessageHighlightMatcher } from './customMatcher';
+import { ChatMessageEmojiMatcher } from './emojiMatcher';
 import styles from './ChatUserMessage.module.scss';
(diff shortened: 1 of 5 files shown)


gabek:
Cool! That's a nice addition, something I never thought much of before. Seems like we're inching closer to supporting your "Big emotes" feature :) #1883

PR #3155, opened 2023-07-09
Increase width of volume slider to 100px

janWilejan:
The 41px volume slider changed volume in 2.43% increments, making certain volume levels such as 1% impossible. A volume slider that's exactly 100px wide makes exact 1% increments possible.


gabek:
Sure, no downsides to this. Works for me!

PR #3157, opened 2023-07-10
Add resize handle to chat

### LINKED ISSUE #1918: Web UI v2: Adjustable chat width
Sometimes I wish chat had more real estate, but sometimes I don't. It would be really cool if we could drag the width of the chat column to adjust the layout dynamically.

I haven't looked into what it would take to pull this off. It can get pushed out if it's too much for this rewrite, but it'd be a nice touch.

janWilejan:
closes #1918


gabek:
This is so cool! And far less had to be done to pull it off than I expected.

PR #3160, opened 2023-07-10
replace grapheme-splitter with graphemer (updated fork)

jprjr:
While doing some testing I discovered grapheme-splitter would not correctly split out some grapheme clusters. Turns out it hasn't been updated in a while, it looks like [graphemer](https://github.com/flmnt/graphemer) is the preferred fork.

It provides a drop-in compatible module, this should provide more accurate character counts in the chat.


gabek:
Thanks for catching all of these chat items!

jprjr:
> Thanks for catching all of these chat items!

I mean technically I'm the one that created this issue by using grapheme-splitter in the first place, haha.

But no problem, I'm happy to help!

PR #3161, opened 2023-07-10
contenteditable: remove warning about unknown property onContentChange

jprjr:
Noticed in the Javascript console there was a warning about onContentChange not being a react property, which is true - it just gets stored in props, and isn't needed by React to render.

So this just removes it from the call to create the div and keeps it in this.props.


gabek:
I didn't even notice that, thanks!

PR #3162, opened 2023-07-10
chat: check that a new username is different before sending out updates

jprjr:
Just prevents a case where a user say, updates their current name by adding a bunch of spaces at the end, so the resulting new name is the same as the old, and the "user updated their name" goes out with the same name as before.


gabek:
Thank you for thinking about these things!

PR #3164, opened 2023-07-11
Name change: better unicode handling

### LINKED ISSUE #3159: DisplayName Unicode Handling
## Current situation

Using unicode in the displayname is a bit broken. In my explanations I'm using emoji, just because I know they're Unicode characters that can take up several codepoints, but this all applies to Unicode text characters as well.

The input box enforces a maximum character length, and I'll be honest, I'm not sure how that's being measured, it's not really accurate when you get into more complex emoji clusters like 🫱🏿‍🫲🏻 - that takes up 9 UTF-16 code points, and 19 bytes in utf-8. The name change box reports the length as 5. And doing things like pasting past the character limit causes the emoji to be broken up.

Assuming the dialogue box determines the string is an acceptable length - the server-side processing caps the string based on the number of bytes, and can then send back broken Unicode if it removes the final bytes of a UTF-8 sequence.

## Proposal

I did some quick prototyping and would like to propose a few changes:

Client-side:

* remove the maxCount property from the name change field to prevent breaking/splitting input.
* replace the "showCount" with a custom implementation that counts grapheme clusters (so things like "🫱🏿‍🫲🏻" count as 1).
* allow username input to go over the max characters but disable sending once over, similar to the current chat box.

Server-side:

* Update MakeSafeStringOfLength to count the length using [uniseg](https://github.com/rivo/uniseg), and remove whole grapheme clusters rather than bytes, which ensures no broken Unicode is sent back.

If this all sounds good, assign me the issue and I'll get everything cleaned up and get a PR going!

jprjr:
Client-side:

* Changes the NameChangeModal to show text "Over limit" when a proposed display name is too long.

* Allows names to go over limit to prevent splitting graphemes on input.

Server-side:

* Changes the MakeSafeStringOfLength to count number of unicode code points instead of string bytes.

Closes #3159

### DIFF
--- a/utils/strings.go
+++ b/utils/strings.go
@@ -18,10 +18,14 @@ func MakeSafeStringOfLength(s string, length int) string {
 	newString := s
 	newString = StripHTML(newString)
 
-	if len(newString) > length {
-		newString = newString[:length]
+	// Convert utf-8 string into Unicode code points.
+	codePoints := []rune(newString)
+
+	if len(codePoints) > length {
+		codePoints = codePoints[:length]
 	}
 
+	newString = string(codePoints)
 	newString = strings.ReplaceAll(newString, "\r", "")
 	newString = strings.TrimSpace(newString)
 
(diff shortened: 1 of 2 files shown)


gabek:
Nice! :100:

PR #3165, opened 2023-07-11
replace margins with padding in the chat container

janWilejan:
Virtuoso can't calculate the size of elements that have margins. This causes strange behaviour with scrolling in chat.


gabek:
Oh awesome! Maybe this is the root issue behind the flaky chat scroll to bottom behaviour. Thank you for discovering this!

PR #3167, opened 2023-07-11
change overflow behaviour in chat input

janWilejan:
image

Firefox only breaks at word boundaries by default, meaning we need a horizontal scrollbar to handle long words like AAAAAAAAAAAAAAAAAAAAAAAAAAAAA.


gabek:
Makes sense!

Issue #1138, opened 2021-06-24
Custom target framerates


bsncubed:
Hi 

Is it possible for you to add 25fps and 50 fps? In v0.05 i could manually edit the config file to make it work, but i can't find a work around now.

gabek:
I haven't seen others ask for non-standard framerates, but I'll keep this open for a bit to let others chime in.  Adding this would add complexity to the configuration and I'm not sure there's many people who would utilize it.

YohannLYD:
I too would be interested in having 25/50fps as an option.
This is actually a broadcast standard framerate in the European broadcast sector and haven't been able to use the software in production for this reason.

gabek:
If all we want to do is add it to our current selector control, we can do that.  I fear it'll start confusing people with too many options and that we would need to change to a completely different way to select framerates.  But if somebody wants to add this to the control we can give it a spin.

YohannLYD:
If too many framerates is confusing to people (is it really?) I'd be down to replace 24fps with 25fps to keep the number of options to the minimum.
After all 24fps is not really a video standard, but a cinema/film one.

bsncubed:
Would you be open to adding 50fps as well? just to keep the feature set on par with the US framerates (30/60)

gabek:
>  I'd be down to replace 24fps with 25fps to keep the number of options to the minimum.

I'm sure you would be ok with it, you're the one who asked for it :)

The concern I have is things are kind of setup to be "low, medium, high, stupid high".  So adding more is like saying "low, a different low, medium, another medium"... and that doesn't make a lot of sense to people trying to make a decision.  Like why would we have 24 _and_ 25?  I know it makes sense to you, but I can see that being kind of weird to normal people.

Just for my own knowledge, why does "broadcast standard" matter when it's the internet?  If you were to downsample 25fps to 24fps, what is the major drawback there?  I understand if this was broadcast television where standards need to be adhered to or the television won't play back a signal, but internet video is certainly more flexible.

gabek:
But again, if somebody wants to add these FPS options let's give it a spin.  If nobody does it then I'll eventually get around do it and if it's confusing or the control just doesn't work with it then it'll have to be readdressed.

YohannLYD:
[...]

PR #3179, opened 2023-07-16
Custom framerate in admin UI

### LINKED ISSUE #3178: Custom Framerate

Is there a way to add a custom framerate other than the ones offered? I would like to add 48fps. but I can't find the option anywhere.

janWilejan:
closes #3178

### DIFF
--- a/web/components/admin/VideoVariantForm.tsx
+++ b/web/components/admin/VideoVariantForm.tsx
@@ -15,7 +15,7 @@ import {
 import classNames from 'classnames';
 import dynamic from 'next/dynamic';
 import { FieldUpdaterFunc, VideoVariant, UpdateArgs } from '../../types/config-section';
-import { TextField } from './TextField';
+import { TextField, TEXTFIELD_TYPE_NUMBER } from './TextField';
 import {
   DEFAULT_VARIANT_STATE,
   VIDEO_VARIANT_SETTING_DEFAULTS,
@@ -133,7 +133,7 @@ export const VideoVariantForm: FC<VideoVariantFormProps> = ({
         </span>
       );
     if (dataState.framerate > 60)
-      return <span>High framerates will use a lot of bandwidth and CPU. {}</span>;
+      return <span>High framerates use more bandwidth and CPU. {}</span>;
     return undefined;
   }
   const cpuUsageNote = () => {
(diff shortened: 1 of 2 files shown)


gabek:
I don't think we should have this. It's very rare it's required and I believe more people will try to take advantage of it and cause problems for themselves than be useful for people.

If in the future enough people really need free form inputs for video settings we can discuss the best way to do that.

janWilejan:
How is changing the frame rate a rare issue? This has come up before ( #1138 ) and the solution then was to add more framerates to the list in the admin UI. 

Streamers should be able to pick any standard broadcast framerate in the UI without needing to make API calls: https://en.wikipedia.org/wiki/List_of_broadcast_video_formats

The UI should
* Encourage selecting a framerate from the list of reasonable defaults
* Warn when using excessive frame rates
* Allow setting any framerate (within a reasonable range)

In what ways do you think people will take advantage of it? Could this be fixed with more / sterner warnings?
image


<blockquote><div><strong><a href="https://en.wikipedia.org/wiki/List_of_broadcast_video_formats">List of broadcast video formats - Wikipedia</a></strong></div></blockquote>

gabek:
Oh I'm not saying changing the framerate is a rare issue, but changing the framrate to a non-standard, custom value is a rare issue. I do think we should have the values for any commonly used framerates, and if a handful of people said 48fps was something they wanted maybe we could add it specifically. But in general I try very hard to not have free form text fields when it comes to values that can have negative impact or allow for typos or bad judgement on the user side.

I tend to use the 80/20 rule with things, though I'm not strict with it. If 80% or more of people would find it useful then it's worth discussing the proper way to add it in a way that is hopefully still easy to use with little room for user error. In this case only one person has ever saked for a non-standard framerate. The issue you referenced was somebody asking for 25/50fps, and that is a standard that I had not thought about previously, so once it was discussed it was added.

I'm certainly not opposed to adding things, but I do think nothing should be quickly added in response to one person. Everything should be useful to most people without just adding more things people have to make choices about.

I hope this makes sense and doesn't come off as me just rejecting changes, but if we made a change for every request everything would be a freeform text field.

rmens:
On a random note; UX wise I like this dropdown more than the current slider.

gabek:
Personally I really like the consistency of the slider. We use it for Latency, CPU utilization, bitrate, and framerate. Just from a mental model standpoint it's nice to have similar ways to select values. Unfortunately the resolution has to be freeform text inputs, but I like how the other values are selected in the same way.

PR #3180, opened 2023-07-16
fix(chat): fix chat not using the correct messages selector. 

### LINKED ISSUE #166: Feature: Invitation links for "guests" to a wedding being streamed
I wonder if it would be a possibility to create a feature where you can create unique invitation links to allow access to the stream (and send them through email/facebook/etc).

When people use these links, they'll obviously be logged in automatically with whichever username the administrator has given them (imagine for instance "uncle Bob and auntie Mary"), and they won't be able to use features like change their username or set a password.

An extra possibility for this feature: each such invitation would then allow for a limited amount of concurrent logins (so in the cause of "uncle Bob and auntie Mary", two simultaneous sessions).

Usecase for this feature would be to make it as easy as possible for the older generation to use the stream to participate remotely in a weddingparty.

### LINKED ISSUE #3166: Hidden messages are not hidden unless the page is refreshed

1. Type a message in chat
2. Hide that message using moderator tools or via /admin/chat/messages
3. The message should be hidden on all clients

Currently the message does not get hidden unless the client refreshes the page

gabek:
Looks like at some point the wrong array of messages was changed to getting passed to chat. The two embed chat views are still correct, though.

Closes #3166

### DIFF
--- a/web/components/stores/ClientConfigStore.tsx
+++ b/web/components/stores/ClientConfigStore.tsx
@@ -174,7 +174,7 @@ export const ClientConfigStore: FC = () => {
   const setAppState = useSetRecoilState<AppStateOptions>(appStateAtom);
   const setGlobalFatalErrorMessage = useSetRecoilState<DisplayableError>(fatalErrorStateAtom);
   const setWebsocketService = useSetRecoilState<WebsocketService>(websocketServiceAtom);
-  const [hiddenMessageIds, setHiddenMessageIds] = useRecoilState<string[]>(removedMessageIdsAtom);
+  const setHiddenMessageIds = useSetRecoilState<string[]>(removedMessageIdsAtom);
   const [hasLoadedConfig, setHasLoadedConfig] = useState(false);
 
   let ws: WebsocketService;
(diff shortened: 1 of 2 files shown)


take2222:
EXCELLENT !

janWilejan:
Well this hides the message for the moderator, but not for the user whose message was hidden. Also, it seems messages get un-hidden if you hide multiple in a row?

janWilejan:
It seems only the last message muted gets hidden. If you hide 2 or more messages, the first gets un-hidden.

gabek:
Good catch! I only tested the inline chat moderation where you can only remove one message at a time, I didn't think to test the admin tools. I'll see what's going on over there.

janWilejan:
You don't need the admin tools to reproduce this. Mute 2 messages in the inline chat moderation.

Looks like the culprit is `web/components/stores/ClientConfigStore.tsx:279`. The `hiddenMessageIds` keeps the original value (an empty array). So after the first message it gets set to `[] + [message1]` and after the second it gets set to `[] + [message2]`

gabek:
Fixed!

PR #3183, opened 2023-07-17
Fixed localhost IP ban

### LINKED ISSUE #3172: Don't IP ban localhost

to reproduce:
1. Connect to an Owncast server running locally and make this account moderator.
2. Connect again using a separate browser or container tab.
3. Ban the second user.
4. Both users are banned.

Owncast makes IP bans by default, but we never need to ban localhost (127.0.0.1 or ::1).
Banning local users is sometimes useful for testing, but it shouldn't IP ban all local users.

mrsoong:
Fixes an issue described in #3172 by not enforcing IP bans on localhost (127.0.0.1 on ipv4 and ::1 on ipv6).

### DIFF
--- a/controllers/admin/chat.go
+++ b/controllers/admin/chat.go
@@ -164,12 +164,17 @@ func UpdateUserEnabled(w http.ResponseWriter, r *http.Request) {
 		disconnectedUser := user.GetUserByID(request.UserID)
 		_ = chat.SendSystemAction(fmt.Sprintf("**%s** has been removed from chat.", disconnectedUser.DisplayName), true)
 
+		localIP4Address := "127.0.0.1"
+		localIP6Address := "::1"
+
 		// Ban this user's IP address.
 		for _, client := range clients {
 			ipAddress := client.IPAddress
-			reason := fmt.Sprintf("Banning of %s", disconnectedUser.DisplayName)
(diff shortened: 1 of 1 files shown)


janWilejan:
This looks good to me. Thanks!

@gabek Does Owncast's backend really store IP addresses as human-readable strings? Is that common practice in go? I'd expect it to be parsed to something like `union {ipv4addr, ipv6addr}` in C or  `enum IpAddr` in rust

gabek:
An IP is either an ipv4 or an ipv6 address, so I guess I don't understand why you'd store an address as something other than a primitive type. If there's missing functionality or something isn't working we should certainly improve on it.

janWilejan:
There's nothing missing or broken about it per se, I just found it surprising. It looks like the recommended way to deal with IP address in go is `net.IP`: https://pkg.go.dev/net#IP . Maybe this should be fixed as part of the 0.2 refactor?

Also, this discussion isn't really about this PR in particular, so i'm gonna merge.

<blockquote><div><strong><a href="https://pkg.go.dev/net#IP">net package - net - Go Packages</a></strong></div><div>Package net provides a portable interface for network I/O, including TCP/IP, UDP, domain name resolution, and Unix domain sockets.</div></blockquote>

PR #3194, opened 2023-07-21
NotifyReminderPopup uses a custom Popover

### LINKED ISSUE #3191: Notify popup misses the button sometimes

It seems the popover assumes that the thing it's pointing to won't move. If chat opens or closes after the popover appears, it looks strange.
image

janWilejan:
the antd Popover calculated the position when opened and didn't recalculate when chat opens or closes or the window scrolls, causing it to be placed wrong.

closes #3191

### DIFF
--- a/web/components/ui/NotifyReminderPopup/NotifyReminderPopup.module.scss
+++ b/web/components/ui/NotifyReminderPopup/NotifyReminderPopup.module.scss
@@ -3,6 +3,7 @@
 }
 
 .contentbutton {
+  width: 150px;
   background-color: transparent;
   border: none;
   text-align: left;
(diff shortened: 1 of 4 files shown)


gabek:
Storybook shows the popup not aligned with the button. I'm guessing this is just an issue with the popup's story, but do you know why that might be?

https://629132c6e23893003a9e89c5-equleidnrm.chromatic.com/?path=%2Fstory%2Fowncast-components-notify-reminder--active

<blockquote><div><strong><a href="https://629132c6e23893003a9e89c5-equleidnrm.chromatic.com/?path=%2Fstory%2Fowncast-components-notify-reminder--active">Webpack App</a></strong></div></blockquote>

janWilejan:
The storybook wants the popover button to be 100% width, whereas the real interface puts the popover inside a flexbox where it can't grow. When the button's container grows to full width, the popover goes over the unused space.

gabek:
Ah ha! Thanks for the fix and making Storybook happy, I had seen this bug in the past and forgot about it because I didn't see it often, so thanks for catching it.

PR #3208, opened 2023-07-24
s3Storage: fix slice bounds out of range error

jprjr:
Ran into this while testing out storage with a new, empty bucket - on the first iteration of trying to clean up segments, the offset was 30, but the number of segments only 19. This caused a panic when returning the slice `[30:18]`. The offset is now capped to return an empty slice in this case.

Additionally - now that an empty slice can be returned - the number of segments is checked before calling `deleteObjects` to prevent an s3 error about a malformed request.

### DIFF
--- a/core/storageproviders/s3Storage.go
+++ b/core/storageproviders/s3Storage.go
@@ -200,7 +200,9 @@ func (s *S3Storage) Cleanup() error {
 		return err
 	}
 
-	s.deleteObjects(keys)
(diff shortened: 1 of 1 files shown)


gabek:
Thanks for testing this and for the fix! I think I was the only person to have used the new cleanup functionality so far. I hadn't experienced this, but I didn't use an empty bucket either. Good catch!

PR #3214, opened 2023-07-26
Embed uses flexbox instead of height: 96vh+4vh

### LINKED ISSUE #3210: Embed heights seem to still be off

A change was made in #3104 to fix the statusbar not showing on embeds due to an incorrect height. But it seems now it only _kind of_ shows. I'm not sure if the problem is with the embedded content, or if the opengraph tags saying how large the frame should be needs to increased.

Embedded into a Mastodon timeline
image

From https://web.batstationrad.io/
image

janWilejan:
closes #3210

This change fixes the embed, but it also affects the main layout.

### DIFF
--- a/web/pages/embed/video/index.tsx
+++ b/web/pages/embed/video/index.tsx
@@ -1,4 +1,5 @@
 /* eslint-disable react/no-unknown-property */
+import classnames from 'classnames';
 import React, { useEffect } from 'react';
 import { useRecoilValue } from 'recoil';
 import { useRouter } from 'next/router';
@@ -102,7 +103,7 @@ export default function VideoEmbed() {
     <>
       <ClientConfigStore />
       <Theme />
-      <div className="video-embed">{getView()}</div>
+      <div className={classnames([styles.videoEmbed, "video-embed"])}>{getView()}</div>
     </>
   );
 }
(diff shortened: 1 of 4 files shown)


web/pages/embed/video/index.tsx
```diff
@@ -40,10 +40,13 @@ export default function VideoEmbed() {
    * query parameters ourselves
    */
   const path = router.asPath.split('?')[1] ?? '';
-  const query = path.split('&').reduce((currQuery, part) => {
-    const [key, value] = part.split('=');
-    return { ...currQuery, [key]: value };
-  }, {} as Record<string, string>);
+  const query = path.split('&').reduce(
```
janWilejan, line 43: why is this bot re-prettifying code that's already been prettified? I didn't even touch these lines of code.

gabek: Maybe prettier updated rules, or maybe for some reason it didn't get prettified previously. It's annoying, but I wouldn't worry about it.

gabek:
Looks like there's an issue on mobile. The player is covering the entire vertical viewport on mobile, even covering up the chat button.

image

image

gabek:
@janWilejan Not sure if you saw this, but I thought I'd ping you just in case you didn't.

PR #3228, opened 2023-08-01
fix emojis overflowing the chat message

janWilejan:
fix UI bug likely introduced in 58bc3ac173
image


gabek:
I reworked how Javascript linting was setup recently, so there's probably some bugs in that like you see here. I just made a change that hopefully fixes it. I'll update this PR with `develop` so hopefully it'll work.

gabek:
Ugh I haven't been able to fix it. I'll have to look at it later. It's supposed to only lint changed files, but that's not what's happening.

gabek:
Thank you!

PR #3250, opened 2023-08-10
User flag gap is back now.

DineshDhanji:
Added margin ( 2xp to the left & right ) to the badge SCSS in "owncast-develop\web\components\chat\ChatUserBadge\ChatUserBadge.module.scss".
User Flags


DineshDhanji:
It's done.

PR #3256, opened 2023-08-13
Check if proposed name change is an API name

### LINKED ISSUE #3234: [Feature Request] Protect API displayNames in addition to authenticated user displayNames

Had a case today where somebody came into the chat, they changed their displayName and color to match our chatbot. All they really did is send a few messages, and since the username was missing the bot icon it was pretty clear it's not actually the bot. Still, it would be nice to prevent that in the first place.

I think this could be implemented by changing the "IsDisplayNameAvailable" query to:

```
select count(*) from users where display_name = $1 and ( type='API' or (authenticated_at is not null and disabled_at is null));
```

This should effectively make all the API display names unavailable for users.

I can author the change, but wanted to check if anyone using Owncast expects to be able to make their own display name match a bot display name. My guess would be no but, y'know, always worth asking.

jprjr:
Fixes #3234


gabek:
Nice, thanks!

PR #3291, opened 2023-09-04
feat(chat): add support for chat part messages. Closes #3201

### LINKED ISSUE #3201: Feature Request - Chat user leave event

Hi there, one thing I'll occasionally run into with my stream is users not realizing another chatter has left.

Granted, being connected doesn't 100% mean they're going to see it - but being disconnected means they absolutely won't see it, right? So that would be a nice-to-have event.

gabek:
Adding part messages did require changing some previous chat functionality. Previously we weren't showing part messages, so we had the flexibility that if a user had been active within 10 minutes, if they rejoin, it doesn't show a join message. The idea here was to allow people to switch devices, browsers, refresh pages, step away for a minute and come back, etc. without flooding the channel with join messages.

Now, with the addition of `part` messages, this is removed. Since if we're showing part messages, then we have to show the join messages as well without this buffer time. The only buffer time is the 10 seconds between when a user disconnects from chat and when the part message is sent out.


jprjr (approved):
Had some time to check out this branch and take a look - everything looks great!

gabek:
Thanks for checking it out!

PR #3295, opened 2023-09-05
Block Private URLs at `serverurl` API endpoint

### LINKED ISSUE #3268: Admin: Don't let people set private IP addresses as public url

I've noticed a few times people set an internal (192.168.x.x 10.0.0.x) IP address as their public hostname, and then try to register to the directory. This obviously gets rejected at the directory, but we shouldn't let people set these values in the first place. The users probably think it's fine and wonder why it's not working.

shreyass-ranganatha:
Solves #3268 by introducing a regex check for Private URLs of the range `10.0.0.0/8` and `192.168.0.0/16` at the `/api/admin/config/serverurl` API endpoint

### DIFF
--- a/controllers/admin/config.go
+++ b/controllers/admin/config.go
@@ -5,11 +5,11 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/netip"
 	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
-	"regexp"
 
 	"github.com/owncast/owncast/activitypub/outbox"
 	"github.com/owncast/owncast/controllers"
@@ -407,12 +407,11 @@ func SetServerURL(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Block private IPv4 URLs of the range 10.0.0.0/8 and 192.168.0.0/16 with regexp
-	matchPvtURL_10, _ := regexp.MatchString(`10\.\d{1,3}\.\d{1,3}\.\d{1,3}.*`, serverHostString)
-	matchPvtURL_192_168, _ := regexp.MatchString(`192\.168\.\d{1,3}\.\d{1,3}.*`, serverHostString)
+	// Block Private IP URLs
+	ipAddr, ipErr := netip.ParseAddr(utils.GetHostnameWithoutPortFromURLString(rawValue))
 
-	if matchPvtURL_10 || matchPvtURL_192_168 {
-		controllers.WriteSimpleResponse(w, false, "Server URL cannot be a private URL")
+	if ipErr == nil && ipAddr.IsPrivate() {
+		controllers.WriteSimpleResponse(w, false, "Server URL cannot be private")
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


gabek:
Sorry I forgot to add this to the original list, the `172.16.0.0` range is also private, would you mind adding that to your check as well? Thanks so much!

shreyass-ranganatha:
Definitely, I'm on it.

About the lint workflow error- I'm assuming it's because of the `regexp` import ? I'll fix it if that's it, unless it's something else, do let me know

gabek:
> About the lint workflow error- I'm assuming it's because of the `regexp` import ? I'll fix it if that's it, unless it's something else, do let me know

Yeah, it's a misleading error. "Error: File is not `goimports`-ed (goimports)" means the order of the imports don't match the way that `goimports` would order them, so if you run `gofmt` on it, I think it'll reorder them.

Since this is a server-side validation, do you know offhand if the server generated error is showing up in the admin properly when you try to save a private IP? I feel like it should, but admitedly there aren't a lot of error cases when it comes to admin APIs so these don't get checked often.

gabek:
Out of nowhere I remembered that Go added some functionality for IP addresses not long ago. I wondered if they had a built-in way for this check, and surprisingly they do! https://pkg.go.dev/net/netip#Addr.IsPrivate

<blockquote><div><strong><a href="https://pkg.go.dev/net/netip#Addr.IsPrivate">netip package - net/netip - Go Packages</a></strong></div><div>Package netip defines an IP address type that's a small value type.</div></blockquote>

shreyass-ranganatha:
> Since this is a server-side validation, do you know offhand if the server generated error is showing up in the admin properly when you try to save a private IP? I feel like it should, but admitedly there aren't a lot of error cases when it comes to admin APIs so these don't get checked often.

image

I've tested the error replies with a range of IPs, although the admin panel doesn't have any way to show any forms of errors so far, it just doesn't show "Success!"

shreyass-ranganatha:
> Out of nowhere I remembered that Go added some functionality for IP addresses not long ago. I wondered if they had a built-in way for this check, and surprisingly they do! https://pkg.go.dev/net/netip#Addr.IsPrivate

This one's great, I moved the checks to this package

<blockquote><div><strong><a href="https://pkg.go.dev/net/netip#Addr.IsPrivate">netip package - net/netip - Go Packages</a></strong></div><div>Package netip defines an IP address type that's a small value type.</div></blockquote>

gabek:
> > Since this is a server-side validation, do you know offhand if the server generated error is showing up in the admin properly when you try to save a private IP? I feel like it should, but admitedly there aren't a lot of error cases when it comes to admin APIs so these don't get checked often.
> 
> image
> 
> I've tested the error replies with a range of IPs, although the admin panel doesn't have any way to show any forms of errors so far, it just doesn't show "Success!"

Oh bummer. I'll file another issue to address that.

gabek:
Thanks for working on this and being cool with the back and forth, @shreyass-ranganatha! :1st_place_medal:

PR #3299, opened 2023-09-07
Handle error thrown in postConfigUpdateToAPI

### LINKED ISSUE #3297: Handle people trying to save invalid URLs in the admin

The backend will reject any invalid (private) URLs that people try to save in the admin. However, the frontend is unable to handle these errors.

When an error is received, we should display it to the user.

image

See #3295

image

t1ff4ny:
**tldr**
This fix should resolve issue #3297 where the UI would display an uncaught error popup if the user tried to save an invalid url and received a non-2XX response from the server. The issue was broader than just this text field and affected all areas that called `postConfigUpdateToAPI()`.  

**More details**
1. `fetchData` will throw an Error when the `response.ok` is falsey
2. `postConfigUpdateToAPI` calls `await fetchData(...)` and does not handle the error being thrown. Instead it is checking `response.success && onSuccess` to handle success cases and anything else it will handle as error if there is an `onError` prop supplied (`web/utils/config-constants.tsx`)

Anywhere that uses `postConfigUpdateToAPI` will fail on error from server, not just this particular text field. I checked for direct `fetchData` calls throughout the code, most references will wrap it around a try/catch which is good but there are several instances where there is no try/catch so any server error will also most likely result in an uncaught error on the UI.

The two options that I considered and ultimately I went with option B.
A. Refactor `fetchData` to not throw an error when `response.ok` is falsey. CONS: You will need to refactor all `fetchData` calls.
B. Add try/catch in `postConfigUpdateToAPI` to handle errors thrown. PROs: We won't need to refactor any `fetchData` calls nor `postConfigUpdateToAPI` calls.

### DIFF
--- a/web/utils/config-constants.tsx
+++ b/web/utils/config-constants.tsx
@@ -54,15 +54,22 @@ const TEXTFIELD_TYPE_URL = 'url';
 
 export async function postConfigUpdateToAPI(args: ApiPostArgs) {
   const { apiPath, data, onSuccess, onError } = args;
-  const result = await fetchData(`${SERVER_CONFIG_UPDATE_URL}${apiPath}`, {
-    data,
-    method: 'POST',
-    auth: true,
-  });
-  if (result.success && onSuccess) {
-    onSuccess(result.message);
-  } else if (onError) {
-    onError(result.message);
+  try {
+    const result = await fetchData(`${SERVER_CONFIG_UPDATE_URL}${apiPath}`, {
+      data,
+      method: 'POST',
+      auth: true,
+    });
+    if (result.success && onSuccess) {
+      onSuccess(result.message);
+    } else if (onError) {
+      onError(result.message);
+    }
+  }
+  catch (e) {
+    if (onError) {
+      onError(e.message);
+    }
   }
 }
 


web/utils/config-constants.tsx
```diff
@@ -54,15 +54,22 @@ const TEXTFIELD_TYPE_URL = 'url';
...
-    onError(result.message);
+  try {
+    const result = await fetchData(`${SERVER_CONFIG_UPDATE_URL}${apiPath}`, {
+      data,
+      method: 'POST',
+      auth: true,
+    });
+    if (result.success && onSuccess) {
+      onSuccess(result.message);
+    } else if (onError) {
```
t1ff4ny, line 65: One small thing, can `response.ok` be truthy (in `fetchData`, which would cause no exception thrown) AND the `result.success` be falsey? Just depends on your server and how it responds to different scenarios. If so, then I can leave the else if in the original code and just add the catch so we don't break anything existing that could fall under that scenario

gabek:
@t1ff4ny Thanks so much! Tomorrow I hope to sit down and test this change, and then we can get it merged in.

web/utils/config-constants.tsx
gabek: My guess is it's probably unlikely, but I also don't see much of a downside to catching that, just in case. Thanks for asking!

gabek:
Thanks for looking at this, @t1ff4ny. This is great, since it fixes the problem for any field that could possibly return an error, but previously wasn't handling them, not just the single field that was the original issue. A nice improvement.

Great job!

PR #3311, opened 2023-09-16
Add support for remote serving endpoint on local storage

### LINKED ISSUE #3292: Serving Endpoint not used without an S3 backend

When you define a custom URL as a serving endpoint in the web interface, it has no affect.

Screenshot 2023-09-04 at 12 05 04
Screenshot 2023-09-04 at 12 05 21

Using the latest stable version 0.1.1 running on Docker

alyx161:
Fixes #3292

### DIFF
--- a/core/storageproviders/local.go
+++ b/core/storageproviders/local.go
@@ -13,7 +13,9 @@ import (
 )
 
 // LocalStorage represents an instance of the local storage provider for HLS video.
-type LocalStorage struct{}
+type LocalStorage struct {
+	host string
+}
 
 // NewLocalStorage returns a new LocalStorage instance.
 func NewLocalStorage() *LocalStorage {
@@ -22,6 +24,7 @@ func NewLocalStorage() *LocalStorage {
 
 // Setup configures this storage provider.
 func (s *LocalStorage) Setup() error {
+	s.host = data.GetVideoServingEndpoint()
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


core/storageproviders/local.go
```diff
@@ -42,8 +45,16 @@ func (s *LocalStorage) VariantPlaylistWritten(localFilePath string) {
 
 // MasterPlaylistWritten is called when the master hls playlist is written.
 func (s *LocalStorage) MasterPlaylistWritten(localFilePath string) {
-	if _, err := s.Save(localFilePath, 0); err != nil {
-		log.Warnln(err)
+
+	// If we're using a remote serving endpoint, we need to rewrite the master playlist
+	if s.host != "" {
+		if err := rewriteRemotePlaylist(localFilePath, s.host, ""); err != nil {
```
gabek, line 51: Originally, I called this `rewriteRemotePlaylist` because it was only used for rewriting a local playlist to be used as a remote playlist. I feel like the name is no longer accurate. Maybe it should be named to `rewritePlaylistLocations` or something? Can you think of a more representative name?

alyx161: True, it might be more accurate to give it a more generic name
Any particular reason you used the plural `rewritePlaylistLocations` instead of `rewritePlaylistLocation`?

gabek: The master playlist holds multiple media playlists, each has its own URL. So it needs to rewrite each one.

alyx161:
Superseded by #3313

PR #3315, opened 2023-09-19
Update Design.stories.mdx

Anubhav-Chauhan3367:
Fixed some typos.
Added punctuations.
Improved documentation.


gabek:
This seems to be a handful of unnecessary changes. "Improved" is something that should be discussed. Specify exactly what the problem with the document is, and how it should be changed. Adding a couple commas and changing sentences in meaningless ways is not a worthwhile or quality change.

PR #3332, opened 2023-09-27
Mute audio in 'this stream is offline' clip

### LINKED ISSUE #1965: Add an option to change the end stream clip
At the end of every stream there's an annoying clip that has nothing to do with the stream itself and seems only an "Owncast-brand-thing". Nothing wrong with having something like that, but there's no point in having no option at all to remove it, here's a couple of reason:

- if I've just ran a chill stream and I close the feed my viewers gets blasted with a not-very-pleasant audio in their ears
- injecting something different from what the streamer wants to actually stream without any control on it seems pretty bad, expecially on a self-hosted-platform.

Please consider adding an option to disable that or customize a stream-ended message.

vivlim:
per discussion in #1965


gabek:
I won't be near a machine that will be able to look at this file until next week, just know I'm not ignoring it. I simply can't test it.

Before then if you want to describe what the clip is and what your specific thoughts are around it are, some additional details might be useful.

vivlim:
sure, i’ll write something up later.

gabek:
This updated clip causes transcoding errors. See below:

```
ERRO[2023-10-03T18:53:52-07:00] [hls @ 0x5db6400] Unable to map stream at a:0
ERRO[2023-10-03T18:53:52-07:00] [hls @ 0x5db6400] Variant stream info update failed with status ffffffea
ERRO[2023-10-03T18:53:52-07:00] Could not write header for output file #0 (incorrect codec parameters ?): Invalid argument
ERRO[2023-10-03T18:53:52-07:00] Error initializing output stream 0:0 --
```

gabek:
For reference, I believe the problem here is you didn't _mute_ the audio, you _removed_ the audio.

PR #3338, opened 2023-10-05
Initial core functionality for replays and clips

gabek:
This is the core internal functionality to support stream replays and clips with Owncast. There is no UI, and it is disabled by default. Use `--enableReplayFeatures` to enable the feature flag.

Also enables S3 path prefix, to support a custom path within an object storage bucket.

This is not yet useful to people, as it's just a bunch of APIs (that do function), but is the first step to have this functionality. 

There is also no cleanup logic. When enabled all video is saved on disk forever.

---

## Details

- /api/replays returns a list of all the streams that are replayable.

URLs are generated as so:

-  /replay/_streamId_ for master playlists.
- /replay/_streamId_/_configurationId_ for media playlists.

## Testing in progress stream

1. Start a live stream
2. Visit http://localhost:8080/api/replays
3. Note the "manifest " path.
4. Put the manifest url in your livestreaming video player of choice.
5. Play it. It will be treated as an updating (event) style stream.

## Testing completed streams

1. Stop the live stream.
2. Visit http://localhost:8080/api/replays
3. Notice this stream is no longer in progress.
4. Put the manifest url in your livestreaming video player of choice.
5. Play it. It will be treated as a static (VOD) stream and will not update.

## Clips

- Follow the instructions above to have a replay of a stream.
- Get the `streamId` you want to create a clip from.
- Determine the start and end points, in seconds, for this clip, relative to the beginning of the stream.
- Use the clip create API to create the clip as a JSON payload to `/api/clip` with `streamId`, `clipTitle`, `relativeStartTimeSeconds` and `relativeEndTimeSeconds`
```
http http://localhost:8080/api/clip streamId="oYk3sL6Vg" clipTitle="Test clip 5" relativeStartTimeSeconds:=10 relativeEndTimeSeconds:=17
```

image

- List all clips with the clips api `/api/clips`.
image

- Fetch a single clip with `/clip/{clipId}`.
- It will return back a HLS master playlist that is playable.

image


controllers/clips.go
```diff
@@ -0,0 +1,167 @@
...
+		clipId := pathComponents[2]
+		outputConfigId := pathComponents[3]
+		getClipMediaPlaylist(clipId, outputConfigId, w)
+		return
+	}
+
+	BadRequestHandler(w, nil)
+}
+
+// getReplayMasterPlaylist will return a complete replay of a stream
```
gabek, line 131: Incorrect comment

PR #3348, opened 2023-10-08
Stop Firefox from adding mysterious hash

### LINKED ISSUE #3240: Firefox: URL adds a mysterious "#"

This doesn't happen on Chromium-based browsers. It doesn't seem to be impacting anything, but it's kind of annoying because it looks like it's there on purpose and makes you wonder why.

image

pjbollinger:
This resolves https://github.com/owncast/owncast/issues/3240

From the comments:
This was trickier than expected, but the root of the problem is Firefox will set `#` in the URL bar when `window.location.hash` is set to _any_ string, even a blank string. The morale of the story is, don't mutate base data if you just want to copy values. 😅

Sample of Firefox JavaScript console session that demonstrates the issue:
```javascript
>> window.location.href
"https://github.com/owncast/owncast/issues/3240"

>> const setBlankHash = () => { window.location.hash = ''; };
undefined

>> window.location.hash
""

>> window.location.href
"https://github.com/owncast/owncast/issues/3240"

>> setBlankHash()
undefined

>> // My browser just jumped to the top of the page
undefined

>> window.location.hash
""

>> window.location.href
"https://github.com/owncast/owncast/issues/3240#"
```

The main thing I'm uncertain about is, do we need to worry about query parameters? I don't think so since this is feeding the connection builder for websocket but let me know if it should be adjusted.

### DIFF
--- a/web/components/stores/ClientConfigStore.tsx
+++ b/web/components/stores/ClientConfigStore.tsx
@@ -374,9 +374,7 @@ export const ClientConfigStore: FC = () => {
       const { socketHostOverride } = clientConfig;
 
       // Get a copy of the browser location without #fragments.
-      const l = window.location;
-      l.hash = '';
-      const location = l.toString().replaceAll('#', '');
+      const location = window.location.origin + window.location.pathname;
       const host = socketHostOverride || location;
 
       ws = new WebsocketService(accessToken, '/ws', host);


gabek:
> The main thing I'm uncertain about is, do we need to worry about query parameters? I don't think so since this is feeding the connection builder for websocket but let me know if it should be adjusted.

I don't think there's currently a case where we'd care about them. There must have been some issue previously where adding fragments to the URL happened, and I'm guessing those fragments broke the websocket URL, so this was a way to make sure they weren't there. Though you found the downside of that approach :)

gabek:
This is fantastic, thanks for the fix! :1st_place_medal:

PR #3352, opened 2023-10-10
Fix embedded status bar being cut off

### LINKED ISSUE #3210: Embed heights seem to still be off

pjbollinger:
# Description

This should resolve https://github.com/owncast/owncast/issues/3210.

As recommended in the discussion, flexbox was the way to solve this issue. The main thing that needed to be addressed though was not applying too many constraints on the OwncastPlayer component just because it was embedded. By removing the embedded constraints, styling appears to be working as expected. I'll attach screenshots to the pull request.

If there are other screenshots or material that would be good for me to gather for these UI changes to help with the review, please let me know.

# Screenshots

## Various configurations of `iframe`s (embedded)

Screenshot 2023-10-09 at 9 43 29 PM

## `iframe` in docs (embedded)

Screenshot 2023-10-09 at 9 43 53 PM

## Emulated mobile view of main stream site using Firefox (non-embedded)

Screenshot 2023-10-09 at 9 44 09 PM

## Desktop view of main stream site (non-embedded)

Screenshot 2023-10-09 at 9 44 20 PM

### DIFF
--- a/web/components/video/OwncastPlayer/OwncastPlayer.module.scss
+++ b/web/components/video/OwncastPlayer/OwncastPlayer.module.scss
@@ -30,8 +30,3 @@
     grid-row: 1;
   }
 }
-
-.embedded {
-  height: 96vh; // Leave room for the status bar
-  max-height: unset;
-}
--- a/web/components/video/OwncastPlayer/OwncastPlayer.tsx
+++ b/web/components/video/OwncastPlayer/OwncastPlayer.tsx
@@ -29,7 +29,6 @@ export type OwncastPlayerProps = {
   initiallyMuted?: boolean;
   title: string;
   className?: string;
-  embedded?: boolean;
 };
 
 export const OwncastPlayer: FC<OwncastPlayerProps> = ({
@@ -38,7 +37,6 @@ export const OwncastPlayer: FC<OwncastPlayerProps> = ({
   initiallyMuted = false,
   title,
   className,
-  embedded = false,
 }) => {
   const VideoSettingsService = useContext(VideoSettingsServiceContext);
   const playerRef = React.useRef(null);
@@ -302,7 +300,7 @@ export const OwncastPlayer: FC<OwncastPlayerProps> = ({
       )}
     >
       <div
-        className={classNames(styles.container, className, embedded && styles.embedded)}
+        className={classNames(styles.container, className)}
         id="player"
       >
         {online && (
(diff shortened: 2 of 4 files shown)


gabek:
I haven't forgotten about this! I'll be doing some testing this weekend.

gabek:
Looks like these changes result in the player taking up the entire vertical viewport on mobile (iOS iPhone 14 Safari).

image

pjbollinger:
@gabek Could you give more details on that device info?

I tried it with my iPhone 13 (Pro?) running iOS 17.0.2 Safari and it's working OK.

image

gabek:
Sure!  Here's the details behind that screenshot:

image

Looks like iOS 17 on an iPhone 15 Pro is ok, though:
image

Chrome also seems fine:

image

I'll do some poking to see if there's any reason that one use case is off.

gabek:
Chrome on an iPhone 14, iOS 16 has the same issue:

image

gabek:
Samsung Galaxy + Chrome is ok:
image

I tried a bunch of other Android devices and browsers and all of those are fine too.

gabek:
Good news! iPhone 14 Pro iOS 16.5 works. I'm not sure what's specifically weird about 16.3 above, but if 16.5 is working without issues, then we're in good shape.

image

pjbollinger:
Thanks for testing! I tried to read through release notes but nothing jumped out to me that would cause this issue. In iOS 16.4, there was work done on Safari Web Extensions, so perhaps they did some work on WebKit too during that release?

iOS 16.3 is declining for usage, so I'm inclined to not worry about this platform+version specific bug. I'm not the one that is triaging issues though, so let me know if you want me to try and find the fix for iOS 16.3 as well.

image

gabek:
> iOS 16.3 is declining for usage, so I'm inclined to not worry about this platform+version specific bug. I'm not the one that is triaging issues though, so let me know if you want me to try and find the fix for iOS 16.3 as well.

I'm not concerned about 16.3. Like your stats above show, anybody still on iOS 16 is generally going to be on the most recent version of 16 if they haven't yet jumped to 17, so I see no problem there.

I'll do a bit more manual testing with different devices and browsers in the next few days. But this is great!

gabek:
I did a bunch of testing on different devices and browsers, and things look good! Let's get this merged in and it can get it deployed on test servers.

Thanks so much for working on this!

PR #3359, opened 2023-10-15
Fix geo details for viewers not showing on CDN connection

### LINKED ISSUE #3349: Geo details in viewer list is not populated when using a CDN

When going through a CDN, the list of viewers has a string with two IP addresses when listing those clients. One of the IPs is the CDN, the other is the actual client.

I'm thinking that the entire string (with the two addresses, and the comma) is being passed to the geo resolver and failing.

image

I'm assuming this log message is being reported when that doesn't work:

image

armadi1809:
This aims to resolve #3349 

### Description
As per the discussion on the issue, the `X-FORWARDED-FOR` header for a request will contain a comma separated list of the IP address of the client followed by those of all the proxies that the request went through. Therefore, when using a CDN, that header will contain the address of the actual client and the CDN. The solution proposed here is to always refer to the first address present in that header (the actual client) in the `GetIPAddressFromRequest` method.

### DIFF
--- a/utils/clientId.go
+++ b/utils/clientId.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"net"
 	"net/http"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


gabek:
Awesome! Could you add a quick test for this?

armadi1809:
I just added tests. I also realized that the  IPs in the `X-FORWARDED-FOR` header may contain ports so I modified the logic a little bit to account for that scenario. I am still a newbie to golang so any feedback here is appreciated. Thank you!

gabek:
Looks great, thanks so much! We can get this merged in, and then I can deploy it to our demo server (watch.owncast.online) who is behind a CDN, so we can do some real world testing.

armadi1809:
Awesome. Thank you for this opportunity. It's my first open source contribution ever 😀

gabek:
Wohoo! I'm glad it was a good experience :)

PR #3361, opened 2023-10-16
Remove "Live for XXX" message from aria-live area

### LINKED ISSUE #3334: live for <duration> announcements should not be in aria live region
## type

bug

## owncast version/tag:

0.1.1

## description

the live for XXX event fires in the aria live region every time the responsible component changes in the DOM. While this is very often in the first parts of the stream, only to ramp down as the stream progresses, this can get annoying very quickly because the screenreader announces every one of those changes individually.

To make a comparison for a second, imagine a box in the top-right corner of the screen flashing a bright red every couple seconds, distracting you from the chat or whatever the stream is about. While, say, the distance between the flashes increases, it is not an ideal user experience at all, it feels the same way to us, but worse, because it interrupts what the screenreader was saying before the announcement.

More important things should be in the aria live region instead, for example new messages, status events some of which are already there, but that's for another issue.

## expected behaviour

that information is not announced at all, because the element is visible in the DOM and we can inspect it, if we want to know for how long the stream has been ongoing

## steps to reproduce

* turn on a screenreader. For example, narrator for windows, orca for linux, voiceover for ios, or talkback for android
* open a browser to an owncast stream. It doesn't matter if you or someone else is streamming it, but preferable here is that the stream is in its early beginning, before the one hour mark, because then the effect is more observable
* just stay there for afew seconds, you should start getting spammed with live for xxx minutes and yyy seconds, one or so per second

## possible solutions

remove that entirely from the aria live region, we don't need that information live like that, there are better things we would rather be spammed with

armadi1809:
This should resolve #3334 

### Description 
This PR sets the `aria-live` attribute to "off" on the `<span>` responsible for rendering the "live for XXX" message to avoid spamming screen readers with that message on every rendering cycle (i.e every second). 

I made sure to reproduce the issue and that it went away after introducing this fix.

### DIFF
--- a/web/components/ui/Statusbar/Statusbar.tsx
+++ b/web/components/ui/Statusbar/Statusbar.tsx
@@ -79,7 +79,9 @@ export const Statusbar: FC<StatusbarProps> = ({
 
   return (
     <div className={classNames(styles.statusbar, className)} role="status">
-      <span className={styles.onlineMessage}>{onlineMessage}</span>
+      <span aria-live="off" className={styles.onlineMessage}>
+        {onlineMessage}
+      </span>
       <span className={styles.viewerCount}>{rightSideMessage}</span>
     </div>
   );


gabek:
Fantastic, thanks so much for looking at this! I'm sure this will be greatly appreciated by screen reader users. :100:

PR #3362, opened 2023-10-16
fix: issue with lint and prettier during js format build

### LINKED ISSUE #3360: Code formatting not running when it should under CI

When pushing Javascript code, it should be linting and formatting the code. It's getting skipped for some reason. It might be an issue with how we're using the ["changed files" action](https://github.com/owncast/owncast/blob/develop/.github/workflows/javascript-format-build.yml#L39), it might be returning zero changed files.

image

A-5ingh:
Fixes: #3360 

The issue was with tj-action/changed-files step, it was trying to find files under `web` folder while working folder was also the `web`.
I have provided the path as `web` and refactor the `src` in a single line adding all the file formats.

### DIFF
--- a/.github/workflows/javascript-format-build.yml
+++ b/.github/workflows/javascript-format-build.yml
@@ -40,17 +40,13 @@ jobs:
         id: changed-files
         uses: tj-actions/changed-files@v39
         with:
+          path: 'web'
           files_ignore: |
             static/**
             web/next.config.js
           files_yaml: |
             src:
-              - 'web/**.js'
-              - 'web/**.ts'
-              - 'web/**.tsx'
-              - 'web/**.jsx'
-              - 'web/**.css'
-              - 'web/**.md'
+              - '**/*.{js,ts,tsx,jsx,css,md}'
 
       - name: Cache node modules
         uses: actions/cache@v3


gabek:
Oh very cool! Good find! Is there any conflict between setting the [working directory](https://github.com/owncast/owncast/blob/918085d5fcc545f14f82e4ba0d491cbccb51a703/.github/workflows/javascript-format-build.yml#L21C1-L21C33) as `web` as well as telling the changed files action to specifically use `web` as well?

A-5ingh:
> Oh very cool! Good find! Is there any conflict between setting the [working directory](https://github.com/owncast/owncast/blob/918085d5fcc545f14f82e4ba0d491cbccb51a703/.github/workflows/javascript-format-build.yml#L21C1-L21C33) as `web` as well as telling the changed files action to specifically use `web` as well?

Not really. The working directory has to be web only because that is where we want to lint and prettify the code.
If we don't use `web` as path in the changed file action then it keeps returning the files that are changed alone with the `web` folder. For instance, if I change the file `ClientConfigStore.tsx`, it will return `web/components/stores/ClientConfigStore.tsx` in the changed files object. Whereas, we require the relative path - `components/stores/ClientConfigStore.tsx` in order for it to work correctly. Hence, added `web` to the changed file action.

gabek:
Got it! Thanks so much, @A-5ingh! Great fix.

PR #3372, opened 2023-10-18
Added a check for the port before calling the splitHostPort function

### LINKED ISSUE #3371: "missing port in address" error after IP address parsing change

Re #3359 @armadi1809 

From the looks of this error message, it looks like there is an issue with the IP/port split.

image

armadi1809:
Resolves #3371  

The `net.SplitHostPart` returns an error in case a port is not present (i.e, no ":" in the string provided to the function). I therefore added a check for the semicolon before calling that function. 

@gabek Sorry for the confusion. Also, please let me know if you would rather see this bug fixed in a different way.


gabek:
That works! I'll get it deployed to the demo server in the next day or so and give it a spin. Thank you so much for coming back to help with this!

gabek:
Success! Thank you!

image

PR #3376, opened 2023-10-20
Fix parsing of Authorization Bearer header

alyssais:
The semantics of the Authorization header are defined by RFC 9110, which says:

> It uses a case-insensitive token to identify the authentication scheme:

Therefore, "bearer", "Bearer", and "bEARER" are equivalent.  This patch fixes the parsing of the Authorization header to check for the Bearer authentication scheme case insensitively.

I've modified one of the test cases to use lowercase "bearer", so there's test coverage for this.


gabek:
Very cool! I had no idea it was supposed to be case-insensitive, good find. 🥇

PR #3383, opened 2023-10-23
fix: prevent floating mobile action menu button

### LINKED ISSUE #3271: Mobile action menu button does not scroll
As you scroll the page the button stays in the same location. 

image

Pranav2612000:
the absolute positioned mobile action button was attached to the body causing it to float during scrolling. Now, we wrap the action buttons inside a relative div so that they are attached to this div which prevents them from scrolling with page.
Fixes #3271


Pranav2612000:
@gabek The Javascript / Code formatting check is failing due to possible some other issue. Can you take a look?

gabek:
Nothing has changed with these workflows, but now it fails with `failed with exit code 128`. I'll have to do some digging to see what is going on.

gabek:
This might be crazy, but from what I can tell in the logs it's unhappy because you don't have a `develop` branch on your fork? I'm not sure why your fork would be different than `owncast/owncast`, but could you try it?

Pranav2612000:
Yup. I think the test passed now.
Thank you for your help in figuring this out.

> I'm not sure why your fork would be different than owncast/owncast

That's because Github has an option of whether you want to fork the whole project or just the `main` branch. And I think I chose just to fork `main`

gabek:
Thank you so much, it works great!

PR #3397, opened 2023-10-26
Ignore static/web for codeql

### LINKED ISSUE #2687: Stop CodeQL from reporting on compiled Javascript

Compiled Javascript should not be treated as raw source since it's all optimized and minified. CodeQL keeps reporting on these compiled files and I have to manually tell it to shut up each time.

It would be really great to fix this for good but apparently it'll take some effort.

https://josh-ops.com/posts/github-codeql-ignore-files/

mdonoughe:
The [documentation](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#specifying-directories-to-scan) isn't that easy to find, and it only works for interpreted languages, but it's possible to tell the CodeQL action to ignore certain paths.

I've never gotten the reports mentioned in #2687, but with this configuration change, I no longer see `static/web` in the job log so it looks like it's working.

Fixes #2687


mdonoughe:
The Go job generates a warning if you have `paths-ignore` in your configuration, so I split the configuration.

mdonoughe:
And apparently the initialization just fails if you give it a completely empty config file? I'm not sure what the configuration name is for but I set it and now the Go job initializes again.

gabek:
This is awesome, thank you so much!

PR #3437, opened 2023-11-25
fix(admin): sanitize fediverse username

### LINKED ISSUE #3424: Sanitize Fediverse username

Apparently we allow people to type whatever they want, creating a mess like the following where somebody put an `@` in their actual username.

We should reject usernames that have special characters in them.

image

varungujarathi9:
fixes #3424
- The `update` button is disabled if the username contains special chars
- Attached working screenshots below
- In the screenshot it shows the `tip` twice, but I hope it is just a storybook issue
- Let me know if the `tip` message needs to be changed
- Changed the maxLength from 10 to 50, while testing it wasn't taking my full Mastodon username
- Added `.stories.tsx` for config federation page

**Safe username**
Screenshot from 2023-11-24 23-44-02
**Unsafe username**
Screenshot from 2023-11-24 23-43-58
Screenshot from 2023-11-24 23-43-42

### DIFF
--- a/web/pages/admin/config-federation.stories.tsx
+++ b/web/pages/admin/config-federation.stories.tsx
@@ -1,0 +1,17 @@
+import { Meta } from '@storybook/react';
+import ConfigFederation from './config-federation';
+// import Mock from '../../../stories/assets/mocks/chatmessage-action.png';
+
+const meta = {
+  title: 'owncast/Admin/Config/Federation',
+  component: ConfigFederation,
+  parameters: {},
+} satisfies Meta<typeof ConfigFederation>;
+
+export default meta;
+
+const Template = arguments_ => <ConfigFederation {...arguments_} />;
+
+export const Default = Template.bind({});
+
+Default.args = {};
--- a/web/pages/admin/config-federation.stories.tsx
+++ b/web/pages/admin/config-federation.stories.tsx
@@ -1,6 +1,5 @@
 import { Meta } from '@storybook/react';
 import ConfigFederation from './config-federation';
-// import Mock from '../../../stories/assets/mocks/chatmessage-action.png';
 
 const meta = {
   title: 'owncast/Admin/Config/Federation',
--- a/web/components/admin/TextFieldWithSubmit.tsx
+++ b/web/components/admin/TextFieldWithSubmit.tsx
@@ -156,5 +156,4 @@ export const TextFieldWithSubmit: FC<TextFieldWithSubmitProps> = ({
 TextFieldWithSubmit.defaultProps = {
   configPath: '',
   initialValue: '',
-  isValid: true,
 };
--- a/web/components/admin/TextFieldWithSubmit.tsx
+++ b/web/components/admin/TextFieldWithSubmit.tsx
@@ -22,7 +22,6 @@ export type TextFieldWithSubmitProps = TextFieldProps & {
   configPath?: string;
   initialValue?: string;
   hasComplexityRequirements?: boolean;
-  isValid?: boolean;
 };
 
 export const TextFieldWithSubmit: FC<TextFieldWithSubmitProps> = ({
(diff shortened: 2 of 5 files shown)


web/pages/admin/config-federation.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import { Meta } from '@storybook/react';
+import ConfigFederation from './config-federation';
+// import Mock from '../../../stories/assets/mocks/chatmessage-action.png';
```
gabek: Could you remove this commented out line if it's not in use?

varungujarathi9:
Done!

gabek:
It seems like the "Update" button is no longer showing.

Before the change

image

After the change

image

varungujarathi9:
@gabek Is there a button below the `tip` ?

Not sure why is this behavior, but in my previous screenshots as well button is below tip

gabek:
Oh! I see. Why did the button get moved? If I wasn't able to know where it was, I'm guessing others won't see it either. Can it be moved back to where it was originally?

varungujarathi9:
I don't remember making any changes to the `SCSS` or anything else that might change the look.

But I will spend some time tonight comparing with last good code

varungujarathi9:
Hey @gabek 
I checked the last commit where `config-federation.tsx` was untouched it looks like the button is below `tip`.

This might be the behavior in `storybook` view only. (Check below)

I think, after building it and serving on backend, it should look same as before.

Screenshot from 2023-12-02 18-36-29


I also finally got the linter working on my IDE, so I fixed and checked the lint issues!

gabek:
Great news about the linter!

When I previously tested it, I didn't test through Storybook, I just went to `/admin`. But I'll check it out again this weekend and see what's going on.

varungujarathi9:
Screenshot from 2023-12-08 12-47-35

Hi @gabek 
Just ran the whole project (frontend and backend) to check the UI, it looks fine on the /admin/config-federation URL, attached screenshot.

Also for the CI error, I am moving the `config-federation.stories.tsx` to `stories` folder from `pages` folder.

varungujarathi9:
Hey @gabek 
Were you able to check the latest commits of this PR? Let me know if any change is required

gabek:
Hi there!

I'm sorry if the original issue wasn't clear. In this case, we don't want to allow any special characters:

So `me@server.com` should not be accepted. `me` should be accepted. `me123` should be accepted. `me!` should not be accepted. `me#` should not be. For this text field, it's just a username, not a domain or a full account of any sort.

image

varungujarathi9:
Oh cool, I made changes accordingly
Please check @gabek

gabek:
It still seems to let me type any character I want in that text field.

image

varungujarathi9:
Hmm, that's strange, when I run it on my local and give the input with special chars, the 'Update' button doesn't get enabled(See screenshots below).
(Storybook)
image

(Actual app)
image

Alternatively, do you want me to sanitize the input and then set it as the `value` of the `field`? Eg: If user types `me@server.com`, I simply set the field value as `meserver.com`

varungujarathi9:
Hey @gabek any updates?

gabek:
Oh that's the confusion, it still let me type in any characters I wanted, but because the update button got moved below the text, there was no way to know they were invalid.

I do think it should stop you from typing invalid characters completely, so this confusion isn't possible. I also don't know why the update button position has changed. Sorry, I wasn't more clear about that. If we don't filter out the invalid characters, then it's up to the user to guess what characters are valid or invalid because there's no way to know.

image

varungujarathi9:
Hi @gabek 
I've made the changes as you suggested. Sorry, I was confused earlier. Please let me know if any more changes are needed.

gabek:
Thank you! I'll check it out.

gabek:
Works great! Thank you so much for your patience and continuing to update it, you're awesome. :100:

PR #3458, opened 2023-12-01
Draft: Mark strings for translation.

taintedcypher:


taintedcypher:
@gabek Feel to merge whenever, disregarding the Draft status. (**after checking I didn't mangle anything up** :smile: I'm doing this as someone who doesn't know a thing about React and only a bit about JS).

I'll just make a new PR if and when this one gets merged.
I feel like it'll be easier for translators to get strings in smaller batches, than a whole bunch of strings at once later! Also it will reduce the risk of someone else starting marking up the sames files.

If I need to correct things, don't hesitate to tell me.

gabek:
It looks like there are some linter errors being reported by CI. 

I'm not sure if this problem is in the `localization` branch or not but I can check it out later. It seems to show `[Object object]` as the time. But this might not be from your PR.

image

taintedcypher:
I'm afraid I can't find the time to look for the issues and correct them. If anyone want to have a go at it, we could merge this and have plenty of new strings to translate :)

gabek:
I can't focus on localization either, that's why I was hoping others could jump in. Since we don't know if there are issues or not i'm afraid this can't go in until it can get tested.

Issue #3248, opened 2023-08-09
Followers design/layout needs improvement


gabek:
### Share your bug report, feature request, or comment.

At its best it looks bad. At its worst it's really bad. Anybody have any ideas to improve it?

![image](https://github.com/owncast/owncast/assets/414923/de6f5510-1e68-4444-9d73-72cf88c69a47)

thezeeshann:
@gabek I would like to work on this issue. assign me

DineshDhanji:
Hi @gabek! I would like to work on this issue. I have also attached my idea. 
#### Suggestion
![Suggestion](https://github.com/owncast/owncast/assets/56204876/6b67b384-5a8d-4ecc-80b4-8f46ffcd7e13)

In case of longer name or text, we can truncate it (Like the second picture).

gabek:
It does look better! But it would truncate for pretty much everyone, since accounts are almost always going to be longer than that. So if it's going to truncate for everybody, I'm not sure if that's the right design. What do you think?

For clarification, here's what it looks like now.

![image](https://github.com/owncast/owncast/assets/414923/f1ed2555-e233-4e66-bd9f-c9cf6c01d551)

thezeeshann:
@gabek I think this one is better

gabek:
> @gabek I think this one is better

Did you mean to attach something?

thezeeshann:
> It does look better! But it would truncate for pretty much everyone, since accounts are almost always going to be longer than that. So if it's going to truncate for everybody, I'm not sure if that's the right design. What do you think?
> 
> For clarification, here's what it looks like now.
> 
> ![image](https://user-images.githubusercontent.com/414923/259463769-f1ed2555-e233-4e66-bd9f-c9cf6c01d551.png)

I'm just saying this design is better

jerlinjosephh:
I would like to work on this issue. assign me this.

imryanide:
Hii, is this still up for grabs?

gabek:
@imryanide Absolutely! Thanks for wanting to be involved! If you have some ideas or mockups I'm happy to discuss ideas with you before making any changes if you'd like.

Here's a link to the corresponding components in Storybook: https://owncast.online/components/?path=/docs/owncast-components-followers-followers-collection--no-followers

AMalfez:
That looks better.
Here are some of my ideas:
1. give some padding in icon and center everything inside each card.
2. can put a word limit on username so that design can look uniform.
[...]

PR #3469, opened 2023-12-06
State-changed events modified to be in aria live region

### LINKED ISSUE #3335: chat messages and state changed events should be in aria live region
## type

bug

## owncast version

0.1.1

## description

a lively stream needs live feedback. This applies both in the case of users animatedly talking to one another or with the host, as well as if the host needs a screenreader to successfully use the computer.

Getting events as soon as possible, in this case, is important because it's related to live discussion.

The way this is currently done is by manually going in the chat buffer and reading messages. For bigger chat, this is a problem, because catching up to messages is difficult with the current scrolling system, plus it's a big stream after all.

Even as a user it's difficult, because while the placement of their cursor, aka unread messages, isn't as important as for the owner/moderator, not getting a reply to a message because one didn't check the message buffer often enough isn't cool

## steps to reproduce

* turn on a screenreader. Good examples are narrator for windows, talkback for android, voiceover for apple devices
* open an owncast stream
* either generate traffic yourself, or go to a stream which already has a high activity
* observe, screenreader doesn't read some things

## actual behaviour

chat isn't read automatically by the screenreader when it comes in. Same about people joining, changing their nicknames, authorising with fedi, leaving, etc

## expected behaviour

chat is read automatically, perhaps with a sound per message if the window of the browser isn't focused and screenreader can't read the aria region because of that, see mastodon. Also, optionally have other events in the aria region, such as person joins, changes name, leaves, etc

## remarcs

this is the reverse problem to #3334, where two much information was dumped in the live region. If both of these are solved, owncast would suddanly become much more accessible

riybha216:
## Description

Previously, new chat messages were not being read automatically by the screenreader. In order to fix this, we created a hidden div that solely exists in the live region, and added the most recent addition to the `messages` array. Now, the screenreader will read the following script: {user's display name} said {chat message} automatically. Note that this reading is also done chronologically -- if two users send a chat message simultaneously, it will read the first user's message followed by the second chat message. Moreover, the reading of the chat will take a lower priority over users clicking other elements on the screen, ensuring that multiple elements on the screen are not simultaneously read and the user is not overloaded.

## Testing

We've tested this fix by following the steps outlined:

1) Download a screen reader (i.e. NVDA).
2) Run the stream alongside the screen reader, with users inputting chat messages.
3) The screen reader software will read the most recent chat message, along with the username of the chat's user.

## Issue
Resolves #3335


gabek:
This is great, thanks so much! I'll do some testing this weekend and we can get it merged in.

riybha216:
Hi @gabek, we were wondering if there were any updates to this thread. We'd be happy to implement any changes necessary! Thanks!

gabek:
Thanks for being patient! I haven't forgotten about you, sorry for the delay!

gabek:
Just as an update, I'm running into issues testing this on Linux, since NVDA doesn't run on Linux and all the screen readers on Linux seem to be flaky. So I think I'm going to have to change gears.

riybha216:
Let us know if we can do anything to help. I've tested this on MacOS, and my teammate has tested this on Windows. Happy to provide video output if necessary.

gabek:
Thanks so much for working on this! I was finally able to test this with Voiceover.

Sometimes it seems to work really well, and other times it's just throwing random stuff around.

In this example, I typed, and sent the message "This is a new user test one two three" and it only got as far as reading "This is a new user test" and then stopped reading and said "send". I think short sentences are ok, but longer ones get interrupted with the word "send".

https://github.com/owncast/owncast/assets/414923/aa88b864-2c0e-41ff-8b4c-fabb9447183b

Also, because of message collapsing, the user who sent the message only gets spoken the first time (as far as I could tell in my testing anyway, or maybe other things are just getting in the way).

I'm not sure if you have any thoughts about any of this. It's probably way better than it was, though!

riybha216:
Hi, thanks for testing this! Sorry for the delay -- I'll be sure to get back to this by the end of this weekend, if that's alright.

gabek:
I'm going to go ahead and merge this in, since I believe with your changes it's better than it was before. It's a net positive. It can be iterated on over time.

riybha216:
Hi @gabek, I apologize for the delay in response. Somehow things just piled up and I couldn't get back to you in time.

Would you like to continue work on this feature? If so, I'd love to improve what we've worked on so far and be more responsive. If there are higher-priority items to work on, I'd love to contribute to those! Please let me know what you think. Thank you so much!

gabek:
HI @riybha216, great to hear from you! A couple of updates went in for this functionality after the work you put in, but as it stands, I don't believe there's anything outstanding that needs addressing for this particular issue.

However, there's always things! :) We're working on the [v0.2.0 milestone](https://github.com/owncast/owncast/milestone/26) right now, and there's a couple web frontend items that would make a big impact if you're interested.

The biggest one is the design of the followers is looking to be improved. You can look at that here: https://github.com/owncast/owncast/issues/3248

Another is the embed offline state has some potential overflow issues: https://github.com/owncast/owncast/issues/3683

Thanks for checking in, let me know if you have any thoughts or you want to look through other issues that you might be interested in.

PR #3496, opened 2023-12-23
Refactor Ip address parsing to support ipv4 and ipv6

### LINKED ISSUE #3419: Bug: "Error: too many colons in address"

```
ERRO[2023-11-08T17:59:12-08:00] address ::ffff:127.0.0.1: too many colons in address
```

I'm able to reproduce this by accessing localhost. My guess is this is another issue from the changes in #3349.

armadi1809:
Resolves #3419 
@gabek Sorry for taking a little while to get to this.  Here is a brief explanation of what was going on and how I addressed the issue: 

The `net.SplitHostPort` function does not have support for ipv6 addresses unless it is in the `[Host]:Port` format. Therefore, to go around this, I had to add a small check to determine which forma the input address is in (ipv4 or ipv6) and then if it's in the ipv6 format, check whether or not a port was provided (this is done via checking the existence of an open square bracket in the address), and if it is, pass the address to the `net.SplitHostPort`, otherwise, return the original address. 

I also did some refactoring of the `GetIPAddressFromRequest` function and it is now much more concise and readable. 

I finally added more tests for the different kind of scenarios that this function could face. 

Please let me know if you think there's something else that needs to be done to resolve this issue. I apologize for the incomplete/ugly implementation a while back as I was just starting with golang and the open source world in general 😅

### DIFF
--- a/utils/clientId_test.go
+++ b/utils/clientId_test.go
@@ -7,21 +7,82 @@ import (
 
 func TestGetIPAddressFromRequest(t *testing.T) {
 
-	expectedResult := "203.0.113.195"
+	expectedIpv4Result := "203.0.113.195"
+	expectedIpv6Result := "2001:9e8:5221:4201:5221:7dff:fe53"
 
 	request := httptest.NewRequest("GET", "/", nil)
+
+	//Test ipv4 address without X-FORWARDED-FOR header (With Port set up)
 	request.RemoteAddr = "203.0.113.195:41237"
-	//First Test without X-FORWARDED-FOR header
 	result := GetIPAddressFromRequest(request)
-	if result != expectedResult {
-		t.Errorf("Remote address only test failed: Expected %s, got %s", expectedResult, result)
+	if result != expectedIpv4Result {
+		t.Errorf("IPV4 Remote address only (with Port) test failed: Expected %s, got %s", expectedIpv4Result, result)
 	}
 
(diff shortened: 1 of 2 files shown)


utils/clientId_test.go
```diff
@@ -7,21 +7,79 @@ import (
...
+	//Test ipv4 without X-FORWARDED-FOR header (Without Port set up)
+	request.RemoteAddr = "203.0.113.195"
+	result = GetIPAddressFromRequest(request)
+	if result != expectedIpv4Result {
+		t.Errorf("IPV4 Remote address only (without Port) test failed: Expected %s, got %s", expectedIpv4Result, result)
 	}
 
-	//Test with X-FORWARDED-FOR header
+	//Test ipv4 with X-FORWARDED-FOR header (With Port SetUp)
 	request.Header.Set("X-FORWARDED-FOR header", "203.0.113.195:41237, 198.51.100.100:38523, 140.248.67.176:12345")
```
gabek: I didn't notice this before so I didn't ask, but why is the word `header` as a part of the `X-FORWARDED-FOR` header?

armadi1809: Thank you for noticing this. It was resulting in some tests giving false positives too! It should be resolved now.

gabek:
This looks great to me, let's get it in and tested! Thanks so much for your continued help, it's really appreciated 🥇

PR #3508, opened 2024-01-05
fix(video): add eventType support for touch devices

### LINKED ISSUE #3368: Not selectable items in quality settings menu
On touch devices items in quality settings menu is unselectable. They change color to white when I tap on it, but select action is not firing. I try to fix this issue in web/components/video/settings-menu.ts by some changes:

```
let clickevent;
if ('onpointerdown' in window) {
    clickevent = 'pointerdown';
} else if ('ontouchstart' in window) {
    clickevent = 'touchstart';
} else {
    clickevent = 'click';
}

...

item.on(clickevent, () => { ...
```

It's work fine on Android Phones and Windows Tablet, but not working at IOS devices.

varungujarathi9:
fixes #3368 

The select event was not being fired on touch devices since we were using a callback to register only the `click` event. For most of the touch devices we need to register some sort of touch event like `touchstart` or `touchend`.

So now we have added `touchend` support to register touch events.

```
let clickEvent;
if ('ontouchstart' in window) {
    clickEvent = 'touchend'; // Use touchend event for touch devices
} else {
    clickEvent = 'click'; // Use click for all other devices
}
```

Tested these changes on LambdaTest iPhone 15 device in Real-Time

### DIFF
--- a/web/components/video/settings-menu.ts
+++ b/web/components/video/settings-menu.ts
@@ -58,10 +58,17 @@ export function createVideoSettingsMenuButton(
         });
       });
 
+      let clickEvent;
+      if ('ontouchstart' in window) {
+        clickEvent = 'touchend'; // Use touchend event for touch devices
+      } else {
+        clickEvent = 'click'; // Use click for all other devices
+      }
+
       for (let i = 0; i < items.length; i += 1) {
         const item = items[i];
         // Quality selected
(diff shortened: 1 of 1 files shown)


gabek:
Awesome, this is a great fix. Thanks so much for working on this! :1st_place_medal:

PR #3516, opened 2024-01-12
Fixed page size stale state on LogTable component

### LINKED ISSUE #3511: Changing Log-Entries-Per-Page in Admin Console Doesn't Work

At the bottom of the page is a drop-down menu with a few different options for number-of-entries-per-page: 10, 20, 50, 100. Clicking the various options does nothing. Here's a 30-second video demonstrating the problem: https://www.youtube.com/watch?v=p86BobWkUfY

armadi1809:
Closes #3511 

@gabek, In fact, I think I was right about the stale state causing the page size to revert back to whatever was being passed as props. I therefore added a piece of state to the `LogTable` component that gets updated when the state changes and this fixed the issue. 

Another alternative is to just change the table component to be called as follows. 

`<Table
        size="middle"
        dataSource={logs}
        columns={columns}
        rowKey={row => row.time}
        pagination={{
          defaultPageSize: pageSize,
          showSizeChanger: true, 
          pageSizeOptions: [10, 20, 50, 100]
        }}
      />`

However, this will cause the size changer to appear right away even if there is only a few logs as opposed to it appearing automatically after a certain amount of logs have been generated (which is the current behavior)

Let me know which one you prefer and I can adjust the code accordingly.

### DIFF
--- a/web/components/admin/LogTable.tsx
+++ b/web/components/admin/LogTable.tsx
@@ -1,7 +1,7 @@
-import React, { FC } from 'react';
+import React, { FC, useState } from 'react';
 import { Table, Tag, Typography } from 'antd';
 import Linkify from 'react-linkify';
-import { SortOrder } from 'antd/lib/table/interface';
+import { SortOrder, TablePaginationConfig } from 'antd/lib/table/interface';
 import format from 'date-fns/format';
 
 const { Title } = Typography;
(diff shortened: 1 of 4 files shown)


armadi1809:
@gabek Any feedback on this?

gabek:
> However, this will cause the size changer to appear right away even if there is only a few logs as opposed to it appearing automatically after a certain amount of logs have been generated (which is the current behavior)

I think either is fine. If it's annoying one way or the other it can always be changed.

armadi1809:
Sounds good. Then I think this is good to go with the way it is now. Thanks for the feedback.

gabek:
Thanks so much for working on this! It all completely makes sense. This has probably never worked correctly before, so it's a nice fix :100:

PR #3534, opened 2024-01-24
Modified rate limit conditions to exclude moderators

### LINKED ISSUE #3533: Chat moderators should be exempt from message rate limiting

See #3523 for discussion.

armadi1809:
Resolves #3533


armadi1809:
@gabek, I don't know if this is all we need to do to achieve this but from searching through the code base, it looks like it. Please let me know if I missed something.

gabek:
That should be it! Thanks so much!

PR #3582, opened 2024-02-16
fix: remove autoscroll to readwrite chat embed input field

### LINKED ISSUE #3500: When embedding Owncast chat via iframe the page jumps to where the iframe is

Hello.
My Englisch ist not Good. 

If you add the chat to your own website, the website always scrolls down to where the text input field is. 2 videos show this here. If I leave out the chat and only have the video player, there is no problem.  So our users don't find the login area because they don't expect to scroll up. That's what you do for downwards.

 Text error? it is deepl translator :-)
 
 https://www.youtube.com/shorts/m-6CIvgTFcY?feature=share
 
Post for Owncast v0.1.2

DarkFalc0n:
This fixes #3500, 
The issue was mainly with this part of code at `/web/components/chat/ChatTextField/ChatTextField.tsx`

```js
// Focus the input when the component mounts.
  useEffect(() => {
    if (!focusInput) {
      return;
    }
    document.getElementById('chat-input-content-editable').focus();
  }, []);
```

Added a `preventScroll: true` option to `focus()`


gabek:
That's crazy. You'd think an `iframe` would be completely sandboxed. I could see it scrolling within the frame itself, but not being able to scroll the entire page.

Thank you for tracking it down!

PR #3612, opened 2024-03-02
set content-type for customjavascript

heavygale:
Currently (v0.1.2) the resource /customjavascript is served with "content-type: text/plain; charset=utf-8" and is therefore blocked when combined with "X-Content-Type-Options: nosniff", which may be added by a reverse proxy. So we should set the correct content-type header when we deliver this resource. 

I'm a complete stranger to GO, therefore please review/validate/test before merging.


gabek:
Thank you!

PR #3630, opened 2024-03-14
Mute the 'stream ended' clip's audio

### LINKED ISSUE #1965: Add an option to change the end stream clip

vivlim:
The 'stream ended' clip plays at the at the end of every stream broadcast using owncast. It currently contains audio that peaks at -7.1db. (according to ffmpeg's volumedetect audio filter)

This can result in a bad experience for viewers if the stream that they were just watching had a much lower average volume, and they had turned up their speakers or headphones to compensate. In extreme cases this could theoretically cause harm to viewers and/or their equipment.

As an admin running owncast, there is no way to remove this audio *except* for replacing the file. If you do replace the file, you need to notify your viewers to clear their browser caches if they have ever seen the clip, because offline.ts has a cache-control max-age header specifying 365 days. Any attempt to address the caching of the previous version of this clip is out of scope of this PR.

This issue is discussed in more detail in #1965.

Unlike my previous attempt in #3332, which removed the audio track, this PR *mutes* the audio.

Specifically, I used this ffmpeg command, after renaming `offline.tsclip` locally to `offline.ts`:
```
ffmpeg -i offline.ts -filter:a "volume=0.0" output.ts
```

There are no other modifications to the clip.


gabek:
Unfortunately, when I play a stream and end it, with this new file, it creates a distorted flash of colors. This is when the file, verbatim, is being stitched into the stream. However, when the stream is reset, and this file is put through the transcoder, then it works fine. I'll try to figure out what the problem with the file is.

gabek:
As an aside, it looks like a commit with documentation changes got pulled in somehow?

vivlim:
> Unfortunately, when I play a stream and end it, with this new file, it creates a distorted flash of colors. This is when the file, verbatim, is being stitched into the stream. However, when the stream is reset, and this file is put through the transcoder, then it works fine. I'll try to figure out what the problem with the file is.

Thanks, I don't have an environment set up for testing this change so it's challenging to reproduce.

> As an aside, it looks like a commit with documentation changes got pulled in somehow?

I have no idea why. I only committed once and pushed a branch with that single commit.

mahmed2000:
> `ffmpeg -i offline.ts -filter:a "volume=0.0" output.ts`

This will also re-encode the video as mpeg2video instead of h264.

PR #3653, opened 2024-03-29
Spec-based API

### LINKED ISSUE #3302: Switch to spec-first API design

Right now, there are two problems:

- The API layer (router + controllers + request/response models) don't really have a solid structure. It's fine, it's easy to work with, but there are no rules.
- The API documentation (OpenAPI Spec) is always out of date. 

I've been having to manually write the OpenAPI spec for every API change, and I often forget, or it just gets deprioritized because it's so labor-intensive. It takes hours to look into every API, and convert it, by hand, to YAML. And honestly, almost nobody gets value out of it. Including me. It's just busy work. Very long busy work.

See the [Open API Spec](https://github.com/owncast/owncast/blob/develop/openapi.yaml).

I kind of had given up and decided I'd just get rid of most of the API documentation since it's wrong, and takes so long for me to build.  See: https://github.com/owncast/owncast/issues/3086

However, the alternative is to do the complete opposite. Be API spec-first and codegen our API layer from the specs. Then there *would* be value from the spec, and it would be 100% correct, all the time.

This would require:

- Still manually writing the JSON for the OpenAPI spec. This still sucks, it's still error-prone, and it's still labor-intensive.
- Removing our entire existing API layer.
- Using codegen to build our APIs directly from the spec.

The pros:

- The documentation would be 100% correct.
- All APIs would be accounted for. Both internal and external.
- The process of adding additional APIs would be automated and always follow the same structure.
- Any mistakes in manually writing the spec will get caught right away because your API would be incorrect and not work.
- Request/response objects can be modeled out and reused for different APIs.

The cons:

- Writing YAML by hand sucks, it takes a really long time, and it's super easy to make mistakes.
- It'll be harder for contributors to the project to understand how the APIs work, and how they're generated.
- Adding a new API will take longer, since you have to update the spec before you can actually work on anything.
- Would need to start over with the spec, since we'd be relying on it, and it's not correct.

Potential projects for API codegen from specs:

https://github.com/deepmap/oapi-codegen
https://ogen.dev/

brendenehlers:
Issue: https://github.com/owncast/owncast/issues/3302


gabek:
Hopefully, I cleaned up the automated API tests flakiness in `develop`. Seems to be running more predictably now.

gabek:
Hurray, the HLS test are passing! :100:

gabek:
In some basic testing, it looks like the path to the emoji images aren't yet loading. But that's the only thing I've run into so far!

gabek:
Merging into develop so we can push forward. Exciting! Thank you so much for all your hard work, @brendenehlers! I'll file a todo for writing documentation around how to add routes and what tooling is required.

PR #3661, opened 2024-04-05
Fix stream chat bug

### LINKED ISSUE #3657: An incorrect message is shown when the stream is live, but chat is disabled

When chat is purposefully disabled, but the stream is live, the "Chat is offline" message says "Chat will be available when chat stream is live". This is incorrect. This upper right chat notice should not be displayed at all if chat is not enabled on the server, if the stream is live or not. No need to notify a user about a service that isn't enabled, especially incorrectly.

Selection_024

BreakTos:
If the stream is running and chat is disabled appropriate message would be shown
Fixes #3657


web/components/ui/Header/Header.tsx
```diff
@@ -4,6 +4,7 @@ import cn from 'classnames';
 import dynamic from 'next/dynamic';
 import Link from 'next/link';
 import styles from './Header.module.scss';
+import { DisabledChat } from '../../chat/ChatTextField/ChatTextField.stories';
```
gabek, line 7: Looks like you accidentally imported a Storybook file here.

web/components/ui/Header/Header.tsx
```diff
@@ -57,9 +59,10 @@ export const Header: FC<HeaderComponentProps> = ({ name, chatAvailable, chatDisa
         <UserDropdown id="user-menu" hideTitleOnMobile showToggleChatOption={canHideChat} />
       )}
       {!chatAvailable && !chatDisabled && (
-        <Tooltip
+        
+       <Tooltip
           overlayClassName={styles.toolTip}
-          title="Chat will be available when the stream is live."
+          title={!online?"Chat will be available when the stream is live.":"Chat disabled by streamer"}
```
gabek, line 65: I think if chat is disabled, we shouldn't say anything. The user can't do anything about it, and it's probably just confusing if they didn't know chat was a feature in the first place.

web/components/ui/Header/Header.tsx
```diff
@@ -57,9 +59,10 @@ export const Header: FC<HeaderComponentProps> = ({ name, chatAvailable, chatDisa
         <UserDropdown id="user-menu" hideTitleOnMobile showToggleChatOption={canHideChat} />
       )}
       {!chatAvailable && !chatDisabled && (
-        <Tooltip
+        
+       <Tooltip
```
gabek, line 63: I'm thinking we shouldn't show any of this if chat is disabled. No tooltip, no text, no anything.

BreakTos: if chat is disabled then chatDisabled becomes true and !chatDisabled becomes false we won't get tooltip or text anyway right ?

gabek: I'd have to look into the specifics behind those booleans. I filed this because I saw a couple of servers with disabled chat that said "Chat is offline". So if it's currently doing the right thing, and nothing is being shown, then nothing needs to be changed. I can do testing this weekend.

PR #3663, opened 2024-04-08
Fix for copying stream key without revealing it only copies asterisks

### LINKED ISSUE #3460: Copying stream key without revealing it only copies asterisks (when using clipboard manager?)

Just installed version 0.1.2 and when configuring OBS I copied the streaming key from the /admin/config/server/ owncast page, but without revealing it first. But only asterisks were copied, not the actual password. I had to click the button to reveal the password, then click the copy button again to copy the actual password.

mahmed2000:
See https://github.com/owncast/owncast/issues/3460#issuecomment-2042610228


gabek:
Works great for me, thanks!

PR #3684, opened 2024-04-11
stream offline status embed updated

### LINKED ISSUE #3682: New offline embed state not handling HTML

image

Jeetch8:
The issue was that we were trying to put  stringified html in the dom, which lead for the current behaviour.
The fix is injecting the html tags/text/string with "dangerouslySetInnerHTML" attribute  (which is actually not dangerous in  
this situation as we control the source)

before ---
Screenshot from 2024-04-11 10-40-51

after---
Screenshot from 2024-04-11 10-47-13


gabek:
Thank you so much!

PR #3687, opened 2024-04-12
Storybook 8 Update

gabek:


mahmed2000:
Do the empty doc pages for screenshots need to exist? Adding the path for the csf js files to `main.js` fixes the build, and the individual Online/Offline pages display fine. The doc page just breaks though.

gabek:
Nope, not at all! They weren't previously there when I set that up, but something changed.

PR #3690, opened 2024-04-13
Change tag to self-closing since the linter complains otherwise

mahmed2000:
Hopefully this fixes the current build errors? Builds locally

### DIFF
--- a/web/components/ui/OfflineEmbed/OfflineEmbed.tsx
+++ b/web/components/ui/OfflineEmbed/OfflineEmbed.tsx
@@ -96,7 +96,7 @@ export const OfflineEmbed: FC<OfflineEmbedProps> = ({
         <Spin spinning={loading}>
           <div className={styles.content}>
             <div className={styles.heading}>This stream is not currently live.</div>
-            <div className={styles.message} dangerouslySetInnerHTML={{ __html: subtitle }}></div>
+            <div className={styles.message} dangerouslySetInnerHTML={{ __html: subtitle }} />
 
             <div className={styles.pageLogo} style={{ backgroundImage: `url(${image})` }} />
             <div className={styles.pageName}>{streamName}</div>


gabek:
Thank you, you beat me to it! :)

And you can ignore the Chromatic errors. I have a support ticket open with them for why it stopped building on their side but builds fine locally.

mahmed2000:
> And you can ignore the Chromatic errors.

@gabek Probably should have checked earlier, but I don't think its on Chromatic. I can't get the components to show up in storybook locally either. I'm getting similar errors, but not exactly the one's on Chromatic. I've tried checking out the last commit that built successfully on Chromatic but even that errors.

I've verified v0.1.3 works, and 9bbb09c doesn't. What all the component errors look like locally:  `_Logo_module_scss__WEBPACK_IMPORTED_MODULE_1__.default is undefined`. No clue why Chromatic didn't pick them up before or why they look different. All have the scss so maybe a dep bump? I can try to figure out which commit broke things.

gabek:
I thought it was a build problem, so once I saw it building without an issue locally, I assumed it was something going weird on Chromatic's side. But now that you looked at it, it's clear it's not a build problem, it's a rendering problem. 

There's a major upgrade to the whole Storybook/Chromatic stack that [I started on the other day](https://github.com/owncast/owncast/pull/3687), but after about an hour I gave up for now. Every time Storybook has a major upgrade, it takes a long time to get working, since they always have substantial breaking changes. But maybe it's not worth fixing v7, and it's worth me continuing investing into trying to get v8 working instead.

mahmed2000:
Probably yeah. Also found the commit. c618fb0a4ec486f992cc61e130c8711f0010c13a works but 4b52800061e7338c080ce48ba80f756e5e351177 doesn't. I doubt actually downgrading would help with builds since it should have broken Chromatic tests earlier than it has. They probably use a different environment for publishing which might be good to keep in mind?

PR #3692, opened 2024-04-14
prevent css-loader v7 renaming default to _default, needed for storybook

mahmed2000:
Storybook works again locally. Might still need this for v8.

See [here](https://www.github.com/webpack-contrib/css-loader/issues/1589) for ref


gabek:
This is fantastic, great find, thank you!

PR #3694, opened 2024-04-14
Fixes for aria-live bugs

### LINKED ISSUE #3664: aria live: user name remains logged in name
## steps to reproduce

* open a stream
* if you're not a regular user, you may have to change your name from the random one, do so. Alternatively, remember your random name for the duration of this experiment
* if the stream has a system message, aria live will contain `<username> said <message>`, where `<username>` is the name you got from owncast or set yourself
* have someone else send a chat message
## expected behaviour

aria live should contain the username of the person who wrote the message

## actual behaviour

the same name as the username you initially started with gets reported for all chat messages

### LINKED ISSUE #3665: aria live: html entities are not decoded
## steps to reproduce

* open a stream
* type something with an apostrophy, such as "don't"
## expected behaviour

aria live should contain "don't"

## actual behaviour

aria live contains "don&#

mahmed2000:
For #3665 and #3664.

Basic testing with orca.


gabek:
This is fantastic, thanks for taking care of these two issues! You're awesome.

Issue #1462, opened 2021-10-08
Allow bitrates higher than 6000kbps (way higher).


hluengas:
Owncast is awesome because its the only easy way for me to stream gamplay at over 60FPS to my friends. However 6Mbps isn't even close to a high enough bitrate for 90FPS 1440p or 4K streaming even with the highest quality CPU ussage. Why not make the bitrate slider cap out at something really high-end like 50Mbps (or if you think people will have a bad experience when they set that without actually having the bandwidth, then at least provide a manual entry method). I tried to change this myself and build from source, but I believe this is controlled by the owncast-admin side of things and I can't figure out how to pull in changes to that using the provided Dockerfile.

gabek:
You're right, the reason it's not a freeform text box, or a slider that goes to crazy values is because people will put in crazy values.

Owncast aims to be a sane solution for people looking to self-host their live streams. Most of these people self-hosting won't have a super server or infinite bandwidth. Sure, some will, like you.  But instead what will happen is people will *think* they have a super server with infinite bandwidth, everything will go horribly, and they'll delete Owncast and tell everyone how it sucks. We already see this now with other values. Because we offer it people believe it will work for them.

I suggest using the APIs to set whatever crazy values you want, even though they're not supported in the admin. https://owncast.online/api/latest/#tag/Admin/paths/~1api~1admin~1config~1video~1streamoutputvariants/post

If this is something that becomes a normal use case we can think of some special, hidden place to go and type in whatever configuration you want, but as it stands any setting that's "use at your own risk" doesn't work. The risk is ours by offering it.

ma-karai:
I understand that the unique selling point of owncast is that you can run this even on a raspberry pi, and its great.  However, why not cater to people who have high-end infrastructures. You could easily create an option/checkmark, with a question if this is hosted on a high bandwidth infrastructure and then have the slider expand to 20000kbits or whatever.  It does not cost you anything, right ?

Since you linked the api do you have an example curl script to create a stream profile that creates a 12k kbit profile ?

gabek:
It's about guiding the most number of people to the best possible scenarios, and there's multiple things at play for these scenarios.

It's not just about your infrastructure, it's about your viewers. That's the most important part of any of this. If your viewers can't view the stream, then what's the point?

[...]

PR #3721, opened 2024-04-24
Increase max video bitrate and slider marks

JeremyMahieu:
Increases the max video bitrate to 15000 kbps, doesn't require changes in the backend.
Because it's 2024.

image

### DIFF
--- a/web/utils/config-constants.tsx
+++ b/web/utils/config-constants.tsx
@@ -440,7 +440,7 @@ export const FRAMERATE_TOOLTIPS = {
 // VIDEO VARIANT FORM - bitrate
 export const VIDEO_BITRATE_DEFAULTS = {
   min: 400,
-  max: 6000,
+  max: 15000,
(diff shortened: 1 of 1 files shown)


gabek:
Please search for, or open a new an issue for discussion before making a potential change.

The values are what they are for a reason, and they have been asked to be changed a handful of times, and I've been hesitant to do so. And I am hesitant to do so now. https://github.com/owncast/owncast/issues/1462

JeremyMahieu:
Sorry I searched but didn't find it. I guess I didn't search the closed issues.
I need these changes anyway for my own fork so it's not a waste.

15000kbps isn't a crazy value or a manual input however 6000kbps is a crazy low max value now.
Opposite of what you say in the issue, ppl might now start saying owncast sucks because it can only stream at 6000kbps.

gabek:
Moving conversation to #1462.

PR #3729, opened 2024-05-02
Rework stream duration string formatting for the status bar

### LINKED ISSUE #3727: Live streams that are between 24 and 25 hours display the time as "Live for 1 day undefined hours"

I set up a long streaming instance of owncast and after 24 hours the time shown under the window says

> Live for 1 day undefined hours

There are no sourcemaps with the release container image but I did find the raw react code that renders it and the object it's using to render the time is:

```javascript
{ days: 1, minutes: 41, seconds: 11 }
```

I think it is the [`makeDurationString()`](https://github.com/owncast/owncast/blob/b583fa67e1fed697a81547960cf69e23336d4699/web/components/ui/Statusbar/Statusbar.tsx#L22) function.

After 25 hours [`intervalToDuration()`](https://github.com/date-fns/date-fns/blob/fa8046575c30cd255c41a38eab88701159013680/src/intervalToDuration/index.ts#L33) starts returning the hours again and the web page shows:

> Live for 1 day 1 hour

mahmed2000:
Closes #3727

Only change should be `X days undefined hours` -> `X days`.

Tried to keep it consistent with how it is. The `formatDistance` functions only keep the largest unit instead of the largest 2.

### DIFF
--- a/web/components/ui/Statusbar/Statusbar.tsx
+++ b/web/components/ui/Statusbar/Statusbar.tsx
@@ -1,9 +1,8 @@
-import { intervalToDuration, formatDistanceToNow } from 'date-fns';
+import { intervalToDuration, formatDistanceToNow, formatDuration } from 'date-fns';
 import { FC, useEffect, useState } from 'react';
 import dynamic from 'next/dynamic';
 import classNames from 'classnames';
 import styles from './Statusbar.module.scss';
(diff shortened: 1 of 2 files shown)


gabek:
Awesome, thanks @mahmed2000! Would you mind commenting on the original issue so I can assign it? It's nice to assign it so it can be seen later who worked on it.

PR #3750, opened 2024-05-30
Rework utils/restendpointhelper to use the new chi router functionality

### LINKED ISSUE #3712: Remove usage of utils/restendpointhelper.go in favor of the new router

Once https://github.com/owncast/owncast/pull/3653 is in, we can refactor out all the use of the endpoint helper and use the native functionality built into the router to support dynamic path components as arguments.

mahmed2000:
Closes #3712

Doesn't fully remove the file. Replaces the functionality with a single function. Can let it stay there or move it as needed.

Best not to store URL parameters in the request context since chi already does that internally.


gabek:
Makes sense, thank you!

PR #3754, opened 2024-05-31
Implement admin password hashing with bcrypt

### LINKED ISSUE #2495: implement admin password hashing

mahmed2000:
#2495

Works, but not tested extensively. Stream keys still plaintext.

Edit: ~api tests are failing because they still expect the raw password instead of the hash, will get to them in either a few hours or tomorrow~

Edit 2: Tests work now, just grabbed the first package for bcrypt I found, hasn't been updated in a while. Lmk if there's a different package you'd rather use, otherwise this should be good?


test/automated/api/009_configmanagement.test.js
```diff
@@ -166,7 +167,10 @@ test('verify default admin configuration', async (done) => {
 	expect(res.body.yp.enabled).toBe(defaultYPConfig.enabled);
 	// expect(res.body.yp.instanceUrl).toBe(defaultYPConfig.instanceUrl);
 
-	expect(res.body.adminPassword).toBe(defaultAdminPassword);
+	// expect(res.body.adminPassword).toBe(defaultAdminPassword);
```
gabek: I don't remember why some of these other assertions are commented out, but since the old admin password ones are no longer valid, they should probably just be removed completely.

gabek:
All looks great in my usage testing! Though I went to use the updated admin to try to change the password there and I ran into #3775, so I haven't yet been able to test that because of that bug. The only change there, though, would be not displaying the password in the admin UI.

gabek:
When using another browser I had no problems, so something must have been cached.

gabek:
I'll have to fix this conflict, but then we can merge this in.

PR #3756, opened 2024-05-31
Update README.md

germainelee:
- Add MIT license badge
- Update TOC
- Update Contributing and add Backer section that lists backers from Open Collective

(PS. practising my html so that they don't go rusty)


gabek:
These are some really nice updates, thank you for taking the time to improve this!

Issue #3422, opened 2023-11-13
Document the licensing of our non-code assets


gabek:
### Share your bug report, feature request, or comment.

Our code is licensed under MIT, completely free and open source. Our non-code assets, however, have not been given a license and creates ambiguity.

In general this project does not have a lot of custom assets, but there's a few. This includes things like our logo, tshirt design, random graphics, and videos.

The best fit is probably the [Creative Commons CC BY-SA license](https://creativecommons.org/licenses/by-sa/4.0/). It states that people can use our assets, derive creative works from them, but credit must be given.

PR #3770, opened 2024-06-04
Add CC license in readme

germainelee:
#3422 

Update readme to include CC license as we have custom assets


gabek:
There's a larger issue filed at #3422 to work on this, and make sure we have actual documentation around our assets and its license. I think a badge can be fine, but I'm not sure it's that useful without something more substantial that can be referenced. If the badge displays it has a license, but there's no actual place that's written down, I don't think that gets us quite where we need to be.

It's also more than just the docs we need to worry about. Things like the logo, custom icons, video content we make, etc. These should all have a license as well.

But I also haven't put any thought into what the format is that makes sense to do such a thing. What do you think?

germainelee:
I'm not sure which custom icons we have. But according to [CC FAQ](https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-databases) website, they said that it is not recommended to use the license on logo. Maybe I understand it wrongly?
> Could I use a CC license to share my logo or trademark?
Creative Commons does not recommend using a CC license on a logo or trademark.

Other than putting it on the readme, can we put it on the footer owncast.online site too so that it is always there? Looking at CC website footer [https://chooser-beta.creativecommons.org/](https://chooser-beta.creativecommons.org) and [wikipedia](https://wiki.creativecommons.org/wiki/Recommended_practices_for_attribution), the CC license notice shows up on every page of the website. That's if we are licensing or marking the whole page or blog, you can paste the code at the bottom of the page.

> If you are licensing or marking one work, paste the code next to it. If you are licensing or marking the whole page or blog, you can paste the code at the bottom of the page.

<blockquote><div><strong><a href="https://beta-chooser.creativecommons.org">Choose a License</a></strong></div><div>Want to license your work with Creative Commons, but not sure where to start, or which license is right for you? Use our license chooser!</div></blockquote>

gabek:
We don't have many custom images, just a couple for social actions like follow, share like: https://owncast.online/components/?path=/docs/owncast-frontend-assets-images--docs. We also have misc images used for random places that aren't used within the product itself.

We don't currently have a trademark on our logo, or even an application for it, so as it stands it's completely unlicensed and ambiguous. It seems like a CC license would at least define what is allowed with it, but I could be wrong.

<blockquote><div><strong><a href="https://owncast.online/components/?path=/docs/owncast-frontend-assets-images--docs">@storybook/cli - Storybook</a></strong></div></blockquote>

germainelee:
So for those created graphics, would need to Add a metadata tag to each graphic asset file, indicating the chosen CC license. Though I'm not sure how to go about doing so in storybook. I can think of marketing assets as one of the items to have CC license, not sure if people would use follow, like, repost graphics on their sites.

Since the icons is primarily used in our product as button icons, I don't think it needs license because I can't think of a use case being used outside of Owncast.
image

This could help with CC license notice: https://chooser-beta.creativecommons.org/

<blockquote><div><strong><a href="https://beta-chooser.creativecommons.org">Choose a License</a></strong></div><div>Want to license your work with Creative Commons, but not sure where to start, or which license is right for you? Use our license chooser!</div></blockquote>

PR #3773, opened 2024-06-07
Build Flatpak Package

### LINKED ISSUE #3512: Flatpak Packaging

I think it would be nice to provide a [Flatpak package](https://flatpak.org/) as an alternative installation method.

For reference, I found Jellyfin very easy to set up: https://flathub.org/apps/org.jellyfin.JellyfinServer

varungujarathi9:
closes #3512
Build Flatpak package of Owncast build


Earthfile
```diff
@@ -142,6 +145,36 @@ docker:
...
+    ARG NAME=custom
+  END
+
+  ENV FLATPAK_NAME owncast-$version-$NAME.flatpak
+  COPY (+build/owncast --platform $TARGETPLATFORM)  /build/flatpak/owncast
+  COPY . /build
+  RUN flatpak-builder --privileged --force-clean --repo=owncast-repo build-dir online.owncast.Owncast.json # build flatpak package from JSON file
+  RUN flatpak build-bundle owncast-repo /build/flatpak/owncast.flatpak online.owncast.Owncast # create a single-file bundle of flatpak package
+
+  SAVE ARTIFACT --keep-ts /build/flatpak/owncast.flatpak owncast.flatpak AS LOCAL flatpak/$FLATPAK_NAME
```
gabek: Since we're saving the rest of the distributable files in `dist` it probably makes sense to put this file in there too.

gabek:
I'm surprised this was working for you. I had to make a few changes to get it to work. Could we compare to see why it was working for you but not for me? See my recent commit to see the changes I had to make.

Also `--allow-privileged` needs to be used with the `earthly` command for these changes to work.

varungujarathi9:
Hmm that's strange, but the current changes are working for me. We can go forward with these.

gabek:
Now that we have some actual Flatpaks I'll do some testing of actually using them.

online.owncast.Owncast.json
```diff
@@ -0,0 +1,21 @@
...
+			"build-commands": ["cp ./owncast /app/"],
+			"sources": [
+				{
+					"type": "file",
+					"path": "/build/flatpak/owncast"
+				}
+			]
+		}
+	],
+	"finish-args": ["--socket=x11", "--share=network"]
```
gabek, line 20: When installing this it required X11 permissions, even though this is server software and not going to necessarily be installed on a desktop machine. Is X11 permissions required?

image

gabek:
- Version number isn't set, it thinks it's just a development build

image

- Transcoding isn't working. My guess is a flatpak has no access to ffmpeg?

image

- Seems to be a problem even running the application with the io read/write error.
- I'm not sure where the data directory is located.

image

mahmed2000:
X11 permissions aren't required and that arg can be omitted, but the runtime adds x11 libs to the build anyway along with wayland and others. https://docs.flatpak.org/en/latest/available-runtimes.html#freedesktop, https://docs.flatpak.org/en/latest/sandbox-permissions.html#standard-permissions

It has ffmpeg and access to it, however I'd guess it wouldn't be compiled with codec support. There's a dedicated runtime extension `org.freedesktop.Platform.ffmpeg-full` for all codecs which would include x264.

The IO error is ffmpeg's input stdio pipe closing, probably because it can't handle x264 due to ^ and crashes.

For the data folder I'd check
- ~/.var/app/
- ~/.local/share/flatpak/
- /var/lib/flatpak/
Probably best to add a dedicated directory instead `~/.owncast/data`?

<blockquote><div>Flatpak</div><div><strong><a href="https://docs.flatpak.org/en/available-runtimes.html">Available Runtimes</a></strong></div><div>This page provides information about available Flatpak runtimes. It is primarily intended as information for application developers and distributors. There are currently three main runtimes availab...</div></blockquote>
<blockquote><div>Flatpak</div><div><strong><a href="https://docs.flatpak.org/en/sandbox-permissions.html">Sandbox Permissions</a></strong></div><div>One of Flatpak’s main goals is to increase the security of desktop systems by isolating applications from one another. This is achieved using sandboxing and means that, by default, applications tha...</div></blockquote>

varungujarathi9:
Hey @mahmed2000 
I made the changes to support ffmpeg runtime extension by doing the following
```
"add-extensions": [
  {
    "org.freedesktop.Platform.ffmpeg-full": [
      {
        "version": "23.08",
        "directory": "lib/ffmpeg",
        "add-ld-path": "."
      }
    ]
  }
]
```
But it doesn't seem to work. 

I also tried changing the `runtime` to `org.freedesktop.Platform.ffmpeg-full`.

How do I correctly add the `ffmpeg` support?

mahmed2000:
I'm not entirely sure, I don't particularly use flatpak.

You could try just running the ffmpeg command in the flatpak to check the installed codecs. `flatpak run --command=ffmpeg <flatpak.name.here>`. The banner output should show the codecs supported, specifically `--enable-libx264`. You could also inspect the flatpak for the `libx264.so` library file.

I'm guessing either that ffmpeg still doesn't have libx264 support, in which case it needs to be done through a module source instead, see [OBS studio's flatpak for an example](https://github.com/obsproject/obs-studio/blob/69d274074e8d1a6bdcf6777a0feba8596a8ee9b0/build-aux/com.obsproject.Studio.json).

Or it can't find the libx264 library which isn't included, in which case that needs to be added as well.

PR #3780, opened 2024-06-11
Update Antd to v5.12+

### LINKED ISSUE #2359: Update to Ant Design v5

mahmed2000:
Closes #2359 

See https://github.com/owncast/owncast/issues/2359#issuecomment-2156801537 for additional context
- Some of the existing less variables did not have a clean mapping
- Some of the CSS selectors in `styles/ant-overrides.scss` need to be adjusted to match specificity.
- A number of things were changed that make things look very weird, like the global error modal. Padding, etc needs to be overridden.
- Its not completely clean, and computes the theme twice on the main page (on load to match admin and on `Theme` inserting into the DOM. But antd theming works at least.

Will make changes over time, but doable from here.


mahmed2000:
Storybook is going to be weird in general, as are embeds if I had to guess.

Most of the general global rules in the v4 less file have been moved to the `.ant-layout` class instead of `body` or `html`, which means those won't apply to most rules unless wrapped in a layout (like the page and not storybook/embeds) or they're set manually as global rules. Ideal situation would be wrap everything in at least the antd layout to maintain old consistency.

There's also discrepancies in, for example, the chromatic snaphot for "Custom Page Content" and the actual published storybook. It shows the link decorations but none are there in the actual storybook or the actual page.

Some of the mappings still need clarification. Hovers are weird. Most of them follow the `--primary-hover-color` which is blue, but some (like the active tab) follow the intended(?) light purple like v4.

gabek:
I'm really impressed it's working as well as it is, nice job! We can certainly wrap layouts if that's what's needed.

web/styles/antd-default.js
```diff
@@ -0,0 +1,37 @@
...
+
+const lodash = require('lodash');
+
+const AntdStaticTokens = {
+  fontFamily:
+    "'Inter', system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'",
+};
+
+const AntdDefaultCustomTokens = {
+  colorLink: '#6544e9',
```
gabek, line 11: Can we use `getComputedStyle` here to look up CSS variable values from Javascript instead of the hardcoded values? Even if we did generate this using style-definitions, it would still be hard-coded values unless we supported a dynamic lookup.

web/styles/antd-default.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+  borderRadius: 5,
+  colorPrimary: '#6544e9',
+  colorPrimaryHover: '#2386e2',
+  colorPrimaryActive: '#7a5cf3',
+  colorBgBase: '#e2e8f0',
+};
+
+const AntdCustomTokens = (customTokens = AntdDefaultCustomTokens) => {
+  return lodash.merge({}, AntdStaticTokens, customTokens);
```
gabek: Can probably import just `merge` from lodash instead of all of `lodash`.

web/styles/antd-default.js
mahmed2000: `getComputedStyle` only exists as a window function which means runtime only. I set it up with static values for the initial page load. At build it extracts those for the initial style via `StyleProvider`.

Could ignore this for the main page, but the admin will have issues. I can't remember if just a basic `useEffect` triggers it after the CSS is processed or before. I'm using the `Theme` component's inserted style as the trigger, but it doesn't exist in the admin pages, nor the storybook stuff. Could forgo the static values for the antd defaults if there's some trigger to use it with, but the initial load might look off.

There's also a potential way to simplify all this by using a CSS file instead of this js, there would just be a ton to go through and would need to figure out the selector priority:

```
body .css-var-antd-hash.ant-component {
  --antd-component-color-primary: var(--theme-color-primary);
  ...
}
```

Sorry, I kind of started procrastinating doing this.

gabek:
Not stale.

mahmed2000:
Things to note so far:

- ~button widths are going to be larger since antd now defaults to `inline-flex` instead of `inline-block`. Can either set that in the button config (has to be done for each), or override it, or use the new width.~
- ~icon only buttons are off because for some reason antd now only sets the `width` to 40px, but does nothing about the height.~
- just changing the font should not have fixed _all_ of that. Everything just works fine now and I don't know why.
- Stories cannot be wrapped in `Layout` without causing drastic changes. Adds screen-wide backgrounds to all components and stretches things like buttons to full width. Best to just replicate the old general CSS attributes in the body or html blocks.

Edit: Chromatic test went better than I expected at least. ~Still a large amount of weird stuff to fix in the admin~ ~Seems fine now for some reason~ nvm, I am dumb. That and some areas need checking for theming. And need to rebase to develop.

mahmed2000:
Things needing doing:

General:
- [ ] Some styling changes from antd's previous defaults. Padding, margins etc
- [ ] numbers have some more concise kerning for some reason
- [x] hovers need color adjustment
- [x] Input fields darken to the color of the background now. `.ant-input-outlined:hover`
- [x] Modals headers now have a bottom margin of 8px, can look like a colored line with theming.

Main page:
- [ ] antd now adds a 1px border to `.ant-avatar` Logo shifts right slightly
- [ ] text margins are different in the bottom container

Chat:
- [x] Name change modal: color is invisible
- [ ] Authenticate modal: Deselected tab does not follow theme colors. css class has changed. Looks better now (?) unless modals are to be themed as well fully.
- [ ] Authenticate modal: `ant-input-group-addon` is not being themed.

Admin general:
- [x] The admin menu does not follow the `:active` override in `ant-overides.scss`
- [ ] Admin menu has less `padding-bottom`
- [x] "Compose Post", the margin-left rule for spans in buttons next to icons has been removed.
- [ ] `.ant-card-meta-title` now have a font-weight of 600 instead of 500.
- [ ] antd layout footer is darker

Admin: Chat & Users:
- [x] Messages: The `.ant-btn-round.ant-btn-sm` rule of height: 24px is now just `.ant-btn-sm`. It has lost specificity to `ant-btn-primary` at 2rem.
- [ ] Messages: currently sorting columns now have a highlight color.
- [x] Messages: `.ant-table-row-selected` css rules are weird and don't show up in firefox. Do not follow owncast's overrides.
- [ ] The no data icon is cleaner? but different.
- [x] List selections are not colored

Admin: Config:
- [x] Sliders are not colored correctly.
- [x] Number inputs do not get colored correctly.
- [ ] Table footer backgrounds have a different color
- [x] Checkboxes are also incorrectly colored

mahmed2000:
Should be mostly done. I don't think what's left is really worth trying to be exact on.

There's now both the scss rules and the ConfigProvider both trying to apply styles, and its making everything very scuffed.

The css lint error is from before this PR. There's 3 (not 2) different blocks for `&:focus` there, and each sets a different `border-color`. Not sure what to do about that.

mahmed2000:
Component errors exist even on develop:

https://owncast.online/components/?path=/story/owncast-layout-main--offline-desktop

All relate to i18n, so I'm assuming this started when the translation stuff was merged in.

#3999 strikes again.

With antd, at this point, it'd be better to split the update into 2. Merge all this into develop/a separate branch like with localization. Then fix up changes and/or start using the new antdconfig component for styling instead of the random css in ant-overides.scss that may or may not have priority.

<blockquote><div><strong><a href="https://owncast.online/components/?path=/story/owncast-layout-main--offline-desktop">@storybook/core - Storybook</a></strong></div></blockquote>

gabek:
The error seems to point to `useRouter()`, something that has been in the codebase and working for many years. But this seems to be a known problem with Next and testing https://nextjs.org/docs/messages/next-router-not-mounted, saying to just use [next/compat/router](https://nextjs.org/docs/pages/api-reference/functions/use-router#the-nextcompatrouter-export) instead.

<blockquote><div><strong><a href="https://nextjs.org/docs/messages/next-router-not-mounted">`NextRouter` was not mounted</a></strong></div></blockquote>

mahmed2000:
Config problem with the stories. next-export-i18n v2 uses the pages router. 7028276 added `appDirectory: true` to a bunch of stories. That makes storybook mock the navigation router instead of the pages one that's expected. Also explains why the read-write chat embed didn't error out even using the Header component since it mocks uses the default pages router without that config.

https://github.com/owncast/owncast/actions/runs/13160274359/job/36726999154#step:3:66

Same as I pointed out before, the storybook generated on my fork's develop branch, and not the antd branch. The errors are gone locally.

PR #3796, opened 2024-06-27
Override default videojs volume panel width

### LINKED ISSUE #3497: Usability: Volume Bar Fades Too Fast

Owncast's video player is great and slick-looking, but I noticed today while watching Hatnix that changing the volume is pretty difficult.

When you hover over the video, the volume bar isn't accessible (why not? Playing with stream volume is a pretty common task, at least for me, as I do things in the background).

When you hover over the 'volume' icon, the volume bar fades into view. The real problem is that, if your mouse cursor floats off of the volume cursor / bar for even a split-second, the bar fades away and is inaccessible. In order to lower or raise the volume, you have to very carefully move your cursor exactly rightward without ever drifting off of the volume bar. It's like playing a game of Operation.

This can't be intentional. The solution would be to create a grace period, just like how moving your cursor off of the video doesn't immediately hide the player UI. I personally think a part of this solution should involve considering whether or not the volume bar should be hidden in the first place. It's a slick animation to see it fading in-and-out for sure, but it's not like it's saving much space in the UI and I think it's a little bit confusing, since clicking a volume icon will usually mute something (as it does here) and that's not what I want to do, so why would I move my mouse there except for, "wait, where's the volume bar? Oh, here it is."

Just my two cents! But either way, yeah I would say at the very least there needs to be a delay between moving your mouse off of the volume bar and it fading away.

EDIT 1: Just did some more playing with it. The volume bar fades away even if I move perfectly along the bar, going rightward (from the volume icon to the bar) if I go too fast. Maybe because, if I go fast enough, the cursor position maybe doesn't stroke along every position of the bar?

EDIT 2: Something is up with the volume bar. I don't think that it's only that it's fading too fast, actually (although that is still a problem). It seems like I can't touch anywhere about 55% volume without it fading away, no matter how careful I am? Here's a video: https://www.youtube.com/watch?v=dLuPk-BP4Mk

mahmed2000:
Closes #3497.

Fixes the fade out for > 55% volume. Height may or may not also need to be changed.

### DIFF
--- a/web/components/video/VideoJS/VideoJS.scss
+++ b/web/components/video/VideoJS/VideoJS.scss
@@ -9,6 +9,12 @@
     flex-shrink: 0;
   }
 
+	.vjs-hover.vjs-volume-panel-horizontal {
+		// vjs default is 5em icon+padding + 5em slider.
+		// should match the value above.
+  	width: calc(5em + 100px) !important;
+	}
+	
   .vjs-menu li {
     color: var(--theme-color-components-text-on-dark);
   }


gabek:
Works great, thank you!

PR #3810, opened 2024-07-03
Update video passthrough warning banner styling

### LINKED ISSUE #3798: Admin video passthrough banner: Low contrast between the text color and the background

I have poor eyesight and the red text on grey background causes eye strain as I try to decipher the words. It seems like an important note, but the low-contrast element decrease legibility. See image below.

### Steps to replicate:

1. Log in to admin
2. Navigate to configuration > Video
3. Add a new variant in stream output
4. Toggle "Use Video Passthrough?" to ON
5. Click "Yes" when prompted
6. Scroll up and observe the note

image

### Expected behvior:
In order to meet the guidelines at Level AAA, element should be White text `color: var(--theme-color-palette-4);` on dark background using theme color `background-color: var(--theme-color-palette-1)`, shown image below

image

germainelee:
Update styling for better contrast and legibility.

Closes #3798

### DIFF
--- a/web/public/styles/admin/config-video-variants.css
+++ b/web/public/styles/admin/config-video-variants.css
@@ -10,10 +10,10 @@
 .config-variant-form .passthrough-warning {
   text-align: center;
   padding: 1em;
-  color: var(--theme-color-palette-error);
+  color: var(--theme-color-palette-4);
   font-size: 0.88em;
   font-weight: 500;
-  background-color: var(--theme-color-palette-10);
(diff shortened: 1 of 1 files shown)


germainelee:
failed checks for my first .css edit😞

gabek:
The failure is because it can't find the hash that it was originally forked from. I'm not sure why that might be the case, unless there was some weird forking going on. But I wouldn't worry about it, it hasn't happened to anybody else previously.

Thanks!

PR #3818, opened 2024-07-05
Adjust CleanupDirectory function to support tmpfs mounts.

LogalDeveloper:
On my Owncast instance, I wanted to reduce the number of writes to my server's solid state drives caused by constant writes to the "hls" folder during live streams. To accomplish this, I set up a tmpfs mount on the "hls" folder to use a RAM-based filesystem rather than the normal SSD-backed filesystem. Below is the relevant snippet from my Docker compose file:

```yml
tmpfs:
  - /app/data/hls:mode=755,uid=101,gid=101
```

However, Owncast currently clears out the "hls" directory by deleting and recreating it entirely. While simple and effective, this method isn't compatible with mounts at that location. As a result, when Owncast tries to clear the "hls" directory on startup, it fails with the following error:

```
owncast    | time="2024-07-05T00:18:55Z" level=info msg="Owncast v0.1.3-linux-64bit (6ae3167b61d8485d3f269bd422f0c1a1f072ea5c)"
owncast    | time="2024-07-05T00:18:55Z" level=fatal msg="Unable to remove directory. Please check the ownership and permissions unlinkat data/hls: device or resource busy"
```

To work around this issue, I modified the `CleanupDirectory` function to iterate through and delete each file and folder within the "hls" directory individually. This approach maintains compatibility with mounts and prevents the "device or resource busy" error while still achieving the desired cleanup.

I have tested this change on my main instance with a tmpfs mount on the "hls" folder, as well as on a brand new normal install without any custom mounts.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -306,17 +306,17 @@ func VerifyFFMpegPath(path string) error {
 func CleanupDirectory(path string) {
 	log.Traceln("Cleaning", path)
 	if err := os.MkdirAll(path, 0o750); err != nil {
-	    log.Fatalf("Unable to create '%s'. Please check the ownership and permissions: %s\n", path, err)
+		log.Fatalf("Unable to create '%s'. Please check the ownership and permissions: %s\n", path, err)
 	}
 	entries, err := os.ReadDir(path)
 	if err != nil {
-	    log.Fatalf("Unable to read contents of '%s'. Please check the ownership and permissions: %s\n", path, err)
+		log.Fatalf("Unable to read contents of '%s'. Please check the ownership and permissions: %s\n", path, err)
 	}
 	for _, entry := range entries {
-	    entryPath := filepath.Join(path, entry.Name())
-	    if err := os.RemoveAll(entryPath); err != nil {
-	        log.Fatalf("Unable to remove file or directory contained in '%s'. Please check the ownership and permissions: %s\n", path, err)
-	    }
+		entryPath := filepath.Join(path, entry.Name())
+		if err := os.RemoveAll(entryPath); err != nil {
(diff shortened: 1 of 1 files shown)


utils/utils.go
```diff
@@ -302,14 +303,21 @@ func VerifyFFMpegPath(path string) error {
...
+	    log.Fatalln("Unable to create directory. Please check the ownership and permissions", err)
+	}
+	entries, err := ioutil.ReadDir(path)
+	if err != nil {
+	    log.Fatalln("Unable to read contents of directory. Please check the ownership and permissions", err)
+	}
+	for _, entry := range entries {
+	    entryPath := filepath.Join(path, entry.Name())
+	    if err := os.RemoveAll(entryPath); err != nil {
+	        log.Fatalln("Unable to remove directory. Please check the ownership and permissions", err)
```
gabek: This error message should probably be updated to clarify that it's unable to remove the contents of the directory, and not the directory itself.

LogalDeveloper: Good point! I have adjusted the error message to correctly state what it is doing and what it encountered an error on. While I was at it, I added the original path to the outputs for all possible errors in this function. As far as I can tell, this function is currently only used for the "hls" folder, but I figured providing the original path would be helpful if it gets used more in the future.

Below is sample output from this patch with intentionally incorrect filesystem permissions:

```
FATA[2024-07-06T00:10:45Z] Unable to remove file or directory contained in 'data/hls'. Please check the ownership and permissions: unlinkat data/hls/0: permission denie
```

```
FATA[2024-07-06T00:11:12Z] Unable to read contents of 'data/hls'. Please check the ownership and permissions: open data/hls: permission denied
```

gabek:
I'll do some testing this weekend, and then we can get this in.

gabek:
All seems to work great! Thanks for suggesting the improvement!

Issue #3857, opened 2024-07-30
Player stream time can show "undefined" 


gabek:
### Share your bug report, feature request, or comment.

I'm not sure exactly in what case this happens, but we should be able to easily reproduce it in Storybook by trying different time combinations.

![image](https://github.com/user-attachments/assets/d2f50f2f-ade8-4f3d-a250-2d989b3fc1b7)

cnfairall:
Hi! I would like to take this on :)

gabek:
Thanks so much, @cnfairall!

cnfairall:
@gabek I'm stuck in setting up dev environment. sent you msg on rocket chat as well. I have go installed but I don't think I managed to install anything else, not sure how
![Screenshot (10)](https://github.com/user-attachments/assets/c238a865-bcaf-4e1d-a91d-7a98890f5eff)

cnfairall:
hey, I don't think I'll be able to do this ticket. I tried for over a week to set up my environment but it proved too complicated for my knowledge level. sorry about it!

zockicookie:
@cnfairall I had some problems with the first setup as well. You will need to set up ffmpeg and gcc. I managed to avoid the cgo error by calling "CGO_ENABLED=1 go run main.go" when powering up the backend.

zockicookie:
@gabek since this is open again, I would love to try my luck with this one.

gabek:
Sure, thank you @zockicookie!

mahmed2000:
Does this happen on builds after #3727 was fixed?

gabek:
@mahmed2000 I thought it was, but that might be incorrect. I think it's good to get this test in that @zockicookie is working on, and then we can close this since it hasn't been able to be reproduced, and I could be wrong on the version(s) that I saw this on. Thanks for chiming in.

RonMordo:
Hi, if availble i would like to try to solve the problem.

gabek:
We actually think it's probably fixed, and it might be an outdated version that this was seen in. I'm going to go ahead and close this and if it happens again after `v0.2.0` is released I'll go ahead and reopen it. Thanks for wanting to help, do feel free to chime in on another issue if you'd like!

PR #3912, opened 2024-09-05
Statusbar-Duration-FallbackValues

zockicookie:
Regarding issue #3857 

Since I could not reproduce the error within an hour of testing on the storybook, I figured implementing fallback values for values of the passed DurationInterval was the best approach to this.

I tested all cases on the storybook and did not encounter an error.


gabek:
My concern with this is it would end up displaying incorrect information, and it would look broken differently as the number stays as zero instead of counting up. So while zero does look better than `undefined`, it's still wrong in a different way.

gabek:
I wonder if a good way to try to reproduce this would be to write a test for `makeDurationString`, and loop over all possible time values within a couple days until the output has the string `undefined` in it.

zockicookie:
Hi, I will comment out the changes in the branch then and will look into writing the test. Haven't done that to this day but it sounds fun.

zockicookie:
@gabek so i tried my best with the test but encountered an error with jest-testing. When calling "npx jest" in ./web I got an error regarding importAssertions and importAttributes. I did some research and found out that there seems to be a problem with the combination of the next version I guess? Maybe you could try if jest is still functional with the currently installed packages?

https://x.com/alexanderkondov/status/1826612490509418561
https://github.com/babel/babel/issues/16754
https://github.com/storybookjs/storybook/issues/23063

<blockquote><div>X (formerly Twitter)</div><div><strong><a href="https://x.com/alexanderkondov/status/1826612490509418561">x.com</a></strong></div></blockquote>

gabek:
Yup, this just started happening. So bad timing. I reverted to an older version of Next, but that didn't fix it. I won't be able to look into this until next week, unfortunately. So I'll fix it when I can and then I can look at this PR.

zockicookie:
Alright, just let me know when it's possible to test through jest again. Since I couldn't actually test my tests, some work is still due regarding this issue.

gabek:
Ok, tests should be fixed now!

zockicookie:
Alright, I'm getting back to work then :)

PR #3937, opened 2024-09-26
Storybook 8 fixes

mahmed2000:
#3687

Fixes storybook builds under v8. Rebasing to develop has some more work because of 5765290.


gabek:
Thank you!

PR #3938, opened 2024-09-26
fix: update Alpine base image

fgreinacher:
Alpine 3.11 is out of support since nearly a year.

This also brings the versions in `Dockerfile` and `Earthfile` in sync again.

:hammer_and_pick: with :heart: by [Siemens](https://opensource.siemens.com/)

PR #3943, opened 2024-09-30
chore: use correct manager name in Renovate config

fgreinacher:
The Renovate config change from https://github.com/owncast/owncast/pull/3938 was not entirely correct. This change fixes it as suggested in https://github.com/renovatebot/renovate/discussions/31688. I have verified it via https://github.com/fgreinacher/renovate-earthfile/pull/3.

:hammer_and_pick: with :heart: by [Siemens](https://opensource.siemens.com/)


gabek:
Thanks for the update, @fgreinacher!

PR #3945, opened 2024-10-06
refactor: update vjs beforeRequest to onRequest

### LINKED ISSUE #3874: Vjs Deprecation warning: beforeRequest is deprecated, use onRequest instead

_No response_

unclebinary1001:
this commit fixes #3874


gabek:
Hi there! After making this change, did things still work for you? I'm seeing now that it looks like the functionality is no longer working correctly, as the `cachebust` value is no longer appended to the URL with this change in place.

### Previously (working)

`http://localhost:3000/hls/0/stream.m3u8?cachebust=21409f`

### After change (no longer working)

`http://localhost:3000/hls/0/stream.m3u8`

unclebinary1001:
I will investigate it immediately. I didn't test the network requests. Thank you for catching that.

unclebinary1001:
> I will investigate it immediately. I didn't test the network requests. Thank you for catching that.

it turns out the onRequest hook takes two arguments: xhr (the XMLHttpRequest object) and o (the request options object). and the xhr argument is the actual XMLHttpRequest object that will be used to make the request.

I'll push out an update shortly.

gabek:
Is this working for you? In my testing, I still don't see the cachebust query parameter.

image

web/components/video/VideoJS/VideoJS.tsx
```diff
@@ -36,14 +36,14 @@ export const VideoJS: FC<VideoJSProps> = ({ options, onReady }) => {
     ) {
       (
         videojs.getPlayer(videoRef.current).tech({ IWillNotUseThisInPlugins: true }) as any
-      ).vhs.xhr.beforeRequest = o => {
+      ).vhs.xhr.onRequest = (xhr, o) => {
```
mahmed2000: The code this is referencing is here:

https://github.com/videojs/http-streaming/blob/8456cb3c1c8af207156c0330a9cc2050592e8880/src/videojs-http-streaming.js#L558-L560

Replacing this wouldn't work. This callback needs to be passed as a parameter to the onRequest call instead of replacing it.

PR #3956, opened 2024-10-11
Replace `github.com/shirou/gopsutil/v3` with v4

Juneezee:
This PR does two things:

1. PR https://github.com/owncast/owncast/pull/3925 only added the new v4 dependency into our `go.mod` file. The actual usages in the source code were not updated to use the new v4 version. This PR updates them correctly.
2. Tidy `go.mod` correctly so that we only have two `require` sections: one for direct dependencies, another one for indirect dependencies.


gabek:
Thank you, @Juneezee!

PR #3959, opened 2024-10-12
Fix websocket reconnection

### LINKED ISSUE #3958: Websocket doesn't reconnect

If the websocket connection gets interrupted for any reason, it doesn't seem to ever reconnect.

Steps to reproduce locally:
- Run Owncast and start a stream. I've got OBS streaming a blank scene to it.
- Open http://localhost:8080/ - or wherever your test instance is running.
- Run a proxy on another port: `socat TCP-LISTEN:8081,reuseaddr,fork TCP:localhost:8080`.
- Open http://localhost:8081/ in a private browsing window.
  - You should now have two users in chat, both able to chat to each other.
- Stop (Ctrl+C) and re-start the proxy.
  - The :8080 tab shows that the :8081 user has left the chat.
  - The :8081 tab shows nothing out of the ordinary, but the websocket has been silently disconnected.

In this disconnected state, everything looks normal, except that no further chat messages are received. On latest develop, attempting to send a message simply does nothing, but on v0.1.3, it will look to the user like it succeeds - it's added to the chat log on the client side, but never actually sent, so they're just stuck talking to themselves.

Opening the dev tools and looking at the websocket connection in the network inspector does show it as disconnected, but nothing is logged to the console, including [this error message](https://github.com/owncast/owncast/blob/c712c5f7939b862cbd6e3516addadd93206c4a2c/web/services/websocket-service.ts#L161-L165), which suggests to me that for some reason, the `onError` handler never gets called.

While "kill socat to disconnect chat" is a bit of a contrived example, we've observed this problem live during a friend's streams, where the embedded chat overlay onscreen would sometimes silently stop showing new messages until manually reloaded.

Confirmed on both latest develop (deb938d52127ecf9fc3474a2489323b9d2ee118b, 2024-10-11) and on v0.1.3.  
The web assets were built with `cd web/ && rm -rf node_modules && npm install && npm run build`, before doing a  `go run .`.

liclac:
The `onerror` event is only triggered if the buffer is full while the socket is closed, while the `onclose` event is called for any kind of disconnection: https://websockets.spec.whatwg.org/#closeWebSocket

Fixes https://github.com/owncast/owncast/issues/3958

### DIFF
--- a/web/next-env.d.ts
+++ b/web/next-env.d.ts
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/basic-features/typescript for more information.
+// see https://nextjs.org/docs/pages/building-your-application/configuring/typescript for more information.
--- a/web/services/websocket-service.ts
+++ b/web/services/websocket-service.ts
@@ -56,7 +56,7 @@ export default class WebsocketService {
 
     const ws = new WebSocket(url.toString());
     ws.onopen = this.onOpen.bind(this);
-    ws.onerror = this.onError.bind(this);
+    ws.onclose = this.onClose.bind(this);
     ws.onmessage = this.onMessage.bind(this);
 
     this.websocket = ws;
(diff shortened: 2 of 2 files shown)


gabek (approved):
This works! Thanks for looking at it and making the tweaks.

PR #3970, opened 2024-10-18
fix: fix runtime caching error and adjust rules

### LINKED ISSUE #3967: Service worker throw error Cannot read properties of undefined (reading 'startsWith')
## About the Error
When serving Owncast web production build, I notice there is an error that will thrown from the service worker by inspecting the web console output
Image

I did some research and found out that URL is one of the object property as describe by [RouteMatchCallbackOptions](https://developer.chrome.com/docs/workbox/modules/workbox-core/#type-RouteMatchCallbackOptions), which causing access to `pathname` property will cause error

When register route callback throw error, the rest of the runtime caching that come after will not run

The above error is tested on commit bedbf54 develop branch and error output is observed on google chrome dev tool

## Discussion
By destruct the `url` property or change the functions to access the `url` property before access `pathname` can solve the error, but by doing so will bring back the issue #3097 where Firefox will often need a hard reload to have basic auth prompt appear

I currently come out with an idea where if the main purpose is for caching web components, this seems can be achieve by copy out the config inside `next-pwa/cache` and modify it to not match any route starting with `/admin` or `/api`, while still allow emojis cached by service worker

heongle:
Fix the error where service worker throw error on `urlPattern` callback access invalid property
A modified version of config based on `next-pwa/cache` is passed to `runtimeCaching` option in order to prevent matching of `/admin` and `/api` routes

fixes #3967


web/next.config.js
```diff
@@ -3,27 +3,133 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
   enabled: process.env.ANALYZE === 'true',
 });
 const { PHASE_DEVELOPMENT_SERVER } = require('next/constants');
-const runtimeCaching = require('next-pwa/cache');
 
 const withPWA = require('next-pwa')({
   dest: 'public',
   runtimeCaching: [
     {
-      urlPattern: /\.(?:ts|m3u8)$/i,
```
gabek, line 12: Why was this removed? Previously, the service worker was caching these without forcing it to pull from the network.

heongle: I see, the original code used `NetworkOnly` handler which network request will always happen for this match, so I assumed the intention is to prevent ts and m3u8 being cached

How about I put this matching back using `CacheFirst` handler, so cached response will be used

heongle:
Added runtime cache config for ts and m3u8 files.

For m3u8 file, when directly include all m3u8 files for caching, there is by chances the request to `/hls/{variant}/stream.m3u8` is without `cachebust` param on the very first request when visiting the site will cause the stream to buffer indefinitely. So current config match only the `/hls/stream.m3u8`, `maxAgeSeconds` is used to handle the case where user may change their stream output variant config someday, current `maxAgeSeconds` may not optimal, but let see how it goes

For ts file, `maxEntries` is added to explicitly prevent cache too much old files in cache storage
Seems `offline-v2.ts` will always appended to m3u8 when stream end, so I separated it out as different config to allow it to stay in cache as long as possible

web/next.config.js
gabek: We don't ever want HLS assets to be cached, we want those requests to access the network every time. So `NetworkOnly` makes sense, we don't want `CacheFirst`.

heongle: Thanks for clarification, I will do the changes accordingly

web/next.config.js
```diff
@@ -13,17 +12,128 @@ const withPWA = require('next-pwa')({
       handler: 'NetworkOnly',
     },
     {
-      urlPattern: (url) => { return url.pathname.startsWith("/admin/"); },
-      handler: 'NetworkOnly',
+      urlPattern: /^https:\/\/fonts\.(?:gstatic)\.com\/.*/i,
```
gabek: We don't use google fonts, or anything from google.

heongle: Got it
By the way, most of the config is a copy from the next-pwa runtime cache config, I wonder if there is any config that can to be removed

From what I inspect from browser, seems the following is not used and can also be removed
- `urlPattern: /\/_next\/image\?url=.+$/i` 
- `urlPattern: /\.(?:mp3|wav|ogg)$/i`
- `urlPattern: /\/_next\/data\/.+\/.+\.json$/i`
- `urlPattern: /\.(?:json|xml|csv)$/i`

Do let me know if I didn't got it correct

gabek: That looks right, none of those match any resources we have. Thank you!

web/next.config.js
```diff
@@ -13,17 +12,61 @@ const withPWA = require('next-pwa')({
...
+      handler: 'CacheFirst',
+      options: {
+        rangeRequests: true,
+        cacheName: 'static-video-assets',
+        expiration: {
+          maxEntries: 32,
+          maxAgeSeconds: 24 * 60 * 60, // 24 hours
+        },
+      },
+    },
```
gabek: We don't have any static mp4 files.

web/next.config.js
```diff
@@ -13,17 +12,61 @@ const withPWA = require('next-pwa')({
...
+      urlPattern: /\.(?:css)$/i,
+      handler: 'StaleWhileRevalidate',
+      options: {
+        cacheName: 'static-style-assets',
+        expiration: {
+          maxEntries: 32,
+          maxAgeSeconds: 24 * 60 * 60, // 24 hours
+        },
+      },
     },
```
gabek: What could be the downsides here? I'm a little concerned about a scenario where the Owncast web resources are updated due to a new release, and some old items are pulled from the cache, and some new items are downloaded, causing an application exception. Same with CSS, a mix of old and new could break the UI. I don't know if this could happen or not.

We have HTTP cache header expirations already set on assets, so they should all get expired at the same time. But I don't know how this change would impact any of that.

What are the difference between these manual definitions and the previous `runtimeCaching` that was provided by the library?

From the docs:
> The [stale-while-revalidate](https://web.dev/articles/offline-cookbook#stale_while_revalidate) pattern allows you to respond to the request as quickly as possible with a cached response if available, falling back to the network request if it's not cached. The network request is then used to update the cache. As opposed to [some implementations](https://github.com/GoogleChrome/workbox/issues/2807) of stale-while-revalidate, this strategy will always make a revalidation request, regardless of the age of the cached response.
This is a fairly common strategy where having the most up-to-date resource is not vital to the application.

gabek:
There's a linter error that needs to be resolved. https://github.com/owncast/owncast/actions/runs/11484299013/job/31986899925?pr=3970#step:10:7

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast">fix: fix runtime caching error and adjust rules · owncast/owncast@e8ccfab</a></strong></div><div>Take control over your live stream video by running it yourself.  Streaming + chat out of the box. - fix: fix runtime caching error and adjust rules · owncast/owncast@e8ccfab</div></blockquote>

web/next.config.js
heongle: Files that is compiled for production, nextjs will include `contenthash` as their name, shouldn't be a problem. If the files that placed inside public directory will be changed anytime, for example CSS files inside `public/styles/admin`, then I think there will be chances the admin UI will break.

The impact can be vary, from what I know, service worker will prevent to request network if the age specify in runtimeCaching is not expired. When runtimeCaching have shorter age than HTTP cache header, the request from service worker will be served by browser cache

About the different from previous `runtimeCaching`, from the library [cache.js](https://github.com/shadowwalker/next-pwa/blob/master/cache.js), the combination of last three config starting from [here](https://github.com/shadowwalker/next-pwa/blob/master/cache.js#L128) will match all apis and others routes, in previous version it is directly included to `runtimeCaching` which made all routes to run through service worker, causing `admin` route basic auth not showing on Firefox

Due to I lack of sufficient knowledge which Owncast assets is intended to be served by runtimeCaching, initial changes I only did is by removing `NetworkOnly` routes and copy exactly what written in library except last three config is removed, so the routes that is not explicitly stated will just skip service worker. I think if having the cache controlled by HTTP cache header is preferred, maybe we can just give empty array to runtimeCaching to prevent any unexpected caching happen on client side while still having the benefit of precache

heongle: @gabek After some thought and testing, passing empty array to `runtimeCaching` seems to be a possible solution, `precaching` will working as usual.  When there is any assets is confirmed will benefit from `runtimeCaching` rather than only using HTTP cache header someday, we can then add config for that too

What do you think?

gabek: That sounds like a good solution to me! It's mostly the precaching that this was all wired up for anyway, so not worrying about the runtime caching is probably the best bet. Thank you!

gabek:
I'll do some testing with this, and then we can get it in.

gabek:
This looks to be working great! Thanks for going back and forth discussing it. A simpler config is a better config! :)

PR #3975, opened 2024-10-20
fix: username wrapping

### LINKED ISSUE #3842: Chat - Join/leave messages layout breaks on narrow chat viewports
## Current behaviour

I noticed that the join/leave messages looks like 2 columns when chat size is at min width. Nothing critical, but strange
image

It looks like this on default chat width size
image

### another example of chat:
default width
image

narrow width. notice some characters are not fully visible.
image

## Expected Result 
- join/leave message for long usernames should flow smoothly like a sentence.
- join/leave message is responsive on narrow chat width 
- Look similar to images below
Image
Image

## Proposed Solution
Possible to do it like this for long usernames, shown in image below?
image

<hr>

### Notes
Use display name: truesuperheroeswearkiltsnocape

Note to OP: To look at name change notification messages.

### Additional information
Web browser
Owncast v0.1.3

nekojanai:
closes #3842


germainelee:
Hey @nekojanai , can you share a screenshot of how it looks (on narrow chat with) please? Thanks.

nekojanai:
Hey @germainelee !
It looks like this in the storybook:
owncast-narrow-viewport-chat-username

germainelee:
Just to clarify, your implementation looks like that image? i can't tell the difference between yours and the one in [Storybook](https://owncast.online/components/?path=/story/owncast-chat-messages-chat-join--moderator&args=displayName:truesuperheroeswaerkiltsnocape&globals=viewport:iphone5Portrait).

nekojanai:
yeah, the difference it that in the current storybook join message display, the username would wrap, this pr prevents that from happening. Try adding a minus somewhere in the middle of the username.

Well tbqh there's probably a better way to "solve" this, but this way the username doesn't wrap in the middle anymore

germainelee:
it would look like this in storybook 
image
How would it look in this pr using `truesuperheroes-wearkiltsnocape` display Name?

nekojanai:
like this:
owncast-narrow-viewport-chat-username-00

gabek:
I appreciate you looking at it! I'm guessing those three components in that single line are split up logically in the DOM, and that's why we're seeing each perform its own layout when running out of room. Would you be able to look at that, @nekojanai?

nekojanai:
> I appreciate you looking at it! I'm guessing those three components in that single line are split up logically in the DOM, and that's why we're seeing each perform its own layout when running out of room. Would you be able to look at that, @nekojanai?

Yeah, I'm not quite sure what to do about that in particular but how about this?

My proposed solution is to keep the username and icon part from wrapping and then let the container overflow horizontally? (The horizontal scrollbars are not visible in the attached screenshot)

overflow

gabek:
I guess in theory, the entire thing should be treated as one single sentence, not treating each component separately and having each element have its own wrapping rules. It should wrap like any other piece of text would, since it's just a sentence.

nekojanai:
> I guess in theory, the entire thing should be treated as one single sentence, not treating each component separately and having each element have its own wrapping rules. It should wrap like any other piece of text would, since it's just a sentence.

So, as shown in the attached screenshot?

germainelee:
I've updated my comment to be more clear with My Proposed Solution: 
> Look similar to images below
> Image
> Image

Originally posted by @germainelee in https://github.com/owncast/owncast/issues/3842#issue-2409454707

nekojanai:
@germainelee @gabek Ok, how about this?
username-wrap
username-wrap-00

germainelee:
Great work @nekojanai! This looks good to me. 

Maybe you can take a look at [Chat leave message](https://owncast.online/components/?path=/story/owncast-chat-messages-chat-part--regular) when Gabe is ok with your implementation as they kind of come in a set.

nekojanai:
@gabek

gabek:
Thanks so much for looking at this!
I should probably combine the join and part components together. No need for two of them.

PR #3997, opened 2024-11-05
Offline banner i18n formatting

### LINKED ISSUE #3979: Fix OfflineBanner component translation support

The [OfflineBanner](https://github.com/owncast/owncast/blob/localization/web/components/ui/OfflineBanner/OfflineBanner.tsx) timestamp is showing `[Object object]` as a side effect of adding translation support to this component.

Image

Component: 
https://629132c6e23893003a9e89c5-kdzbhzbhes.chromatic.com/?path=/story/owncast-layout-offline-banner--example-default-with-notifications

mahmed2000:
Fixes #3979.

I changed any json that was still in English to use the formatting for that string. Also updated the "the next time _ goes live" as well. There's a few more strings that would need similar changes, this is just the only one that fully breaks.

As an aside, is the https://github.com/owncast/owncast/blob/localization/web/i18n/strings.json file supposed to be empty? Apr 25 is when that happened and is also the last activity on Crowdin. Might be nothing, but it seems related given it shows "0 translatable strings" for everything.

### DIFF
--- a/web/components/ui/OfflineBanner/OfflineBanner.tsx
+++ b/web/components/ui/OfflineBanner/OfflineBanner.tsx
@@ -102,7 +102,7 @@ export const OfflineBanner: FC<OfflineBannerProps> = ({
         {lastLive && (
           <div className={styles.lastLiveDate}>
             <ClockCircleOutlined className={styles.clockIcon} />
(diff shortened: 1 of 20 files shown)


gabek:
Since there were very few strings to translate, the setup with Crowdin wasn't a priority. Like you saw, at some point it overwrote itself with emptiness. Now that it has changed, I could probably fix that up. Thanks for adding this!

Issue #3991, opened 2024-11-02
Issues when running on airgap environment


gaby:
### Share your bug report, feature request, or comment.

I'm currently testing `owncast` on a server with no internet access, where LAN users will be able to see the stream.

One thing we noticed is the default "Stream Offline" page is trying to reach `owncast` website to render a video in a thumbnail.

Second issue is during a stream the emojis menu timeouts because its trying to reach CDN resources.

gabek:
Owncast is completely self-contained. All the assets are available within Owncast itself. The page content has an embedded video as an example of what you can do with it. It's meant for you to remove and add your own content.

I'll need more detail about what this CDN is that you're referring to. Owncast does not point to a CDN.

gaby:
@gabek owncast doesn't make the requests to the internet, but npm dependencies do. Here are the requests happening on the background when opening the `emoji` menu within a chat. These are caused by `picmo` which is deprecated: https://www.npmjs.com/package/picmo

![Image](https://github.com/user-attachments/assets/b68c2a97-5edf-45c1-a4b5-0a357f105fac)

The documentation for `picmo` has a fix for using local data, that is documented here https://github.com/joeattardi/picmo/blob/main/docs/docs/02%20-%20usage/06%20-%20emoji-data.md#provided-static-data

gabek:
I'm afraid I'm not able to reproduce this.

![Image](https://github.com/user-attachments/assets/7159befa-c822-41da-9c62-427b423dafc8)

gaby:
@gabek This only happens during a stream, disconnect the host from the internet and open the Emoji Picker. Here is the code where `picmo` uses CDN urls: https://github.com/joeattardi/picmo/blob/main/packages/picmo/src/data/emojiData.ts#L13

<blockquote><img src="https://opengraph.githubassets.com/19d41837846f449a5f6d9ec3c647cff6ea065e7fa6625fdf985bdef669deeab8/joeattardi/picmo" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/joeattardi/picmo/blob/main/packages/picmo/src/data/emojiData.ts">picmo/packages/picmo/src/data/emojiData.ts at main · joeattardi/picmo</a></strong></div><div>JavaScript emoji picker. Any app, any framework. Contribute to joeattardi/picmo development by creating an account on GitHub.</div></blockquote>

gaby:
[...]

PR #4001, opened 2024-11-07
Replace picmo with emoji-mart

### LINKED ISSUE #3598: Replace unsupported Picmo emoji picker with something else

Unfortunately, our emoji picker, [Picmo](https://github.com/joeattardi/picmo), is no longer supported. While Picmo isn't going to stop working out of nowhere immediately, it may stop working due to changes in React or some other dependency at some point down the road. I'd rather us remove this library on our schedule instead of somebody else's schedule.

See [EmojiPicker.tsx](https://github.com/owncast/owncast/blob/develop/web/components/chat/ChatTextField/EmojiPicker.tsx) for our implementation of the emoji picker.

I don't have a preference on what we replace it with, but if anybody has recommendations, feel free to suggest them here!

mahmed2000:
Closes #3598 and #3991.

~Emoji-mart does not have a way to specify the initial category. Custom emojis cannot be set to the start of the category list either currently.~ Functionality exists, but has been broken since emoji-mart@5.3.0. Manually set version for now to 5.2.2.

~The emojis in `/static/img/emoji/mutant/` do not render in the picker (still work fine in the textbox).~ Something about emoji-mart's css does not like viewbox only svgs. Fixed now.

### DIFF
--- a/web/components/chat/ChatTextField/ChatTextField.tsx
+++ b/web/components/chat/ChatTextField/ChatTextField.tsx
@@ -152,15 +152,14 @@ export const ChatTextField: FC<ChatTextFieldProps> = ({ defaultText, enabled, fo
     contentEditable.innerHTML += textToInsert;
   };
 
-  // Native emoji
-  const onEmojiSelect = (emoji: string) => {
-    insertTextAtEnd(emoji);
-  };
-
-  // Custom emoji images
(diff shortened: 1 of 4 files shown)


gabek:
This works great! Thanks for swapping this out!

PR #4002, opened 2024-11-07
fix: registerUser request body is invalid

auyer:
I found two issues:
1. This `options` object is being passed down to fetchData, where it is deconstructed by the "data" attribute, not body.
2. the `data` object is being transformed into JSON downstream, thus the stringification done here makes for a string json object only

To find it, I added a log message to the Backend in an error that is being ignored:
```go
// core/chat/chat.go
decoder := json.NewDecoder(r.Body)
	var request registerAnonymousUserRequest
	if err := decoder.Decode(&request); err != nil { //nolint
		// this is fine. register a new user anyway.
+		fmt.Println("Failed decoding request body ", err)
	}
```
At first, the body was always empty (because of the `body` attributes being used instead of `data`)
Fixing it resulted in the following error:

> Failed decoding request body  json: cannot unmarshal string into Go value of type controllers.registerAnonymousUserRequest

This PR fixes them both. I will also look for similar issues in the project.


gabek:
Thanks for looking at this. What is the piece of functionality that currently isn't working? Something to do with display names, but what exactly?

auyer:
Hi @gabek 
I believe this is not affecting "normal usage" of Owncast.
I am testing a customization to authenticate users in a different system, and that's where I noticed it..

This existing function seems to be intended to be called with an optional predefined name ( tho it is not being used like this yet, so `optionalDisplayName` is always empty atm).
```js
  // ClientConfigStore.tsx
  const handleUserRegistration = async (optionalDisplayName?: string) => {
```

gabek:
Got it! Thanks for the clarification.

gabek:
~~Looks like there's a linter warning. Would you mind looking at that?~~

Nevermind. It's unrelated to your change.

auyer:
I just updated my branch with `develop`, just in case it fixes this unrelated linter warning. :smiley:

gabek:
Ok, I fixed that warning in `develop`. Sorry about that.

gabek (approved):
Thanks for catching this, and for the fix!

Issue #3229, opened 2023-08-02
Updating server details are not reflected in the UI when social features are enabled


gabek:
### Share your bug report, feature request, or comment.

It was mentioned to me that the flow for adding/managing tags in the admin isn't a great experience and doesn't provide feedback.

gabek:
I was unable to reproduce any issues here. I'll have to get more information. @silksow can you help provide details?

This is what I see in my testing of tags:

1. Click on the tag input text box
2. Type in a tag string
3. Hit enter
4. The tag shows up in the list
5. "Tags updated" success message shows.

silksow:
In Firefox and (flatpak) Brave on Fedora Silverblue, I can reach step 4 on reloading the page although it can take multiple tries, but I don't get the message in step 5 in either browser since 0.1.1

gabek:
Are there any messages in your error console?

silksow:
No, I'm afraid not.

gabek:
If anybody else sees this, I'd love if you could report back if you're having issues with tag management as well, since in my testing I haven't been able to reproduce this.

germainelee:
This is what I do and see when testing tags in admin:

**Add tags:**
1. Navigate to Configuration -> General
2. Click on the tag input text box
3. Type in a tag string
4. Hit enter once
5. Nothing happen
6. Hit enter multiple times and nothing happens
7. Reloads page
8. The tag shows up in the list
9. "Tags updated" success message did not show

Not sure if this message in inspect console is normal.
![image](https://github.com/owncast/owncast/assets/118464031/491892ca-7985-49cb-a23f-acb66daf4ae7)

Note: it doesn't work on mobile but that's not the main focus here
<hr>
Environment: Brave browser on windows

gabek:
For reference, `workbox` isn't related to tags, it's a background service worker that downloads assets, but it does have the ability to block the main thread. So it could possibly be indirectly connected, or not at all connected.

I'm not able to reproduce this. Are you running a production build or a dev server? I'm assuming it's a production build because I don't think we run workbox in dev.

germainelee:
Where can I see / find if it is production build or a dev server? at the bottom it says Owncast v20240617-nightly

gabek:
Ohh you're looking at nightly. Any server that's actively running publicly is a production build. If you're running the web app source code locally, then the web app is running under a dev server to allow for development.

germainelee:
Is nightly a production or dev build?

[...]

PR #4009, opened 2024-11-11
Communicate and handle apub outgoing updates being delayed

### LINKED ISSUE #3811: Admin Social links UI not reflecting modifications

When I **add**, **edit**, **delete**, and **reorder** a social link, the UI is not updated immediately with revised data. I have to refresh the page to view the revised data.

## Expected behaviour
All kinds of edits should be updated in real time.

## Steps to replicate (add a social link)

1. Navigate to Configuration -> General
2. Click 'Add a new social link' Button
3. Enter https://directory.owncast.online/
4. Click 'Ok'
5. Observe button changes to loading state, link is not added to list

image

6. Click 'Cancel'
7. Refresh page
8. Observe Link is added to list

image

<hr>
Note: Something might be affecting the UI stuff as there's issue #3229 

Should the user be able to add a social link without a name?

mahmed2000:
#3229 and #3811. Mitigation, not a full fix.

Currently, some API calls will federate updates to followers. While this does not wait for all requests to receive a response, this will wait for *all requests to be queued*. This only happens once (n followers - workerPoolSize) requests have been processed in the best case. This will block the API call from returning anything, and since the progress status indicator has also not been working, there is no feedback to the admin. This PR:

- Makes icon-only status messages (current default for processing) show up for all input fields.
- Buffer the apub outbound queue so more messages can be queued without blocking.
- Add a warning message in the event that the outbound queue is full.

Potential additions, not sure if needed:
- Just having a delay means there's a potential for weird desync issues. Maybe disable the input if the request is processing?
- Can make this execution a goroutine: https://github.com/owncast/owncast/blob/41a64498366d6dbe29a72d02cd868bbefe32cb1c/activitypub/outbox/outbox.go#L237 This will fix the delay entirely, but can spawn way too many goroutines.
- The queue being full will cause the same problems as now with this PR. Can simply skip any requests that can't be queued, but will cause problems if anything important gets skipped instead of an update.

Beyond the scope of this:
- Since this is being caused by the API calls also federating activity, it might make more sense to split the functionality into separate calls. That has the advantage of allowing a configuration of auto-updates on changes, or consolidating a shotgun of multiple rapid changes into a single `UpdateFollowersWithAccountUpdates` call to respect rate-limiting.

### DIFF
--- a/activitypub/workerpool/outbound.go
+++ b/activitypub/workerpool/outbound.go
@@ -19,7 +19,7 @@ var queue chan Job
 
 // InitOutboundWorkerPool starts n go routines that await ActivityPub jobs.
 func InitOutboundWorkerPool() {
-	queue = make(chan Job)
+	queue = make(chan Job, workerPoolSize*5) // arbitrary buffer value.
 
 	// start workers
 	for i := 1; i <= workerPoolSize; i++ {
--- a/activitypub/workerpool/outbound.go
+++ b/activitypub/workerpool/outbound.go
@@ -31,11 +31,11 @@ func InitOutboundWorkerPool() {
 func AddToOutboundQueue(req *http.Request) {
 	select {
 	case queue <- Job{req}:
-		log.Tracef("Queued request for ActivityPub destination %s", req.RequestURI)
 	default:
 		log.Warnln("Outbound ActivityPub job queue is full")
 		queue <- Job{req} // will block until received by a worker at this point
 	}
+	log.Tracef("Queued request for ActivityPub destination %s", req.RequestURI)
 }
 
 func worker(workerID int, queue <-chan Job) {
--- a/activitypub/workerpool/outbound.go
+++ b/activitypub/workerpool/outbound.go
@@ -29,8 +29,13 @@ func InitOutboundWorkerPool() {
 
 // AddToOutboundQueue will queue up an outbound http request.
 func AddToOutboundQueue(req *http.Request) {
-	log.Tracef("Queued request for ActivityPub destination %s", req.RequestURI)
-	queue <- Job{req}
+	select {
+	case queue <- Job{req}:
+		log.Tracef("Queued request for ActivityPub destination %s", req.RequestURI)
+	default:
+		log.Warnln("Outbound ActivityPub job queue is full")
+		queue <- Job{req} // will block until received by a worker at this point
+	}
 }
 
 func worker(workerID int, queue <-chan Job) {
--- a/web/components/admin/FormStatusIndicator.tsx
+++ b/web/components/admin/FormStatusIndicator.tsx
@@ -12,7 +12,7 @@ export const FormStatusIndicator: FC<FormStatusIndicatorProps> = ({ status }) =>
   const classes = classNames({
     'status-container': true,
     [`status-${type}`]: type,
-    empty: !message,
+    empty: !message && !icon,
   });
   return (
     <span className={classes}>
--- a/web/components/admin/config/general/EditSocialLinks.tsx
+++ b/web/components/admin/config/general/EditSocialLinks.tsx
@@ -19,7 +19,12 @@ import {
   DEFAULT_TEXTFIELD_URL_PATTERN,
 } from '../../../../utils/validators';
 import { TextField } from '../../TextField';
-import { createInputStatus, STATUS_ERROR, STATUS_SUCCESS } from '../../../../utils/input-statuses';
+import {
+  createInputStatus,
+  STATUS_ERROR,
+  STATUS_PROCESSING,
+  STATUS_SUCCESS,
+} from '../../../../utils/input-statuses';
 import { FormStatusIndicator } from '../../FormStatusIndicator';
 
 const { Title } = Typography;
@@ -140,6 +145,10 @@ export default function EditSocialLinks() {
 
   // posts all the variants at once as an array obj
   const postUpdateToAPI = async (postValue: any) => {
+    if (!displayModal) { // only create the processing status if the modal is inactive
+      resetTimer = null;
+      setSubmitStatus(createInputStatus(STATUS_PROCESSING));
+    }
     await postConfigUpdateToAPI({
       apiPath: API_SOCIAL_HANDLES,
       data: { value: postValue },
--- a/web/components/admin/config/general/EditSocialLinks.tsx
+++ b/web/components/admin/config/general/EditSocialLinks.tsx
@@ -145,7 +145,8 @@ export default function EditSocialLinks() {
 
   // posts all the variants at once as an array obj
   const postUpdateToAPI = async (postValue: any) => {
-    if (!displayModal) { // only create the processing status if the modal is inactive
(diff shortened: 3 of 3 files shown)


gabek:
You point to this being connected to the admin UI not updating, but I'm not seeing how they're related. What am I missing?

mahmed2000:
At this point we know that the POST request is being sent. Refreshing the page shows the API backend is correctly saving it. silksow mentioned needing multiple tries, which is also part of why I'm not really calling this a full fix, but that might have been back when api calls were being cached accidentally. I don't really know.

From here there's 2 possibilities:
- Something is actually wrong with the UI code. Given the last updates on the issue, and the nightly container not being updated thus 0 code changes, that seems unlikely if the problem can just fix itself to some degree. As far as I can tell, nothing about the web code even indicates this being a possible state because of the UI outside of the missing progress status.
- The backend handler is being held up at some point between saving the changes and returning a response (federating updates). This would explain why the problem never shows up testing locally and why it can be inconsistent. There is no status code yet to run either onSuccess or onError hooks, so it looks like nothing is happening.

My point is that saying the UI isn't updating begs the assumption that the API is responding in a timely manner.

The second problem is essentially demonstrated by this code:
```go
package main

import (
	"fmt"
	"time"
)

func lazy_worker(c chan int) {
	for true {
		job := <-c
		time.Sleep(time.Second * 2) // simulate response time for any apub request
		fmt.Println(job)
	}
}

func main() {
	c := make(chan int)
	for i := 1; i < 5; i++ {
		go lazy_worker(c)
	}
	for i := 1; i < 20; i++ { // 20 followers, basically
		c <- i
	}
	fmt.Println("main done")
	for {
		if len(c) == 0 {
			time.Sleep(time.Second * 5) // just waiting for workers to finish, main done is what matters
			break
		}
	}
}
```

This is trying to simulate `InitOutboundWorkerPool` and `SetTags` -> `UpdateFollowersWithAccountUpdates` -> `AddToOutboundQueue`

This is as it is currently with an unbuffered channel. "main done" is basically when the API call returns and would then update the UI. Compare the result with changing `make(chan int)` to `make(chan int, 20)`. The lack of a progress status indicator would also make it look like nothing is happening. Now imagine, those 2 seconds per job being rate-limited, from a backed up remote instance, from a reverse proxy 504, etc and you get "not reflecting updates". Doesn't show up is not *never* shows up.

Again, actually sending and receiving apub activities is async, but queuing them is not, currently. This doesn't help the exact issue for the social links, but I'm not sure how to communicate the fact that what took barely a second can now take upwards of 30 just because of enabling federation and is directly proportional to the amount of followers - outside of just buffering everything.

mahmed2000:
I setup a testing environment to demonstrate this. Running owncast with federation on a 2 core VPS (so `workerPoolSize` is 2), and a snac2 server to generate 200 followers. A single code change was made to add a 100ms sleep inside the `SendActivityPubMessageToInbox` function in `activitypub/workerpool/outbound.go` to simulate latency on apub http requests since its otherwise all on 0 delay. This is only ever called by the apub workers, and so is never executed directly by an API handler.



With everything completely settled with `Accept` replies, and federation enabled, both the tag and social link issues happen as described for ~10-11s in this setup. This is expected given what I've described above: (200 - 2) followers / 2 workers * 100 ms per request = 9,900 ms of delay. Trying to spam submit the tag extends it even further. Reloading shows tag values do indeed get saved well before the API returns.



Also occurs with a basic 10 requests per second rate limit on the snac server and no 100ms delay

This problem goes away completely on disabling federation:



While processing several `Follow`s, owncast gets backed up completely in sending `Accept` replies, and is basically guaranteed to hit some timeout on trying to change tags.

In current public servers, I'm assuming something similar happens. On a 4 core VPS, if you have 4 followers that had their servers go down at that time, resulting in nginx gateway timeouts (504s *not* 502s), that one request will hog all 4 workers until they receive a 504 after a minute (nginx default) or whatever go's default timeout is, *even after the API request returns*. Since those followers will be sent a request with every federated activity since there's no system to keep track of that and stop, a whole bunch of requests will keep queuing. Because request n must wait for the 4 actors in request 1..n-1 (each) to timeout for the channel to be free to take another job. 

As a result `UpdateFollowersWithAccountUpdates` will not return for a while because it can't hand off all of its generated requests, and that 10s goes up substantially, also compounding on just high latency servers in general and any rate-limiting delays. Without the status indicator, it looks like nothing is happening. The implicit assumption is that this should be queuing in the backend (or not even happening), so the delay is misinterpreted as a UI bug.

Since this is involved in the larger issue around federation queuing, I'm fine with this being changed to just the status indicator changes. In hindsight, this PR doesn't really help other than just making it clear that the queue is full, polluting the log since there's no backing-off, and adding a tiny buffer.

gabek:
It took me a bit here to remember that `SetTags` actually sends out an ActivityPub update to followers, that's why I was confused at first where the connection between tags and federation was. Now it all makes sense, sorry about that.

There's probably a case here that no `UpdateFollowersWithAccountUpdates` should ever fire synchronously in an API handler. Perhaps we could just wrap the firing of them in a goroutine, or like you mentioned just `workerpool.AddToOutboundQueue(req)`.

Let's merge this in and get it deployed to `nightly`. I'll do some testing! Thanks for looking at it so in depth. It should help at least in some capacity. At the very least, it can't hurt :)

gabek:
I wanted to test in production with real users, and I just did on `nightly`.

I'm not saying it's connected, but the other day I was able to stream on nightly without an issue, but today was non-stop buffering, making streaming impossible. I'm yet to dig into what's causing it. I'll have to do some A/B testing in production. I didn't think any of these changes could have any negative side effects, however.

The following are expected log messages, but I wanted to save them here anyway.

image

mahmed2000:
> non-stop buffering

Just the hls video segments? Or all http requests were being delayed, the video segments are just the most clear case where its "visible" at a glance?

As far as I tested, the rest of the config API calls were unaffected even when the tags or social requests were stuck, video isn't something I thought to test. The simplest test would be to disable federation and see if the problem is still there. If it is, some change here likely messed something up, somehow. A clearer test would still be with reverting the change.

On a side note, 5x the CPU core count still seems low. I'm wondering if a more sensible default might be 5 * n_followers.

gabek:
@mahmed2000 Could you share how you setup snac2 to automate creating 200 users who requested follows from Owncast? This would be very useful.

mahmed2000:
snac2 can be managed from the command line for basically everything, so you can use a shell script and loops. Automating the users is simpler than setting everything up.

```
snac adduser $snac_data_path $i
snac follow $snac_data_path $i @streamer@owncast.instance
```

Just wrap that in your chosen shell's loop, iterating `i`.

The actual setup was a self-signed cert for local domains (I used *.fedi.local but anything should work, have a script for that too if wanted), editing `/etc/hosts` with said domain resolving to localhost, and nginx.

Setting up snac is just `snac init $dir; snac httpd $dir`.

gabek:
Very cool. I'll have to see if I can automate the setup of this for testing. Thanks!

PR #4038, opened 2024-11-28
fix(social): set federation enabled state after confirmation  

### LINKED ISSUE #4021: Federation enabled state does not respect the "Enable Social Features" modal cancellation
The fediverse enabled state only cares about the toggle switch in the admin. Enabling the switch prompts a modal. Cancelling or closing said modal only updates the UI for the toggle. On reloading the page, it will flip to an enabled state. The `enable` API call gets sent on toggling, which means federation is enabled before the modal even opens.

### Repro steps
 
- Create a server and configure it to have a server URL.
- Config -> Social
  - Disable social features if they have been enabled.
- Toggle the "Enable Social Features" switch and wait for the modal.
  - The "Compose Post" button also shows up at this point.
- On the "Enable Social Features" modal, either click "Do not enable" or close the modal.
  - The toggle flips off, but the "Compose Post" button is still visible.
- Reload the page.
  - The toggle is on now, and the main stream page shows the Follow button. Federation is enabled even though the modal's disclaimer was not agreed to.

varungujarathi9:
fixes #4021 
Set federation enabled state only after confirmation in modal.

Earlier the `enabled` value was set using the `postConfigUpdateToAPI` in `ToggleSwitch.tsx` component's `handleChange` function, whenever the switch was toggled, but since we wanted to set the state after the modal confirmation I am setting the `useSubmit` as `false` for this switch and am calling the `postConfigUpdateToAPI` in `handleEnabledSwitchChange`, `federationInfoModalCancelPressed` and `federationInfoModalOkPressed` functions based on user action.

This implementation doesn't feel clean to me since we are calling the `postConfigUpdateToAPI` outside the `ToggleSwitch`, but I couldn't think of a better way that did require complete overhaul of `ToggleSwitch`. Let me know if I should do this in some other way.


web/pages/admin/config-federation.tsx
```diff
@@ -37,7 +37,7 @@ const FederationInfoModal: FC<FederationInfoModalProps> = ({ cancelPressed, okPr
   <Modal
     width="70%"
     title="Enable Social Features"
-    visible
```
varungujarathi9, line 40: `visible` is being deprecated, instead `open` is used now

gabek:
> This implementation doesn't feel clean to me since we are calling the postConfigUpdateToAPI outside the ToggleSwitch, but I couldn't think of a better way that did require complete overhaul of ToggleSwitch. Let me know if I should do this in some other way.

Yeah, I get that, but I think it's ok.

Thanks so much for this fix!

Issue #4020, opened 2024-11-20
Brainstorm improvements to even better handle the Mastodon DDoS


gabek:
### Share your bug report, feature request, or comment.

The Mastodon Hug of Death, where sharing a link to an Owncast server on the Fediverse will then get accessed all at the same time in a DDoS is a well known issue. Owncast is particularly sensitive to it because if somebody shares a link to a live Owncast server, and the entire Fediverse starts slamming the Owncast HTTP server all at the same time, those who are trying to consume the live stream will be actively fighting for web assets.

While I've done work in the past to improve this, it's only better, not resolved. I've implemented a cache for the metadata page, so when all the link previews get populated on people's newsfeeds at least many of those requests to the small metadata-only page will be served via the cache.

I'm thinking perhaps of some kind of worker pool for when Mastodon (and others) make requests to Owncast they get queued up instead of handled right away. Having those requests responded to slowly is no big deal, nobody is waiting on them. But video requests are important to be immediately served.

https://www.jwz.org/blog/2022/11/mastodon-stampede/
https://news.itsfoss.com/mastodon-link-problem/

gabek:
I wonder if that approach would actually make things worse by keeping these requests open longer, meaning the web server actually has more concurrent requests in flight. It would be a good thing to load test.

- [x] Could also probably bump the [cache duration](https://github.com/owncast/owncast/blob/develop/webserver/handlers/index.go#L131) from 10 to 60 seconds, since 60 seconds is the upper bound that Mastodon will wait to make the request. And this response being a minute stale really isn't a big deal.

- [ ] Prevent multiple requests from re-rendering the template on a cache miss by possibly wrapping that logic in https://pkg.go.dev/golang.org/x/sync/singleflight
- [ ] Result of `utils.DoesFileExists` is not going to change, even during a cache miss. This can be cached separately, and aggressively.
- [ ] Only pull the template contents of `static.GetBotMetadataTemplate()` once, as it cannot change even on a cache miss.
- [ ] Set an eTag header `w.Header().Set("ETag", fmt.Sprintf("%x", md5.Sum(cachedHtml)))`
- [ ] Possibly cache the results of `configRepository`? Though technically behind the scenes, these values are living in an in-memory cache anyway.

renchap:
Can you clarify what kind of request those are? Are those requests to the ActivityPub endpoint, or to some URL that is used in the post content and that a preview needs to be generated for?

gabek:
[...]

Issue #3189, opened 2023-07-19
Sharing Owncast fediverse posts to a few thousand accounts causes video buffering


gabek:
### Share your bug report, feature request, or comment.

This sounds weird, but:

I've experienced when somebody with enough followers shares an Owncast ActivityPub post the video starts buffering. I particularly notice this when the Owncast account shares somebody's "Go live" post. I thought it was a coincidence at first, but it's happened enough times now where I'm confident it's not a coincidence.

Without yet investigating here's a few thoughts:

- Too many concurrent worker threads are sending out the messages, blocking web requests?
- The result of all the posts showing up in a few thousand timelines is causing a bunch of requests to Owncast to populate the inline "card" + embeded player all at once.

I'm guessing it's probably #1 and reducing the number of outbound workers should fix it.

gabek:
Nope, wasn't that. Must be the second one. Maybe.

gabek:
See #3253

gabek:
I reverted the changes since it was too heavy-handed and went outside the scope of this issue. Will readdress.

PR #4049, opened 2024-12-08
Increase outbound worker pool size to use follower count

### LINKED ISSUE #3811: Admin Social links UI not reflecting modifications

gabek:
@mahmed2000 I made this change based on your recommendation to bump this up. It'll only calculate this number at app launch, but it should be good enough for most cases, to be in the right ballpark due to the the multiplier.

Re: #4009

### DIFF
--- a/activitypub/activitypub.go
+++ b/activitypub/activitypub.go
@@ -1,6 +1,8 @@
 package activitypub
 
 import (
+	"math"
+
 	"github.com/owncast/owncast/activitypub/crypto"
 	"github.com/owncast/owncast/activitypub/inbox"
 	"github.com/owncast/owncast/activitypub/outbox"
@@ -17,7 +19,9 @@ import (
 func Start(datastore *data.Datastore) {
 	configRepository := configrepository.Get()
 	persistence.Setup(datastore)
-	workerpool.InitOutboundWorkerPool()
+
+	outboundWorkerPoolSize := getOutboundWorkerPoolSize()
+	workerpool.InitOutboundWorkerPool(outboundWorkerPoolSize)
 	inbox.InitInboxWorkerPool()
 
 	// Generate the keys for signing federated activity if needed.
@@ -29,6 +33,17 @@ func Start(datastore *data.Datastore) {
 			log.Errorln("Unable to get private key", err)
 		}
 	}
+}
+
+func getOutboundWorkerPoolSize() int {
+	var followerCount int64
+	fc, err := persistence.GetFollowerCount()
+	if err != nil {
+		log.Errorln("Unable to get follower count", err)
+		fc = 50 // Arbitrary fallback value.
+	}
+	followerCount = int64(math.Max(float64(fc), 50))
+	return int(followerCount * 5)
 }
 
 // SendLive will send a "Go Live" message to followers.
--- a/activitypub/workerpool/outbound.go
+++ b/activitypub/workerpool/outbound.go
@@ -2,13 +2,9 @@ package workerpool
 
 import (
 	"net/http"
-	"runtime"
 
 	log "github.com/sirupsen/logrus"
 )
-
-// workerPoolSize defines the number of concurrent HTTP ActivityPub requests.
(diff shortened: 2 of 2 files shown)


mahmed2000:
This will bump both the queue size and the number of goroutines handling outbox activities. I'm not sure what the downstream consequences of that might be (if any) by spawning that many. What I was considering was more just the queue specifically, because of the original issue, but this should help too since its still mostly just waiting on http requests than hogging computation. It was more a case of handling bursts than keeping a proportional throughput.

Does this still work if the `GetFollowerCount` call returns 0 followers? It would spawn 0 workers in that case, and make it impossible to follow the server since follow accepts never get sent. Might be better to use the number of followers for the max size and the old number of processors for a blanket minimum.

gabek:
> Does this still work if the GetFollowerCount call returns 0 followers? It would spawn 0 workers in that case, and make it impossible to follow the server since follow accepts never get sent.

I actually set a minimum number of `50` followers (arbitrary value), so even if there are zero, or few followers, the number of available workers should be enough.

https://github.com/owncast/owncast/pull/4049/files#diff-193ff8009eaca18502f2a9974a3da62b5fc00f6fbf55215229c80594bb4c84cfR45

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/owncast/owncast/pull/4049">Increase outbound worker pool size to use follower count by gabek · Pull Request #4049 · owncast/owncast</a></strong></div><div>@mahmed2000 I made this change based on your recommendation to bump this up. It&#39;ll only calculate this number at app launch, but it should be good enough for most cases, to be in the right ball...</div></blockquote>

mahmed2000:
> I actually set a minimum number of 50 followers

I see that now, brain somehow glossed over the Max call. This should be good then, outside of the other issue of decoupling federation updates from the API calls.

Does the buffering issue you mentioned in the original PR thread still occur?

gabek:
> Does the buffering issue you mentioned in the original PR thread still occur?

Unfortunately, nightly only gets updated with the nightly build, and it's all automated. So to perform the exact same test, I'd have to merge this in. So I guess I'll do that and test that and the API call performance.

gabek:
This change did improve a couple of things.

- No more buffering just from simple playback.
- When adding tags or social links the UI successfully updates.

However, buffering does occur briefly when saving any of those pieces of server metadata that cue up ActivityPub activities, probably as expected. I wonder if throttling would help at all. Probably not, as in this test case there aren't actually that many followers.

mahmed2000:
Tried a local test again for video. The latency on segments doesn't seem to be affected by either setting tags or spamming in a bunch of new follows mid-stream in any meaningful level of variance. Might be something only over latency connections, might not. I did see #3189, so it probably is apub related in some way.

gabek:
There are two different issues. One is ActivityPub related. Any type of application blocking that is taking place when queuing up outbound ActivityPub activities. The second is the DDoS the server receives by the Fediverse all hitting the Owncast server. This also happens if you simply share the link to the Fediverse. This isn't ActivityPub related. I'm tracking this issue here: #4020.

And I think you're right, you probably didn't see any issue due to latency and general stability. In the real world there are timeouts and stuff going on with all these ActivityPub connections happening to servers that might be down, or broken or whatever. (I also think this machine is probably more resource constrained haha)

But #3189 is still completely reproducible and it's probably time to add some instrumentation around it.

danhendo89 (approved):

PR #4062, opened 2024-12-20
Make the error alert in OfflineEmbed closable

### LINKED ISSUE #4056: Can't fix incorrectly entered fediverse account in Offline embed
Error is blocking the input field. 
Image

Is it possible to do it like this?
Image

mahmed2000:
For #4056

This adds an option to close the alert in the OfflineEmbed component. Since the error message only ever pops up in a failed follow, the callback resets the embed state back to `FollowPrompt`, which returns to the input prompt state.

This does not address what happens if the close button ends up out of bounds.


gabek:
I think that's a completely reasonable middle ground. Thanks!

PR #4063, opened 2024-12-21
embed screen style adjustments

### LINKED ISSUE #3683: New offline embed has overflow problems

image

### LINKED ISSUE #4051: Slightly tweak the offline embed layout
Currently, the name of the stream is at the very bottom, and I think that's counterintuitive when looking at it. Reading from top to bottom you read the description first, and it doesn't make a lot of sense if you don't know that it's the description for the stream.

See the below example:
> A streaming platform which uses and promotes open-source software.

What is a streaming platform which uses and promotes open-source software?

I'm not looking to redo this from scratch, or make any visual component changes. I'm just looking to re-order the text items so it makes more sense when you read it from top to bottom.

Image

gingervitis:
restyle and relayout embed screen to account for smaller screen displays

- address https://github.com/owncast/owncast/issues/3683 to address overflow issues
- address https://github.com/owncast/owncast/issues/4051 to move the name of the stream


gingervitis:
Since this embedded content could end up in a variety of display environments, I've tried to account for a variety of responsive behavior.

- Move the Server name and logo to the top for more identifiable visibility.
- Give the embed container a `font-size: 16px`, and something smaller for smaller displays, so that the font sizes in children elements can scale to it using `em`s. 
- Since content could get lengthy, I've tried to account for screen _height_. 
  - If the height is really short, hide the `subtitle`, so that more space can be given to the "follow server" form.
    - _This may be controversial, so let me know if this is a bad idea, and what we should prefer to do instead._
  - If the screen is slightly taller, reveal more of the `subtitle` text, but clamp it to 3 lines.
  - Otherwise display all the text.
 -  At narrow screen widths, the overall font-sizes will also be smaller
 - At really large screen sizes, keep the content to 1024 width max , so if content is sparse things don't look so spread out and hard to read.

gabek:
Thank you :D

For reference here's this PR in storybook: https://629132c6e23893003a9e89c5-vasktmzrfw.chromatic.com/?path=/story/owncast-layout-offline-embed--example-default-with-notifications

<blockquote><div><strong><a href="https://629132c6e23893003a9e89c5-vasktmzrfw.chromatic.com/?path=/story/owncast-layout-offline-embed--example-default-with-notifications">@storybook/core - Storybook</a></strong></div></blockquote>

gingervitis:
Made some updates to better utilize the space for when there's no follow option, as well as to keep the `subtitle` content around.
- If `subtitle` content is lengthy, on short screens it will have an overflow-y, which could allow users to try and scroll in the skinny space.
- Center the content when there's no follow option.

gingervitis:
New changes also made in this update:
- separate out the Follow Form functionality into its own component so it can be used in other places
- do not display the Follow button on the offline embed if the window is less than 300px
- improve styling of Follow error message Alert
- improve styling of Follow form legibility
- improve button arrangement and styling for Follow form buttons
  - The `Join fediverse` shouldn't be a "primary" button as it distracts from the form's primary function. So I've made a "text" style button and placed it in a secondary position.
- general ant-override button styling tweak to accommodate all types of ant buttons


image

---

primary follow modal using same follow form.

image

germainelee:
> ... if the window is less than 300px

I wonder how this would look like on mobile. Would the scrollbar be visible?

gingervitis:
> > ... if the window is less than 300px
> 
> I wonder how this would look like on mobile. Would the scrollbar be visible?

I know this isn't an accurate representation, but I suggest we hide the follow button if the space is just way too small.

The description scrollbar may depend on the OS's native scrollbar settings.

image


Otherwise for 300px width and up
image
image


Unfortunately I'm not sure how to simulate the actual embedding in a social post on a mobile device. If someone is able to conjure up a screenshot for that, I'd love to see it!

germainelee:
im ok to hide the follow button if the space is extremely small.

Here's a quick inaccurate mockup of the embedding in a social post on a mobile and desktop device, using the image from https://github.com/owncast/owncast/issues/3683#issuecomment-2543569119. The mobile dimensions of the embed is 299 x 169 on a 340w "screen".

From what I understand, user should be able to scroll content when he interacts within the embed. tapping on 'Zalasur Media Live Streaming' text below would bring him to owncast instance.
image

gabek:
@gingervitis Looks like there's some linter warnings that are failing.

gabek:
If somebody was using an embed at a larger size (to fill a page with just a player, for example), it looks kind of odd.

- The space between the logo+title and the body content seems spacious.
- While at the same time the space between the title and the subtitle seems very tight.

In the existing version, while it doesn't work in tight spaces, it does work in open spaces because it's all centered.

I'm not sure I have a solution, other than have two completely different layouts.

screenshot_2025-01-04-151255
screenshot_2025-01-04-151636

gingervitis:
Hmm, i'm currently getting this for large screen layout. The spacing seems fine?
imageimage
Though I agree it looks a bit odder when social is enabled.



>> screenshot_2025-01-04-151255

gingervitis:
I made some updates to simplify the layout logic to center everything. Seems to keep everything more compact, less empty?
imageimage

gabek:
This looks great! I'm sold. Thank you for working on this.

PR #4066, opened 2025-01-03
Update Mastodon logo to current brand colours.

andypiper:
This small PR simply updates the Mastodon logo SVG file that is shown in a user profile page to use the current preferred brand colours (per https://joinmastodon.org/branding).

Happy New Year!


gabek:
`static/web/img/platformlogos/mastodon.svg` will get overwritten when the web project gets built. And since it's different from `web/public/img/platformlogos/mastodon.svg` that you committed, it might not be the result you're looking for. Whatever you want this image to be should be placed in `web/public/img/platformlogos/mastodon.svg`.

andypiper:
Sure thing - thanks - let me re-submit.

PR #4067, opened 2025-01-03
Update to current Mastodon logo.

andypiper:
Updates the Mastodon logo to the current version (from https://joinmastodon.org/branding)

Replaces #4066.


gabek:
It looks like it's the same color as the one being replaced. Is this expected?

screenshot_2025-01-03-122052

andypiper:
Wow, I'm doing a terrible job with this PR! Sorry 🤦🏻 that's clearly not correct.

PR #4078, opened 2025-01-05
fix skip link visibility and it moving logo upon focus

gingervitis:
Currently when the skip links gain focus, they will shift the main logo over.
image

This fix keeps the skiplinks in place, and improves their visibility.
image


gabek:
Thank you!

PR #4198, opened 2025-02-08
Update follower cards to have better responsive design

### LINKED ISSUE #3248: Followers design/layout needs improvement

germainelee:
closes #3248 

- Number of columns would change gradually, from 4 to 1 depending on screen width
- when the name is 2-lines long in the card, the entire row should match that height while the other cards should not be affected


gabek:
Thank you!

PR #4202, opened 2025-02-12
Resolve issue where follows would not work if private federation mode was enabled

### LINKED ISSUE #4142: Accepting a Follow request in Private mode does not work

Accepted follow requests seem to not be able to be processed properly by the other server, I noticed that on both GoToSocial and Mastodon follows stay stuck in "Requested" state, despite the request being accepted in Owncast Admin.

I first noticed this with my own GoToSocial instance which prompted me to investigate in that direction and open https://github.com/superseriousbusiness/gotosocial/issues/3652, but in the meantime I noticed the same behaviour on Mastodon.

When disabling Private mode temporarily the Follows go through without any issue.

gabek:
This change adds the follow request into follow response, resolving the issue where if private mode is enabled follows were not working. @mahmed2000 

Closes #4142

### DIFF
--- a/activitypub/inbox/follow.go
+++ b/activitypub/inbox/follow.go
@@ -47,7 +47,7 @@ func handleFollowInboxRequest(c context.Context, activity vocab.ActivityStreamsF
 		}
 	}
 
-	// Save as an accepted activity
+	// Save as an activity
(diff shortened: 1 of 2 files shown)


mahmed2000:
Tested this locally, works correctly with snac.

PR #4209, opened 2025-02-14
fix: username wrapping for parting message

nekojanai:
Great work @nekojanai! This looks good to me. 

Maybe you can take a look at [Chat leave message](https://owncast.online/components/?path=/story/owncast-chat-messages-chat-part--regular) when Gabe is ok with your implementation as they kind of come in a set.

_Originally posted by @germainelee in https://github.com/owncast/owncast/issues/3975#issuecomment-2440462652_

---

Follow up commit to the comment above.


germainelee:
thank you!

PR #4257, opened 2025-03-21
WIP containerized ActivityPub test environment

### LINKED ISSUE #4155: Support local federation testing

Currently, there is no way to test federation without setting a completely functioning, public, ActivityPub environment. With a public hostname, SSL, port forwarding, etc. This makes development and testing very difficult and time-consuming.

We need a way to be able to work on this stuff without setting up an entire production environment.

@mahmed2000 [found a good solution using `snac`](https://github.com/owncast/owncast/pull/4009#issuecomment-2526828692), but setting it up needs to be reproducible.

- We should have a Docker environment that runs `snac`.
- It should create the self-signed cert for a specific hostname and run any kind of reverse proxying needed.
- It should have scripts for common tasks such as create user(s), follow, unfollow.

Without running some kind of DNS proxy, or similar, we'd still have to manually have to edit `/etc/hosts`, however.

gabek:
#4155

### DIFF
--- a/test/activitypub/Dockerfile
+++ b/test/activitypub/Dockerfile
@@ -1,0 +1,18 @@
+FROM alpine:3.20 AS builder
+
+RUN apk add openssl-dev curl-dev git make gcc openssl libc-dev
+
+RUN git clone "https://codeberg.org/grunfink/snac2"
+WORKDIR ./snac2
+RUN make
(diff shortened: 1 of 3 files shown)


gabek:
Was completed with a different approach in another PR.

PR #4311, opened 2025-05-04
Fix prometheus handlers

### LINKED ISSUE #4272: Prometheus Metrics broken after 0.2

Hello! I just upgraded from 0.1.3 to 0.2.1 and unfortunately it looks like prometheus metrics are not exporting in 0.2.X.  I receive a 200 but no data is returned anymore.

mahmed2000:
fixes #4272

Added a test for the prometheus endpoint, just checks if the string `owncast_instance_cpu_usage` metric exists in the response. Should be a good enough bar to clear but no clue on how to handle the rest of the methods, or if promhttp even handles those...

### DIFF
--- a/test/automated/.gitignore
+++ b/test/automated/.gitignore
@@ -1,2 +1,2 @@
 node_modules
-ffmpeg
+ffmpeg-*/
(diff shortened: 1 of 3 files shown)


gabek:
Thank you!

PR #4312, opened 2025-05-04
change go:embed call to use the go:embed all: directive

mahmed2000:
This makes use of [this](https://go-review.googlesource.com/c/go/+/359413) golang embed directive to include all files in the target static directory.

This allows embed of directories like _next without go complaining if those files are absent (like for a custom non-nextjs frontend)


gabek:
Nice, thank you!

PR #4331, opened 2025-05-08
sha-512 hash passwords longer than 72 bytes

### LINKED ISSUE #4270: bcrypt: password is limited to 72 bytes

Hi, I'm trying to upgrade an owncast instance and it shows this log:

```
time="2025-03-28T11:18:35Z" level=info msg="Migration datastore values from 3 to 4\n"
time="2025-03-28T11:18:35Z" level=fatal msg="error migrating admin password: bcrypt: password length exceeds 72 bytes"
owncast.service: Main process exited, code=exited, status=1/FAILURE
```

My admin password is indeed longer than 72 bytes, and I believe I should be able to have a password that is longer than 72 bytes, especially as this worked before.

mahmed2000:
Would close #4270, but would like some input from anyone using passwords longer than 72 bytes what the security implications of this are.

The other suggestion with this I've seen is to use HMAC-SHA instead of plain SHA. That requires adding a salt to the sha512 hash, which requires splicing and unsplicing the salt from the final bcrypt hash to store it which complicates things a bit and idk if the tradeoff is worth it.

Regardless of choice, this change is irreversible without making users reset the password with the cli flag.

TODO:
- add a test for longer passwords
- rework the db migrations to allow errors by wrapping the whole operation in a db transaction to avoid panics or errors breaking things.

### DIFF
--- a/test/automated/api/009_configmanagement.test.js
+++ b/test/automated/api/009_configmanagement.test.js
@@ -1,5 +1,6 @@
 var request = require('supertest');
 var bcrypt = require('bcrypt');
+var shajs = require("sha.js");
 
 const sendAdminRequest = require('./lib/admin').sendAdminRequest;
 const failAdminRequest = require('./lib/admin').failAdminRequest;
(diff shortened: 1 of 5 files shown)


gabek:
As far as I've heard, only one person has reported the issue with the 72 byte password. So I'm not sure we're going to see a lot of people in the wild.

mahmed2000:
Should be good to merge atp, I've tested pre v0.2.0 dbs (bcrypt addition) with this, but more verification is good. The transaction rollback impl should be a separate pr anyway.

Issue #4181, opened 2025-01-28
Webfinger error


simon-anderson:
### Share your bug report, feature request, or comment.

New 0.2.1 installation. Attempting to authenticate chat with FediAuth throws the error...

> Could not send code to fediverse: unable to get webfinger links when sending private message: invalid character '<' looking for beginning of value

![Image](https://github.com/user-attachments/assets/2d446c0f-05c8-491b-b354-dcc53588381a)

gabek:
Would you mind sharing the account you were trying to authenticate with? That looks to mean Owncast tried to look up your account, but couldn't.

mahmed2000:
Looks like a JSON decode error. The "<" would be the start of an HTML response.

Does the server in question use 301 redirects or delegation for webfinger requests by any chance?

gabek:
That's why I was trying to get the server, to see what it's actually returning for Webfinger requests. My guess it's returning an error page or only handling webfinger for specific user agents or something.

mahmed2000:
> My guess it's returning an error page or only handling webfinger for specific user agents or something.

Maybe. But its also possible its a generic 301 "Moved permanently" reverse proxy html and owncast's hadling of redirects is partly to blame for re-using any non-json reply with the `ErrUseLastResponse` for webfingers.

Even if this issue happens to be a niche case, that doesn't seem like correct behavior. Delegation isn't *un*common and it might be worth adding a content-type check anyway?

simon-anderson:
The account is simon_anderson@mastodon.social

I got a little further with the authentication this time, perhaps the earlier failure were to do with being a new instance? Owncast sent me the passcode but rejected the entry of it with the response in the screenshot, and rejected subsequent attempts with a "try again later" message..

![Image](https://github.com/user-attachments/assets/e8abd042-0f25-47ec-bc63-fdc640624d27)

gabek:
Do you see any additional errors in your browser console, or failed network requests in your network tab?

mahmed2000:
`includes...` is a js error.

https://github.com/owncast/owncast/blob/dedb5c5a40aa5d987f221d7c5f91c3a6782518a8/web/components/modals/FediAuthModal/FediAuthModal.tsx#L47

`errorMessageText` gets minified to C in the next chunks. If a raw error is set instead of a string, includes fails and buries the actual error message. 

[...]

PR #4332, opened 2025-05-08
check for webfinger status code before parsing

### LINKED ISSUE #4215: bug: if the account doesn't exist on the server, EOF error is shown
This error comes up when I enter the following in the input field. I can rectify my error by typing a different fedi handle, but what does that error mean? Can it be written differently so that viewers who have low technical knowledge would understand?

Note: I am not ety

_Originally posted by @germainelee in https://github.com/owncast/owncast/discussions/4103_

mahmed2000:
Closes #4215 

Tangentially affects #4181 since it should give some context as to *why* the remote server is returning html when json was requested.

The status code check should cover all cases where invalid/empty json is returned. So the current EOF now shows "webfinger request returned bad status code: Not Found, check account details". Invalid 200 OK gets decoded with a wrapping error to clarify context.

Could add errors more specific to the status code, or include the plain status code int in the error string.

### DIFF
--- a/activitypub/webfinger/webfinger.go
+++ b/activitypub/webfinger/webfinger.go
@@ -55,12 +55,16 @@ func GetWebfingerLinks(account string) ([]map[string]interface{}, error) {
 		return nil, err
 	}
 
+	if response.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


gabek:
Nice! Thanks!

screenshot_2025-05-10-131242

PR #4375, opened 2025-06-07
fix: improves offline embed responsiveness to small screens and long text content

### LINKED ISSUE #4319: Offline embed still looks bad

For whatever reason, we just can't get this layout to look good, even though we've iterated on it multiple times.

Image

fortyoneplustwo:
## Summary
This PR attempts to address issue #4319 by re-implementing the Offline Embed component to look nice on all screen sizes regardless of the length of its text content. This iteration tries to follow #4346 as reference while introducing some new design choices. The `css` has been re-written to be more maintainable.

### Changes
- Enforces 16:9 aspect ratio
- The stream title can span a max of 2 rows and gets clipped with an ellipsis on overflow
- The description can span a max of 3 lines and is scrollable on overflow
- The logo is larger and scales with the height of the title's container
- The background gradient now uses the user's primary theme color instead of the default purple

### Additions
- Clicking on the logo/title takes the user to the stream page

## Screenshot
Screenshot 2025-06-04 9 40 56 PM


gabek:
Looks like there's some linter errors that need to be cleaned up.

gabek:
Looks like there's a linter failure.

gabek:
When I open this embed on a fresh install of Owncast it has a blank background.

screenshot_2025-06-29-213936

fortyoneplustwo:
Looks like `--theme-color-primary` is not defined anywhere in the code base.

gabek:
Yeah, I don't think `--theme-color-primary` is a thing, it should be `--theme-color-action`. https://owncast.online/components/?path=/docs/owncast-styles-colors-palette--docs

<blockquote><div><strong><a href="https://owncast.online/components/?path=/docs/owncast-styles-colors-palette--docs">@storybook/core - Storybook</a></strong></div></blockquote>

web/components/ui/OfflineEmbed/OfflineEmbed.module.scss
```diff
@@ -1,157 +1,181 @@
...
+  container-name: wrapper;
+}
 
-  position: absolute;
-  width: 100%;
+.content {
+  // Should be able to set width: 100% and let aspect ration determine height,
+  // but then somehow doesn't come out 16:9.
+  // However setting height: 100% seems to work.
+  // Idk why, but it works so... ¯\_(ツ)_/¯
```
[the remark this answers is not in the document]
gabek: Lighten up, copilot.

web/components/ui/OfflineEmbed/OfflineEmbed.tsx
```diff
@@ -17,56 +16,58 @@ export type OfflineEmbedProps = {
...
+        <div className={styles.body}>
+          <Title level={2} className={styles.text}>
+            This stream is not currently live.
           </Title>
-
-          <div className={styles.messageContainer}>
-            <Title level={2} className={styles.offlineTitle}>
-              This stream is not currently live.
-            </Title>
-            <div className={styles.message} dangerouslySetInnerHTML={{ __html: subtitle }} />
```
gabek, line 62: The `subtitle` gets passed in here: https://github.com/owncast/owncast/blob/b50b4959d8e24bc753d86e13fcb46e0f9e0c3836/web/pages/embed/video/index.tsx#L63

It can be the offline message. The offline message is custom, and can literally be any HTML. A video, a gif, whatever. So not using `dangerouslySetInnerHTML` would break that. Also, I'm not sure if this layout is taking that custom offline message into account. We should discuss if we should even be using the custom offline message in the restricted space of the offline embed and instead just be using it on the full web page.

Here's an example of an image placed in there via HTML with the current layout (in storybook).

screenshot_2025-07-09-093832

fortyoneplustwo: The current layout does not take the custom message into account. Notice that inserting an image breaks the 16:9 aspect ratio.  

On one hand I think it makes sense for offline embeds to have a unified layout by disallowing custom messages. Restricting the UI in this way eliminates a lot of potential bugs. 

On the other hand, it's not clear to me what makes the space within the offline embed more restricted than that of the full web page. Is it the follow button? If there is not much of a difference, then our constraint isn't space and we should base our UI decisions on other criteria. Making the message container scrollable with that gradient mask at the bottom is a design that really only makes sense for text content imo. If we accept a custom message then we need to discuss how to handle overflow for all possible elements. 

I'm leaning towards disallowing the custom message at least for this milestone.

gabek: On the full web page there's almost infinite space. Here's an example where the offline banner keeps growing because the user kept adding content.

screenshot_2025-07-20-145055

But if you look at that same server that's been linked on a social network, the space is going to be more restricted, even though it (currently) has the same content in it.

screenshot_2025-07-20-145605

But yeah, some people put images in there, not just text.
screenshot_2025-07-20-145747

fortyoneplustwo: In that case, I think we should opt for a consistent layout of the offline embed by allowing text only. 

If you're okay with this I can go ahead and commit the changes to this PR.

gabek: I was just thinking about not allowing the custom offline text like you said (no images, no custom text, etc) and I thought we could do both. Using media queries, we could show the custom offline text if the viewport was large enough (for when the video embed is placed on a web page) and not use the custom offline text for small viewports like the social media use case we've been discussing above.

What do you think of that?

fortyoneplustwo: How do you decide if a viewport is large enough? Is it a fixed, predetermined value or does it depend on the size of the custom offline message? The latter option is very tricky to realize for 2 reasons:

1. We would first have to render the HTML content in a hidden div to take its measurements, so in the worst case we could be rendering twice which may impact performance.
2. Unlike images, text elements don't have width and height attributes, so we would need to compute each measurement differently depending on the type of the HTML element.

It's more feasible to set a predetermined value so that viewports under `x` width and `y` height do not render the custom offline message. The trade-off in this case is that we possibly preemptively deny messages that would have otherwise fit inside the smaller viewport. And what if the custom offline message has insane dimensions, much larger than the viewport?

It seems to me that the untyped nature of the custom offline message is the source of the problem. It makes sense on the stream's web page, where the embed is allowed to grow with the content, but not so much when it comes to embeds whose dimensions are restricted. This is why I wanted to allow text only.

gabek: > How do you decide if a viewport is large enough?

I think we just pick some common-sense values up front. I don't think we need to be measuring things on the fly.

fortyoneplustwo: Hi Gabe, sorry it took so long to complete this PR. I followed some of copilot's suggestions and applied the common-sense values as we discussed -- I picked `480px`. The changes have been pushed to the feature branch. Please review when you have some time.

gabek: Hey there @fortyoneplustwo! Would you mind opening up another PR for it?

PR #4387, opened 2025-06-17
Add a rule that passwords and stream keys must not contain any dashes

### LINKED ISSUE #3105: Can't update default admin password

image
As seen on the picture, using an example password that meets the requirements, the admin panel shows the update button but is not enabled.
Owncast self hosted using the docker image.

larcher:
This is my attempt at fixing #3105 -- ran into this when setting up Owncast for the first time, and Apple's suggested strong password didn't work because it contains dashes.

### DIFF
--- a/web/utils/config-constants.tsx
+++ b/web/utils/config-constants.tsx
@@ -619,9 +619,13 @@ export const PASSWORD_COMPLEXITY_RULES = [
     message: '- at least one digit',
   },
   {
-    pattern: /^(?=.*?[#?!@$%^&*-])/,
+    pattern: /^(?=.*?[#?!@$%^&*])/,
     message: '- at least one special character: !@#$%^&*',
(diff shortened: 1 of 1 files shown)


larcher:
I wouldn't mind trying to add some tests for this, but could use a hint on where/how to do that :)

gabek:
Hey, thanks for being up for that!

Since the UI component does the work of the actual validation, and you provide the regex, I think splitting out the regex from being inline with the component, and exporting it, then you could write a test throwing some example passwords against it to verify it's doing the right thing. For example, pass in a stream key with a dash and make sure it fails.

gabek:
Works as expected, thank you! @larcher Could you comment on the original issue so it can be assigned to you, so you get credit for closing it?

screenshot_2025-06-29-215008

PR #4392, opened 2025-06-18
fixed status and aria-live

gmarav05:
fixed the status bar aria-live issue


gabek:
Thanks for this! What screenreaders did you test this with?

gmarav05:
I have not tested it @gabek I will try to solve this.

PR #4406, opened 2025-06-30
docs: clarify admin password is not the stream key (#4405)

### LINKED ISSUE #4405: Somewhere in our documentation it still says your stream key = your admin password
This needs to be tracked down and fixed because it's incorrect.

Solomon-mithra:
Update Readme/Doc in Admin Authentication section to:

- Separate “admin password” and “stream key” terminology
- Emphasize stream key is only for streaming software (RTMP/SRT)
- Note that admin password is used for HTTP Basic auth to `/admin`

Closes #4405.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -55,9 +55,10 @@ The pages under `/admin` require authentication to make API calls.
 
 Auth: HTTP Basic
 
-username: admin
+- **Username:** `admin`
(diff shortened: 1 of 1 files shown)


gabek:
ha! That might literally be all it was. I was thinking about our documentation, I never once thought about our Readme. Thank you!

PR #4422, opened 2025-07-09
remove conflicting aria attributes and clean formatting

gmarav05:
This PR removes the unnecessary role="status" and aria-live="off" attributes from the Statusbar component. These values were causing accessibility and screen reader conflicts.


gmarav05:
now updated with suggestion of copilot

gabek:
Looks like there's a linter error. Thanks for continuing to look at this!

gmarav05:
yeah linter issue right.

PR #4430, opened 2025-07-11
Removed aria

### LINKED ISSUE #4382: Accessibility: aria-live implication in web/Statusbar could cause highly annoying UX
See https://github.com/owncast/owncast/blob/508895fd90396c594e2b2e712ba307d89746d655/web/components/ui/Statusbar/Statusbar.tsx#L78-L79

Because `role="status"` is set on the parent container, its aria-live implication will override the below explicit `aria-live="off"`declaration for some assistive technology, updating and pulling attention every second. This could be an issue with some screen readers, magnifying applications, or switch control schemes. 

In the worst case, this would reposition the users' focused zoom areas every second, or make screen readers output "Live for x" over and over again, which of course is highly undesired. 

I suggest removing both the `role` and the `aria-live` attributes, as neither the countdown nor the view count is relevant enough to be observed automatically anyway.

gmarav05:
This PR Removes aria in statusbar.

### DIFF
--- a/web/components/ui/Statusbar/Statusbar.tsx
+++ b/web/components/ui/Statusbar/Statusbar.tsx
@@ -73,13 +73,15 @@ export const Statusbar: FC<StatusbarProps> = ({
       rightSideMessage = `Last live ${formatDistanceToNow(new Date(lastDisconnectTime))} ago.`;
     }
   }
-
+ 
+  // removed aria 
+  // Updated formatting and cleaned up aria attributes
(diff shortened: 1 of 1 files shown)


gmarav05:
I have checked it using prettier now I think it should pass.

gmarav05:
sorry @gabek I forgot about the extra spaces I created. Thank you

gabek:
Thank you!

Issue #3218, opened 2023-07-28
Windows installers so owncase can be distributed by certain windows repositories


luzpaz:
### Share your bug report, feature request, or comment.

Would it be possible for owncast to be available with an installer like .msi or NSIS (instead of being zipped in to an archive?)

Reason: (this is winget https://github.com/microsoft/winget-pkgs)
![Screenshot_20230728_191242](https://github.com/owncast/owncast/assets/4140247/2828d685-435a-4a1c-a5ed-a8c0616967ea)

gabek:
Currently Owncast isn't compatible with Windows at all, so it's unlikely a Windows installer will be built unless we had official support.

If you wish to use Owncast under Windows right now the best bet is to follow these instructions using WSL2: 

https://github.com/owncast/owncast/blob/develop/contrib/owncast_for_windows.md

Maybe some day if we had somebody who wanted to spearhead Windows support it could change. Sorry about that!

PR #4491, opened 2025-08-04
Add native support for Windows

### LINKED ISSUE #4487: Add Native Support for Windows

From prior precedent there's attempts by the community to run owncast on Windows, indicating some demand for this capability.

- https://github.com/owncast/owncast/issues/3218
- https://github.com/owncast/owncast/issues/2814
- https://github.com/owncast/owncast/issues/2864
- https://github.com/owncast/owncast/issues/2813

Golang and nodejs are designed for cross-platform usage if the appropriate measures are used and if the component has no direct dependencies on the operating system (e.g. kernel-specific features or hardware accelerators/devices). From inspection of attempts to run owncast, it appears only the Golang portion needs adjustment to how the ffmpeg process is launched in order to make it more platform agnostic.

A running proof of concept is attached demonstrating that it is feasible to run owncast natively on Windows (no WSL needed).

Image

multippt:
## Double check

- [x] I included a screenshot, logs, or example payload to demonstrate the change.
- [x] This change has actually been tested.
- [x] I read the "Read first" details about filing this PR.

---
This change addresses https://github.com/owncast/owncast/issues/4487

The proposed changes allows launching the ffmpeg process in a platform agnostic manner, removing the dependency on the Unix shell, allowing the Owncast to run on Windows without requiring WSL. Owncast still requires symlink support for the log rotation, though any version of Windows within the past 5 years (Windows 10+, Windows Server 2019+) supports it natively without much user intervention.

The distribution of owncast artifacts for Windows is considered outside the scope of this changeset in order to simplify the PR, this changeset will be focused towards allowing owncast to run on the Windows platform natively. We may consider to re-open https://github.com/owncast/owncast/issues/3218 for changes targeted to address the distribution concerns.

owncast_windows_sdr


gabek:
Nice! The only major concern is a logistical one. If somebody updates either of the test scripts for non-windows, they're not going to update it for Windows, and vice versa. So we'll have two completely different sets of test scripts. Ultimately I'm going to be responsible for maintaining those batch scripts, and I'm going to be unable to do so. People are going to get mad when they expect them to be up to date and they won't be.

gabek:
> Nice! The only major concern is a logistical one. If somebody updates either of the test scripts for non-windows, they're not going to update it for Windows, and vice versa. So we'll have two completely different sets of test scripts.

Maybe a solution here can be to put those Windows test scripts in `contrib/windows` instead.

multippt:
> > Nice! The only major concern is a logistical one. If somebody updates either of the test scripts for non-windows, they're not going to update it for Windows, and vice versa. So we'll have two completely different sets of test scripts.
> 
> Maybe a solution here can be to put those Windows test scripts in `contrib/windows` instead.

To ease concerns on the maintenance burden, I've refactored the test script to one that can run via nodejs, allowing one script to be used for all platforms. (Considering there are other JS scripts in the same folder also used for testing purposes, it's likely not out-of-place.)

gabek:
The existing script is wired up to our CI and tests, so removing the shell script would break all those. And all those other things are fired off by shell scripts as well, so it would all need to be rewritten.

These pieces of infrastructure have worked for some time and I don't really want to change what's working. It would only lead to troubleshooting and reworking things that are working just fine.

Rebuilding what works for this particular script, I don't think it's a fair tradeoff. I still think if you want to have a windows version of the script for those who might want to use it is fine, but we should leave the rest alone.

multippt:
> The existing script is wired up to our CI and tests, so removing the shell script would break all those. And all those other things are fired off by shell scripts as well, so it would all need to be rewritten.
> 
> These pieces of infrastructure have worked for some time and I don't really want to change what's working. It would only lead to troubleshooting and reworking things that are working just fine.
> 
> Rebuilding what works for this particular script, I don't think it's a fair tradeoff. I still think if you want to have a windows version of the script for those who might want to use it is fine, but we should leave the rest alone.

Noted, have reverted the shell script

PR #4541, opened 2025-08-27
Update translation.json

Agent-JY:


gabek:
Hi there. Thanks for wanting to help translate!

Updating the JSON files directly will just be overwritten next time the translation job is run. You'll need to [read through the translation instructions](https://docs.owncast.dev/web-translations#17c69c26ef7980dd80ced3c71b9622a3) in order for it to work. Thank you!

PR #4711, opened 2026-01-01
fix: remove optimistic UI updates in external actions to prevent race condition

### LINKED ISSUE #4347: [Bug] Adding and Deleting URLs can lead to visually cloning on a delete.

Sorry I don't really know how to explain this.

On the latest version of Owncast, I've gone to External Actions, I've added a few URLs.  I then am deleting the links at the top and when I do this the line stays, but it clones the line below it.  This seems purely graphical as a page refresh shows the page appropriately:

Image

Image

Sometimes it just doesn't delete it on the screen as well.

costajohnt:
## Summary

Fixes the visual cloning/duplication bug when adding or deleting external actions in the admin panel.

**The Problem:**
When adding/deleting external actions, the UI would sometimes show duplicated or cloned rows. This was caused by a race condition between:
1. The optimistic `setActions()` call updating local state immediately
2. The `save()` callback updating `externalActions` in the server status context
3. The `useEffect` syncing `actions` from `externalActions`

**The Fix:**
Remove the optimistic `setActions()` calls from `handleDelete` and `handleSave`. Instead, let the server response be the single source of truth - the UI updates only after the context is updated via the `save()` success callback.

This follows the pattern recommended by @gabek in the issue discussion.

## Test Plan

- [x] Lint passes
- [ ] Add multiple external actions
- [ ] Delete actions in quick succession
- [ ] Verify no visual duplication/cloning occurs
- [ ] Verify the table updates correctly after each operation

Fixes #4347

### DIFF
--- a/web/pages/admin/actions.tsx
+++ b/web/pages/admin/actions.tsx
@@ -274,8 +274,10 @@ const Actions = () => {
     actionsData.splice(index, 1);
 
     try {
-      setActions(actionsData);
-      save(actionsData);
+      // Don't optimistically update local state - let the server response
+      // update externalActions in context, which will then update actions
+      // via useEffect. This prevents race conditions causing visual glitches.
+      await save(actionsData);
     } catch (error) {
       console.error(error);
     }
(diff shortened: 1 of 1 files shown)


gabek:
Thanks again, @costajohnt!

PR #4713, opened 2026-01-01
fix: prevent chat container from overflowing viewport in embedded view

### LINKED ISSUE #4492: /embed/chat/readwrite chat view gets cut off vertically

I have the same problem as at https://owncast.online/docs/embed/#embedding-standalone-chat
The input box is always cut off at the bottom, no matter what size I set. The chat is running on a website in an iframe. Is there a solution?

costajohnt:
## Summary

Adds `max-height: 100vh;` to the `.chatContainer` class to prevent the chat from getting cut off vertically in the `/embed/chat/readwrite` view.

## Related Issue

Fixes #4492

## Credit

Thanks to @germainelee for identifying the fix in the issue discussion.

---

## Required checklist

### DIFF
--- a/web/components/chat/ChatContainer/ChatContainer.module.scss
+++ b/web/components/chat/ChatContainer/ChatContainer.module.scss
@@ -30,7 +30,6 @@
   flex-direction: column;
   background-color: var(--theme-color-components-chat-background);
   height: 100%;
-  max-height: 100vh;
   font-size: var(--chat-message-text-size);
 }
 
--- a/web/components/chat/ChatContainer/ChatContainer.module.scss
+++ b/web/components/chat/ChatContainer/ChatContainer.module.scss
@@ -30,6 +30,7 @@
   flex-direction: column;
   background-color: var(--theme-color-components-chat-background);
   height: 100%;
+  max-height: 100vh;
   font-size: var(--chat-message-text-size);
 }
 
--- a/web/pages/embed/chat/readwrite/index.tsx
+++ b/web/pages/embed/chat/readwrite/index.tsx
@@ -62,6 +62,28 @@ export default function ReadWriteChatEmbed() {
           .body-background {
             background: var(--theme-color-components-chat-background);
           }
+          .embed-container {
+            display: flex;
+            flex-direction: column;
+            height: 100vh;
+            height: 100dvh;
+            overflow: hidden;
+          }
+          .embed-container > #chat-container {
+            flex: 1;
+            min-height: 0;
+            display: flex;
+            flex-direction: column;
+          }
+          .embed-container #chat-container #chat-container {
+            flex: 1 1 0;
+            min-height: 0;
+          }
+          .embed-container #chat-container #virtuoso {
+            flex: 1;
+            min-height: 0;
+            height: auto !important;
+          }
         `}
       </style>
       <ErrorBoundary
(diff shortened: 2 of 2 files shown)


gabek:
Thanks for looking into this!

Testing using Storybook, using the read/write embed, (/?path=/story/owncast-chat-embeds-read-write-chat--example) It's significantly better than it was, but it still gets cut off the shorter the window gets.

screenshot_2026-01-03-123027

costajohnt:
Thanks for the feedback @gabek! You were right - the original fix still had issues with shorter viewports.

**Root cause:** The readwrite embed was using a fixed `height="92vh"` for the ChatContainer, but the Header component also takes up space. In short viewports, Header + 92vh > 100vh = overflow.

**New solution:** Changed to a flexbox layout:
- Container uses `display: flex; flex-direction: column; height: 100dvh`
- Header takes its natural height
- Chat container fills remaining space with `flex: 1; min-height: 0`
- ChatContainer now uses `height="100%"` instead of fixed vh

This properly handles any viewport height since the chat area dynamically adjusts to whatever space remains after the header.

Tested in Storybook at 800x400 viewport and the chat now stays within bounds regardless of window size.

costajohnt:
Hi @gabek\! Just checking in - have you had a chance to test the updated flexbox fix? Let me know if there's anything else I can adjust. Thanks\!

gabek:
I haven't yet, but I will. Thank you for your patience!

gabek:
Hey there! It looks like there's some kind of double scrolling going on where you can scroll the chat container like you should, but you can also scroll the container where the input box lives.

capture

costajohnt:
Hi @gabek, thanks for catching the double scrolling issue! I've identified the root cause and pushed a fix.

**Root cause:** The Virtuoso component was receiving an inline `style={{ height: "100%" }}` which made it take 100% of the parent container's height, pushing the chat input field below the visible area and creating a second scrollbar on the outer container.

**Fix:** Added scoped CSS overrides in the embed page that:
1. Set up a proper flex chain through the nested `#chat-container` elements
2. Override Virtuoso's inline height with `height: auto !important` so flex layout controls the sizing (`flex: 1; min-height: 0`)
3. This ensures Virtuoso fills the remaining space after the header and chat input, with only Virtuoso scrolling internally

I also rebased onto the latest `develop` to pick up the authentication changes.

Would appreciate another look when you get a chance!

gabek:
Could you provide a before/after demo of the fix in action? Additionally, the checklist validation is failing.

costajohnt:
## Before/After Demo

I tested the fix locally by running the owncast backend + Next.js dev server with a live test stream to enable chat. I then compared the `/embed/chat/readwrite` view at various viewport heights on both `develop` and this branch.

### The Bug (on `develop`)

When the embedded chat readwrite view is rendered at shorter viewport heights, the chat input gets clipped or completely hidden because the container overflows the viewport:

- **600px height**: Chat input is visible (works fine at larger sizes)
- **350px height**: Chat input is partially clipped at the bottom of the viewport
- **250px height**: Chat input is completely hidden below the viewport edge

### The Fix (on `fix/chat-embed-clean`)

With `max-height: 100vh` on `.chatContainer`, the chat input stays visible and usable at all viewport heights:

- **600px height**: Works fine (same as before)
- **350px height**: Chat input is fully visible — the message area scrolls while the input stays pinned
- **250px height**: Chat input is fully visible — even at very small heights, the layout adapts correctly

### Testing Details

- Tested on Node.js with Next.js dev server
- Used an ffmpeg test stream (`testsrc` + `sine`) to RTMP to enable chat functionality
- Verified at multiple viewport widths (400px) and heights (250px, 350px, 600px)
- All 91 Cypress tests pass in CI

costajohnt:
Hey @gabek. 

Here is the before:

https://github.com/user-attachments/assets/c94a1f50-dc26-4a05-8b73-43bcb74dbb73

And here is the after:

https://github.com/user-attachments/assets/b34ad97c-c8a9-4e65-85fd-0537d92d99a5

gabek:
This works great, thanks so much @costajohnt!

PR #4714, opened 2026-01-01
feat: add readonly-chat class to readonly embed

### LINKED ISSUE #4266: Feature Request: Wrap readonly chat in special class
### feature request

Suggestion:

add or wrap chat container at /readonly path with a class like "readonly-chat" or something, so that custom CSS styling can be targeted specifically at the readonly version of the chat stream.

Rationale: specific styling for OBS, while leaving regular chat attached to video stream alone.

costajohnt:
## Summary
Adds a `readonly-chat` CSS class to the readonly chat embed container, allowing streamers to target it with custom CSS in OBS.

## Problem
Streamers using OBS browser sources want to apply custom CSS styling specifically to the readonly chat embed. Currently, there's no way to distinguish the readonly embed from other chat containers since they share the same class structure.

## Solution
- Added a `readonly` boolean prop to `ChatContainer` component
- When `readonly={true}`, the container div gets an additional `readonly-chat` class
- The readonly embed page now passes `readonly` prop to ChatContainer

## Changes
- `components/chat/ChatContainer/ChatContainer.tsx`: Added `readonly` prop to type definition and function parameters, conditionally applies `readonly-chat` class
- `pages/embed/chat/readonly/index.tsx`: Passes `readonly` prop to ChatContainer

## Usage
Streamers can now use CSS like:
```css
.readonly-chat {
  /* custom styling for readonly embed */
}
```

Fixes #4266


gabek:
Great addition, thanks so much!

Component: https://629132c6e23893003a9e89c5-dhvyenpikr.chromatic.com/?path=/story/owncast-chat-chat-messages-container--example&args=readonly:!true

PR #4717, opened 2026-01-04
fix: preserve chat input text across window resize

### LINKED ISSUE #3615: textbox is cleared every time the window is resized

On Owncast v0.1.2, the chat textbox is cleared every time the window is resized below a certain size (when it hides away the chat behind a button).

My assumption would have been that the chat textbox keeps its contents even when the chat popup is closed and reopened.

costajohnt:
## Summary

Fixes the issue where chat input text is lost when the window is resized across the mobile/desktop breakpoint (768px).

**Root cause:** When switching between mobile and desktop modes, the `ChatContainer` component unmounts and remounts, causing the `ChatTextField` to lose its internal state.

**Solution:** Store the chat input draft in Recoil state (`chatInputDraftAtom`) so it persists across component unmount/remount cycles.

## Changes

- **ClientConfigStore.tsx**: Added `chatInputDraftAtom` to store the draft message
- **ChatTextField.tsx**:
  - Save draft to Recoil on every input change
  - Restore draft from Recoil when component mounts
  - Clear draft when message is sent

## Testing

1. Type a message in the chat input
2. Resize the browser window to cross the 768px breakpoint
3. The message should be preserved in the input field

Closes #3615

### DIFF
--- a/web/components/chat/ChatTextField/ChatTextField.tsx
+++ b/web/components/chat/ChatTextField/ChatTextField.tsx
@@ -1,6 +1,6 @@
 import { Popover } from 'antd';
 import React, { FC, useEffect, useState } from 'react';
-import { useRecoilValue } from 'recoil';
+import { useRecoilState, useRecoilValue } from 'recoil';
 import sanitizeHtml from 'sanitize-html';
 import Graphemer from 'graphemer';
 
@@ -8,7 +8,7 @@ import dynamic from 'next/dynamic';
 import classNames from 'classnames';
 import ContentEditable from './ContentEditable';
 import WebsocketService from '../../../services/websocket-service';
-import { websocketServiceAtom } from '../../stores/ClientConfigStore';
+import { websocketServiceAtom, chatInputDraftAtom } from '../../stores/ClientConfigStore';
 import { MessageType } from '../../../interfaces/socket-events';
 import styles from './ChatTextField.module.scss';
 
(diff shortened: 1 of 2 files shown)


costajohnt:
## Addressing Review Feedback

Thanks for the detailed review! I've pushed a commit addressing most of the suggestions.

### ✅ Addressed

1. **Character count initialization** - Now uses a lazy initializer that properly calculates the grapheme count from the restored draft HTML, falling back to `defaultText?.length ?? 0`.

2. **Missing useEffect dependencies** - Added `inputDraft` to the dependency array. Also added `contentEditable.innerHTML !== inputDraft` check to prevent infinite loops when the draft is updated.

3. **defaultText vs inputDraft conflict** - The restore logic now only runs when `inputDraft.length > 0`, so an empty draft won't overwrite `defaultText`. Non-empty drafts take priority (correct behavior for preserving user input across viewport changes).

4. **Cursor position** - Added Selection/Range API code to move the cursor to the end after restoring the draft.

### ⏭️ Intentionally Skipped

5. **HTML sanitization concern** - The `inputDraft` can only contain content that entered through: (a) user typing (safe text nodes), (b) the paste handler (already sanitized via `sanitizeHtml`), or (c) the emoji picker (controlled HTML we generate). There's no path for unsanitized HTML to enter the draft, so additional sanitization isn't needed.

6. **focusInput dependency** - The empty dependency array `[]` is intentional here. We only want to focus once on mount. Adding `focusInput` would cause focus to trigger on every prop change, which isn't the desired behavior.

DEMO
draft-persistence-demo

gabek:
This works great. Nice solution, thank you again!

PR #4742, opened 2026-01-14
Fix banUser API call missing enabled parameter

### LINKED ISSUE #4724: Inline chat moderation still not working
Chat moderation not working after upgrading to v0.2.4. Trying to ban or hide messages does nothing.

Previous related report: Inline chat moderation request returning 500s https://github.com/owncast/owncast/issues/4357

guitsaru:
The banUser function in ChatModerationService was sending an incomplete request body to the /api/chat/users/setenabled endpoint. The API requires both userId and enabled fields, but only userId was being sent.

This caused the endpoint to return a 400 Bad Request with the error: "must provide userId and enabled state"

Added the missing enabled: false parameter to properly disable the user when banning.

New Request:
```
{"userId":"wk9dSzIDR","enabled":false}
```

Fixes #4724
This is related to 3b25e295e835ea776f2054d94326bf00b847389e

## Double check

- [x] I included a screenshot, logs, or example payload to demonstrate the change.
- [x] This change has actually been tested.
- [x] I read the "Read first" details about filing this PR.

### DIFF
--- a/web/services/moderation-service.ts
+++ b/web/services/moderation-service.ts
@@ -28,7 +28,7 @@ class ChatModerationService {
       headers: {
         'Content-Type': 'application/json',
       },
-      body: JSON.stringify({ userId: id }),
+      body: JSON.stringify({ userId: id, enabled: false }),
     };
 
     await fetch(hideMessageUrl, options);


gabek:
Thank you!

PR #4754, opened 2026-01-19
fix: remove special character requirement from stream keys

### LINKED ISSUE #4690: Remove special characters from auto-generated stream keys as not all broadcast software support them

First of all, thank you for this epic software. I really love it.

I came across a small technicality where Prism Live studio (Streaming App) removes special characters from the stream key and URL fields. Meaning it does not permit it at all in the string. So then I tried to create a manual key on OwnCast without special characters and found that OwnCast forces a special character no matter what. So now I am in a death loop. lol. Would it be possible to maybe add a few check boxes in the server config, where someone could tweak how restrictive you want the key field to be? I understand and support the special character criteria, and would use it 99.99% of the time; however, it would be nice for that 0.01% time to change the criteria if faced with a similar situation like am am dealing with now. I will also understand if you can't implement this, due to potential security risk exposure.

Thank you in advance.

costajohnt:
## Summary

Some broadcasting software (e.g., Prism Live Studio) automatically strips special characters from stream keys, making Owncast-generated keys incompatible. This PR removes the special character requirement from stream key generation while keeping admin password requirements unchanged.

### Changes
- Created separate `STREAM_KEY_COMPLEXITY_RULES` without special character requirement
- Updated `generateRndKey()` to only use alphanumeric characters
- Kept `PASSWORD_COMPLEXITY_RULES` unchanged for admin password security
- Updated tooltip text to reflect new requirements
- Added test verifying keys don't contain special characters

## Related Issue

Fixes #4690

## Test plan

- [x] All existing stream key tests pass
- [x] New test verifies keys don't contain special characters
- [x] TypeScript type checking passes
- [ ] Manual testing in Storybook

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web/components/admin/config/server/StreamKeys.tsx
+++ b/web/components/admin/config/server/StreamKeys.tsx
@@ -1,10 +1,10 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useState } from 'react';
 import { Table, Space, Button, Typography, Alert, Input, Form, message } from 'antd';
 import dynamic from 'next/dynamic';
 import { ServerStatusContext } from '../../../../utils/server-status-context';
 
 import { fetchData, UPDATE_STREAM_KEYS } from '../../../../utils/apis';
-import { PASSWORD_COMPLEXITY_RULES, REGEX_PASSWORD } from '../../../../utils/config-constants';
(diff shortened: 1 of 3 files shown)


gabek:
Thank you!

PR #4800, opened 2026-02-17
feat(video): add enable/disable toggle for stream output variants. Closes #4647

### LINKED ISSUE #4647: Profiles / presets for Video configuration - or toggle

It would be great if it was quicker to change between video configurations in the web gui (as I guess this would be easy to do with config files and scripts).

Imagine doing regular streams of two different formats:

Stream 1: Atmospheric and cinematic music stream – 1080p 25 fps, you want it to feel like a live movie.
Stream 2: Glitchy video art (or gaming or something) where you want high framerate - maybe 480p 60fps - framerate is priority not pixels.

It would be fantastic if I could simply go into the settings and very easily either choose a profile that gave me the 1-3 different configurations related to stream 1, and then next time change over to the 1-2 configurations realted to stream 2.

Another options would be to simply add a toggle box next to the different video configurations that are already there so that I could quickly untoggle some and toggle on some.

- [ ] 1080_25_4000
- [ ] 1080_25_1000
- [X] 480_60_2000
- [x] 480_60_800

RutanshS:
## Required checklist

- [X] I have personally tested these changes and verified they work as intended.
- [X] I understand the code I'm submitting and can explain how it works if asked.
- [X] I included a screenshot, logs, example payload to demonstrate the change, or it really doesn't need it.
- [X] This is a frontend change and it supports [translations](https://docs.owncast.dev/web-translations), or it's not a frontend change.
- [X] The code has been run through the proper linters and/or formatters for the language.
- [X] There is an issue discussing this change and it's assigned to me.

---
Closes #4647 

Adds an `enabled` boolean to stream output variants so admins can toggle variants on/off. Disabled variants are excluded from transcoding and hidden from the public API while remaining visible and workable in the admin UI.

**Backend**
- `Enabled` field added in `StreamOutputVariant` model.
- Backward Compatibility.
- At least one variant has to be enabled.
- Disabled variants excluded from transcoding and hidden in the public api.
- Admin api can see all the variants and their status.

**Frontend:**
- Toggle switch column in admin video variants table.
- If there is only one variant enabled, it cannot be disabled.

**Tests**
- API integration tests for `enabled` field persistence and not allowing all variants to be disabled.
image
image


gabek (changes requested):
I concur with Copilot this time around.

RutanshS:
Yes, I have finished most of the changes. But now that I think about it the "Enabled" column doesn't need a Title, a toggle switch is pretty self explanatory. Let me know if you agree so I can implement that change, otherwise I can implement the copilot suggested fix as well.

gabek:
I think the title is helpful so it's clear what toggling the switch is doing. Thank you!

gabek:
Looks like there might be a couple indexing-related items to look at here.

PR #4820, opened 2026-03-06
fix(web): set page title on chat embed pages

### LINKED ISSUE #4794: Firefox under Arch Linux fails to display title for popped-out chat.
Currently popping out chat when viewing the stream in Firefox on Arch results in the chat window title not being set. I tested in Chromium and confirmed that it appears to appropriately display the title there. I don't know if this is fixable or if it's a Firefox quirk that isn't easily worked around, but I'm hoping it would be pretty trivial. I'm not super well versed in Next.js so I got a bit turned around on what was being set where in the project when I took a look.

Also, really not a big deal, but the title in Chromium is not the most human-friendly (`[site-name]/embed/chat/readwrite`), not sure if the distinction between readwrite and readonly in the title needs to be kept but updating that to be more natural language would be nice as well.

costajohnt:
## Required checklist

---

## Summary

Fixes #4794

The chat embed pages (`/embed/chat/readwrite` and `/embed/chat/readonly`) had no `<title>` tag, which caused:
- **Firefox**: Popup chat windows displayed a blank title bar, making window management difficult
- **Chromium**: Showed the raw URL path (e.g., `[host]/embed/chat/readwrite`) instead of a meaningful name

This PR adds a `<Head><title>` tag to both chat embed pages using the server's configured stream name, following the same pattern used in `OfflineEmbed.tsx`. The title format is `{streamName} Chat`.

## Changes

- **`web/pages/embed/chat/readwrite/index.tsx`**: Added `Head` import and `<title>` tag using the existing `name` from `clientConfig`
- **`web/pages/embed/chat/readonly/index.tsx`**: Added `Head` import, `clientConfigStateAtom` + `ClientConfig` imports, and `<title>` tag

## Test plan

- [x] ESLint passes on both changed files
- [x] Open a stream page → pop out chat → verify the popup window title shows `{streamName} Chat` in both Firefox and Chromium
- [x] Navigate to `/embed/chat/readonly` directly → verify the page title is set

## Screenshots
Screenshot 2026-03-06 at 8 50 21 AM
Screenshot 2026-03-06 at 8 43 31 AM

### DIFF
--- a/web/pages/embed/chat/readonly/index.tsx
+++ b/web/pages/embed/chat/readonly/index.tsx
@@ -1,4 +1,6 @@
 import { useRecoilValue } from 'recoil';
+import Head from 'next/head';
+import { useTranslation } from 'next-export-i18n';
 import { ErrorBoundary } from 'react-error-boundary';
 import { ChatMessage } from '../../../../interfaces/chat-message.model';
 import { ChatContainer } from '../../../../components/chat/ChatContainer/ChatContainer';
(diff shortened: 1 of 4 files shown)


web/pages/embed/chat/readonly/index.tsx
```diff
@@ -1,27 +1,41 @@
...
 export default function ReadOnlyChatEmbed() {
   const currentUser = useRecoilValue(currentUserAtom);
   const messages = useRecoilValue<ChatMessage[]>(visibleChatMessagesSelector);
+  const clientConfig = useRecoilValue<ClientConfig>(clientConfigStateAtom);
   const isChatAvailable = useRecoilValue(isChatAvailableSelector);
 
+  const { name } = clientConfig;
+
+  useEffect(() => {
+    document.title = name ? `${name} Chat` : 'Chat';
```
gabek: If we're going to append `chat` at the end of it (in English), we'll need to translate the entire string.

docs/api/index.html
gabek, line 1: API docs shouldn't need to be changed.

gabek:
This PR still has 28 files changed and can't move forward until that's resolved.

PR #4853, opened 2026-03-26
Move from ts to fMP4 HLS segments

gabek:
See #21.

## Required checklist

**Do not remove this section.** These checkboxes are required and validated.

- [ ] I have personally tested these changes and verified they work as intended.
- [ ] I understand the code I'm submitting and can explain how it works if asked.
- [ ] I included a screenshot, logs, example payload to demonstrate the change, or it really doesn't need it.
- [ ] This is a frontend change and it supports [translations](https://docs.owncast.dev/web-translations), or it's not a frontend change.
- [ ] The code has been run through the proper linters and/or formatters for the language.
- [x] There is an issue discussing this change and it's assigned to me.

Issue #3001, opened 2023-05-08
Add support for VP9 and AV1 streaming with Enhanced RTMP


Conan-Kudo:
### Share your bug report, feature request, or comment.

It should be possible to stream to Owncast using open, royalty-free codecs like VP9 and AV1 rather than using encumbered codecs like H.264.

Now that the [Enhanced RTMP specification](https://github.com/veovera/enhanced-rtmp) is available and [OBS Studio implemented support for it in v29.1](https://github.com/obsproject/obs-studio/pull/8522), Owncast should implement support for the new spec so that OBS can stream in AV1+Opus to Owncast.

gabek:
_Should_ is a strong word. Owncast is a very small project.

Currently Owncast aims to be widely compatible with the most common use cases. If VP9 or AV1 become the standard then the project will support what people are using.

Conan-Kudo:
AV1 _is_ becoming the standard. YouTube supports RTMP streaming AV1 _today_. Twitch support is forthcoming. The leading streaming software supports it out of the box using RTMP, RIST, or SRT.

YellowOnion:
@gabek why was this closed? It seems pretty dismissive to not even leave open as a "community vote".

There's no competing tech, there's no other options if you want to use something that isn't a 15 year old proprietary codec. I just want to ditch the shitty ffmpeg AAC encoder for Opus which IS the modern web standard.

This has nothing to do with AV1, VP9 or HEVC per-se, it's flv/rtmp not supporting *anything* better than ACC+H.264, the ingress side is what is holding *live* streaming back, not the playback side, every browser will happily play webm (VP9+Opus+mkv), most browsers will happily play weird combinations like AC3+H.264+mkv too.

AV1 and Opus are the future for streaming, AV1 encoders are already built in to Intel video cards, HEVC (which is already 10 years old) is poorly designed and no one wants to touch it because of that. Opus is already included in WebRTC, and objectively the best lossy audio codec by every metric, the only encoders that compete are Xiph's Vorbis encoder, and Apple's Quick Time ACC encoder, and there's ZERO chance you're using the latter.

OBS and YouTube support enhanced RTMP already, Please clarify what is an acceptable "standard for adoption".

gabek:
I closed it as not planned because I'm currently not planning on it.

Many people ask to support different codecs for one reason or another. None of them address the most common use case.  I can't simply entertain every request to add things to the project every time it's asked. I'm sorry.

[...]

PR #4873, opened 2026-04-04
Separate codecs from encoders in transcoding pipeline

LogalDeveloper:
## Required checklist

- [X] I have personally tested these changes and verified they work as intended.
- [X] I understand the code I'm submitting and can explain how it works if asked.
- [X] I included a screenshot, logs, example payload to demonstrate the change, or it really doesn't need it.
- [X] This is a frontend change and it supports [translations](https://docs.owncast.dev/web-translations), or it's not a frontend change.
- [X] The code has been run through the proper linters and/or formatters for the language.

---

## Description

This is a stab at laying the foundation for supporting additional codecs, particularly AV1. The transcoding system conflates video compression formats (codecs) with encoding backends (encoders). `libx264`, `h264_nvenc`, `h264_vaapi`, etc. are all treated as interchangeable "codecs" even though they are different encoder implementations of the same H.264 codec. This is acceptable with H.264 currently being the only supported codec, but supporting multiple codecs (H.265, AV1) requires separating the two concepts so each encoder can declare which codecs it supports and adjust its flags per codec.

### What changed

- **Separated codecs from encoders in the transcoder**: Removed the monolithic `codecs.go` that treated `libx264`, `h264_nvenc`, etc. as interchangeable "codecs." Replaced it with distinct `VideoCodec`/`AudioCodec` interfaces (representing compression formats like H.264, AAC) and an `Encoder` interface (representing backends like software, NVENC, VA-API). Each hardware encoder now has its own file implementing the `Encoder` interface, which declares what codecs it supports and returns codec-specific profiles (pixel format, scaler, filters).

- **Per-variant codec tracking and pixel format**: Each HLS variant now carries its own `videoCodec` and `audioCodec`. Pixel format is emitted per variant with a stream specifier (`-pix_fmt:v:N`) rather than globally. This allows HLS variants with different codecs and different pixel formats (e.g. AV1 with yuv420p10le and Opus audio + H.264 with yuv420p and AAC audio).

- **Config migration**: Added a migration (version 4 to 5) that converts old ffmpeg codec names to new encoder type names (e.g. `"h264_nvenc"` to `"nvenc"`) and backfills codec defaults on existing stream output variants.

- **API and OpenAPI spec updates**: `videoCodec` became `videoEncoder`, `supportedCodecs` (flat string list) became `supportedVideoEncoders` (structured objects with encoder type, display name, and supported codecs). Added `videoCodec`/`audioCodec` to the `StreamOutputVariant` schema. The `/admin/config/video/codec` endpoint path is kept as is for now.

- **Frontend**: No functional changes. Just replaced a few mentions of "codec" with "encoder" in the admin UI, and switched the encoder selector from a hardcoded if/else chain to display names derived from the API. Also fixed a `setPendingSavecodec` typo. It will not pick up on any added codecs at this time, it still defaults to H.264/AAC.

- **Tests**: All six encoder test files updated for the new API. Expected ffmpeg command strings updated to reflect per-variant pixel format placement.

### Known issues and things still to do

- **Only H.264 and AAC are implemented.** The architecture supports multiple codecs, but no AV1, or Opus implementations exist yet (ref: #3001). Those are the next step, but depend on the completion of #4853 and #4870 first.
- **No V4L2M2M test file.** Every other encoder has a test file (`transcoder_*_test.go`); V4L2M2M is the only one missing. This gap predates this PR.
- **Hardware encoder parameters need review and testing.** The preset mappings and flags for hardware encoders (VA-API, V4L2M2M, OMX, VideoToolbox, NVENC) were carried forward from the old `codecs.go` exactly as-is without modification, but some don't look right to me. Some may be using incorrect or unsupported values (e.g. x264-style preset names on encoders that don't accept them). I haven't looked into this too deeply and I don't have access to most of those encoders, so they need further review and testing.
- **No codec-encoder compatibility validation.** If a user configures a variant with a codec that the selected encoder doesn't support, the system silently falls back to H.264 rather than warning the user.
- **Frontend "codec" vs "encoder" terminology is not fully resolved.** Some variable names in the frontend still use `selectedCodec`, localization keys still reference "codec", and the encoder descriptions are still in a hardcoded if/else chain. A more thorough terminology pass is needed.
- **`DEFAULT_VARIANT_STATE` uses empty strings for codec defaults.** Should probably be `'h264'` and `'aac'` explicitly rather than relying on server-side fallbacks.


LogalDeveloper:
One clarification given I mentioned #3001 but took another look: this pull request is specifically targeting output _from_ Owncast, not input _to_ Owncast. I am not looking at changing anything with the input yet, the output I am more interested in for my use cases.

LogalDeveloper:
Not abandoned, just not a high priority. This is a bit of a blind shot for how to proceed after #4870 is implemented. I'm running this in production, but it definitely not ready to merge.

PR #4884, opened 2026-04-15
fix: disables the tooltip that should not be shown

IEBqp:
## Required checklist

Broken:
image

Fixed:
image

Rest of the tooltips remain the same.


IEBqp:
Sorry for the improper linting, I ran `npm run stylelint` (which did not find issues in the pr) instead of `npx prettier --check components/video/VideoJS/VideoJS.module.scss --write` or `npm run format`.

PR #4899, opened 2026-05-04
fix(webhooks): include serverURL in fediverse follow event

### LINKED ISSUE #4881: Fediverse follow webhook event is missing `status` and `serverURL` fields
## Description

The `FEDIVERSE_ENGAGEMENT_FOLLOW` webhook event does not include the `status` or `serverURL` fields that all other webhook events include.

Every other webhook sender currently includes `status` and `serverURL` in `eventData`: chat-related events do this by embedding `BaseWebhookData`, while stream status events add the fields explicitly. The follow event bypasses this entirely.

## Current behavior

The follow webhook payload looks like:

```json
{
  "type": "FEDIVERSE_ENGAGEMENT_FOLLOW",
  "eventData": {
    "timestamp": "...",
    "id": "...",
    "name": "...",
    "username": "...",
    "image": "..."
  }
}
```

## Expected behavior

The follow webhook payload should include `status` and `serverURL`, consistent with all other webhook events:

```json
{
  "type": "FEDIVERSE_ENGAGEMENT_FOLLOW",
  "eventData": {
    "status": {
      "online": true,
      "viewerCount": 5,
      "streamTitle": "...",
      ...
    },
    "serverURL": "https://example.com",
    "timestamp": "...",
    "id": "...",
    "name": "...",
    "username": "...",
    "image": "..."
  }
}
```

## Root cause

The follow webhook sender (`SendFediverseEngagementFollowEvent` in `core/webhooks/fediverseEngagement.go`) reuses `activitypub/events.FediverseEngagementFollowEvent` directly as webhook `eventData`, instead of wrapping it in a webhook-specific payload type that includes `BaseWebhookData`.

---

=> After discussion and assignment, I'd be happy to contribute a PR for this issue.

costajohnt:
## Required checklist

---

## Description

Fixes #4881.

The `FEDIVERSE_ENGAGEMENT_FOLLOW` webhook reused `events.FediverseEngagementFollowEvent` directly as the payload, so it had no way for the receiver to identify which Owncast instance sent the event. This adds `serverURL` to the follow payload.

Per the issue discussion (gabek and taintedcypher), the live stream `status` block is intentionally not included on this event. A follow happens regardless of whether the stream is online, so viewer counts and stream title are not meaningful context for the receiver. To keep that boundary clear in code, `WebhookFediverseEngagementFollowEventData` does not embed `BaseWebhookData`; it adds `serverURL` directly.

## Example payload

`go test -v -run TestSendFediverseEngagementEventFollow ./core/webhooks/...` passes. The follow webhook now delivers:

```json
{
  "id": "id",
  "image": "",
  "name": "be",
  "serverURL": "http://localhost:8080",
  "timestamp": "1970-01-01T00:01:12.000000006Z",
  "username": "be@witch.me"
}
```

`go vet` and `golangci-lint run ./core/webhooks/...` both clean.

### DIFF
--- a/services/webhooks/fediverseEngagement.go
+++ b/services/webhooks/fediverseEngagement.go
@@ -24,8 +24,15 @@ func (s *Service) SendFediverseEngagementFollowEvent(iri string) {
 
 func (s *Service) sendFediverseEngagementEventFollow(event events.FediverseEngagementFollowEvent) {
 	webhookEvent := WebhookEvent{
-		Type:      models.FediverseEngagementFollow,
-		EventData: event,
+		Type: models.FediverseEngagementFollow,
+		EventData: &WebhookFediverseEngagementFollowEventData{
+			ID:        event.ID,
+			Timestamp: event.Timestamp,
+			Name:      event.Name,
+			Username:  event.Username,
+			Image:     event.Image,
+			ServerURL: s.serverURL(),
+		},
 	}
 
 	s.SendEventToWebhooks(webhookEvent)
--- a/services/webhooks/fediverseEngagement_test.go
+++ b/services/webhooks/fediverseEngagement_test.go
@@ -22,6 +22,7 @@ func TestSendFediverseEngagementEventFollow(t *testing.T) {
 		"id": "id",
 		"image": "",
 		"name": "be",
+		"serverURL": "http://localhost:8080",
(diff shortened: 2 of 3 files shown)


gabek:
There's some merge conflicts, but if that can be fixed we can go ahead and merge this.

PR #4956, opened 2026-06-18
fix(chat): fire USER_JOINED webhook when join/part messages are disabled

### LINKED ISSUE #4950: Bug report: USER_JOINED webhook doesn't fire when "Join Messages" is disabled in admin.

I am not sure if this is intentional or not, but it would seem as if the USER_JOINED webhook only fires if Join Messages are enabled for stream. This doesn't seem intentional though, as many integrations may wish to receive word when someone joins stream, while the streamer may not wish to have the default message appear in chat. An example being an add-on that replaces the default join message.

costajohnt:
## Required checklist

---

## Description

Fixes #4950.

The admin "show join/part messages" setting (`GetChatJoinPartMessagesEnabled`) gated both the visible chat broadcast and the `USER_JOINED` webhook. Turning the setting off therefore silently stopped the webhook from firing, which breaks integrations that rely on `USER_JOINED` even when a streamer doesn't want join lines cluttering chat.

The `USER_PARTED` path already does the right thing: in `sendUserPartedMessage`, the setting gates only the broadcast while `SendChatEventUserParted` fires unconditionally. This change makes the join path symmetric.

### What changed

In `Addclient`, the gate that decided whether to call `sendUserJoinedMessage` was seeded from the display setting and also forced off for two cases (an additional client for an already-connected user, and a quick reconnect before the part was sent). That gate is renamed `isNewJoin` and seeded `true`; it still suppresses both cases (so `USER_JOINED`/`USER_PARTED` webhook pairs stay balanced), but it no longer consults the display setting.

The display setting is now checked inside `sendUserJoinedMessage` to gate only the `Broadcast`, mirroring `sendUserPartedMessage`. The webhook fires for every genuine join.

### Why it won't regress join/part balance

- New single user: `isNewJoin` true, JOIN fires; on full disconnect the part timer fires PART.
- Second client (extra tab): suppressed (not a new join); that client's disconnect is also suppressed in `handleClientDisconnected`. Neither fires.
- Quick reconnect within the part-timer window: the pending part is cancelled and the rejoin is suppressed. Neither fires.
- Messages disabled: webhook fires, broadcast suppressed (the bug being fixed).

### Logs / verification

I am not assigned to #4950 yet, but the change references it and I am happy to have it assigned for release credit.

Added `services/chat/server_test.go` with a regression test asserting the `USER_JOINED` webhook fires whether the join/part display setting is on or off, while the visible broadcast remains gated on the setting. It uses a datastore-backed webhooks service with an `httptest` capture endpoint.

```
$ go test ./services/chat/
ok  	github.com/owncast/owncast/services/chat

$ go vet ./services/chat/   # clean
$ gofmt -l services/chat/   # no output (gofumpt-clean too)
```

### DIFF
--- a/services/chat/server.go
+++ b/services/chat/server.go
@@ -138,24 +138,30 @@ func (s *Service) Addclient(conn *websocket.Conn, user *models.User, accessToken
 		ConnectedAt: time.Now(),
 	}
 
-	shouldSendJoinedMessages := s.configRepository.GetChatJoinPartMessagesEnabled()
+	// isNewJoin tracks whether this connection is a genuinely new chat join, as
+	// opposed to an additional client for an already-connected user or a quick
+	// reconnect. The admin "show join/part messages" setting is intentionally
+	// NOT consulted here: it governs only the visible broadcast in
+	// sendUserJoinedMessage, not whether the join occurred (so webhooks still
+	// fire when join/part messages are hidden). See #4950.
+	isNewJoin := true
 
-	// If there are existing clients connected for this user do not send
-	// a user joined message. Do not put this under a mutex, as
-	// GetClientsForUser already has a lock.
+	// If there are existing clients connected for this user this is not a new
+	// join. Do not put this under a mutex, as GetClientsForUser already has a
+	// lock.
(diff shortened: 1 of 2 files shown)


gabek (approved):

gabek:
Awesome, thank you so much!

PR #4961, opened 2026-06-21
feat: webhook authentication

### LINKED ISSUE #4951: Feature request: Webhook authentication

Webhooks should have some form of authentication to verify that they are being sent by the Owncast server and not being spoofed. The most simple solution would be to create some sort of bearer token for the webhook that can be verified by the webhook listener, but I'm sure better solutions exist; HMAC signatures are used by many sites that support webhooks (including Github!). 

See [the Wikipedia article about Webhooks and Authentication](https://en.wikipedia.org/wiki/Webhook#Authenticating_the_webhook_notification) to read more about webhook authentication methods.

AnonymousXC:
## Required checklist

---

## Description
1. Added a `webhookSecret` key in config datastore 
2. Added a new header `owncast-authentication` in the webhook post request
3. Added the logic of webhook auth at `services/webhooks/auth.go`
4. Added update webhook secret field at `/admin/config/server/

closes #4951


AnonymousXC:
@gabek Could you please take a look at this PR? I'll open an pr to update the docs too. Thanks!

gabek:
I just want to bring up one thought. Currently with this solution there's a single webhook secret that all webhooks share. So if one needs to rotate it, all the webhooks would need to be updated. It's also _slightly_ a security thing if a bunch of 3rd parties know each other's secret.

I wonder if it would make sense to have a per-webhook secret. What do you think?

AnonymousXC:
> I just want to bring up one thought. Currently with this solution there's a single webhook secret that all webhooks share. So if one needs to rotate it, all the webhooks would need to be updated. It's also _slightly_ a security thing if a bunch of 3rd parties know each other's secret.
> 
> I wonder if it would make sense to have a per-webhook secret. What do you think?

I think that would be even better. Keeping different webhook secret for different webhook url will surely increase the security and flexibility. I'll update it soon.

AnonymousXC:
@gabek I was thinking whether to encrypt the webhook secret before saving it to the database(just like the datastore/config repository), it would be better to encrypt the webhook secret for better security but i think it will also add an extra overhead of decryption on each webhook call. What are your views on this trade off? or, is there a better way to handle this problem that I might be overlooking?. Thanks!

gabek:
I don't think it's needed to encrypt it. I think it's more akin to how we treat stream keys than how we treat the admin password.

AnonymousXC:
> I don't think it's needed to encrypt it. I think it's more akin to how we treat stream keys than how we treat the admin password.

Got it, that distinction makes sense. The PR is ready from my end and open to any changes or feedback you have.

gabek:
Thank you!

PR #4968, opened 2026-06-24
feat(web): include chat userId in external action query params

### LINKED ISSUE #4953: Feature request: include chat userId in External Action query parameters

Currently, when opening an external action, owncast passes along the fqdn for the owncast instance as well as the user's display name in chat as query parameters. However, chat display names are not guaranteed unique. For this reason I think it would be useful to include the user's unique chat userId as a parameter as well. This would enable better functionality with add-ons, as, from what I've seen, there is no way to fetch a user's ID from the client-side.

costajohnt:
## Required checklist

---

## Description

Fixes #4953

When an external action opens, Owncast already appends the instance URL and the chat display name (`username`) as query parameters. As noted in the issue, display names are not guaranteed unique, so they aren't a reliable identifier for add-ons that want to key off the current chat user.

This appends the chat user's stable `id` as a `userId` query parameter alongside the existing `username`, inside the same `if (currentUser)` block in `web/components/ui/Content/Content.tsx`. The id is already present on the client (`currentUser.id`, the same value passed as `chatUserId` to the chat container), so this is a small, self-contained change.

Resulting external-action URL params: `instance`, `username`, `userId`.

## Verification

- `userId` is appended via `URLSearchParams.append`, which handles encoding.
- The value comes from the existing `currentUser.id` field (`CurrentUser.id: string`), already used elsewhere in the same component.
- Verified with `eslint` and `tsc --noEmit` (both clean). The change adds no user-facing strings, so there is nothing new to translate. No visual change, so there is no screenshot to add; the effect is the additional `userId` parameter on the external-action URL.

### DIFF
--- a/web/components/ui/Content/Content.tsx
+++ b/web/components/ui/Content/Content.tsx
@@ -161,10 +161,13 @@ export const Content: FC = () => {
       updatedUrl.searchParams.append('instance', currentBrowserWindowUrl);
 
       if (currentUser) {
-        const { displayName } = currentUser;
+        const { id, displayName } = currentUser;
 
-        // Append url and username to params so the link knows where we came from and who we are.
+        // Append url, username and userId to params so the link knows where we
+        // came from and who we are. Display names are not unique, so userId
+        // gives external actions a stable identifier for the chat user.
         updatedUrl.searchParams.append('username', displayName);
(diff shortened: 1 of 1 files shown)


gabek:
Thank you!

Issue #150, opened 2020-09-16
Schedule Livestreams


graywolf336:
This is a big future thinking request/idea that probably should be split into several or handled as a milestone...or maybe it's too niche and very specific use case that we can dismiss.

I would like to schedule livestreams and be able to link to the individual streams. This would then tie into the recording feature where the stream video endpoint turns into the recorded output, very similar to how YouTube handles it.

---
For an example use case, I will use church services since that has become a big thing with the onslaught of covid-19.

Church A has three services and live streams each of those services. Church A also has an app which takes the stream endpoint and embeds it natively inside the app (iOS and Android) instead of taking them to a web browser or streaming via YouTube. The church wants to have a list of services/events and be able to link to the future and past streams, this way they can schedule it and also have statistics from the app.

To implement this, I think that each livestream event would have its own unique HLS video endpoint. When the event is scheduled in the future, a special graphic would be displayed (or the regular offline could work) with the time on it (they would provide it). Then whenever the stream happens at that time, the video would be there along with the regular HLS endpoint. After the livestream ends, the HLS could be replaced with the recorded video.

Thoughts?

gabek:
Sounds cool, I could totally see this being useful.  I see it as adding the notion of events, and each event has its own metadata and optionally a linked recording.

The biggest challenge I see here is is the web admin UI needed to manage all of it.  It seems like the piece of the puzzle that's required (even more so than recordings, as you could have scheduled events that aren't recorded).  And I'll be honest, I'm bad at building user-friendly forms :). I'd almost rather build a command line tool, but I know that's not not a very good long-term plan.

Going to link to some things here that are already filed that tie into some of this:

Rework of HLS playlist generation (required for recordings) #9
Recordings #102 
Admin dashboard #114

gingervitis:
I like UI. :) I can help with that!

gabek:
Here's a couple of examples of scheduled events.  They both encourage chat before the event starts.  I guess from an engagement standpoint it's a good idea, and because they both have authenticated users they can shut down bad actors.  I always felt no good can come of the chat being available when a stream isn't happening, but that's because we're not identifying people.

YouTube's scheduled event:

![image](https://user-images.githubusercontent.com/414923/94059139-de9d6480-fd96-11ea-955f-8c3dfae8e2e4.png)

[Crowdcast](https://www.crowdcast.io/features) scheduled event:

[...]

PR #5007, opened 2026-07-07
Scheduled streams: events data layer and recurrence engine

gabek:
This is the first slice of the scheduled streams work for the v0.4.0 Schedules milestone (#150, #2371). It adds the storage and recurrence foundation only. There is no API or UI yet, so nothing is user-visible until those land in follow-up PRs.

- Add `stream_event_series` and `stream_events` tables via a goose migration. A recurring schedule stores an RFC 5545 recurrence value (a DTSTART carrying an IANA timezone plus an RRULE) and gets expanded into concrete occurrence rows on a rolling horizon. One-off events are occurrence rows with no series.
- Add a `scheduleeventsrepository` following the existing repository pattern, with sqlc-generated queries.
- Add a recurrence engine in `services/schedule` built on rrule-go. Materialization is idempotent, a unique (series, original start) index means re-running it never duplicates a row, never resurrects a cancelled occurrence, and never re-creates one that was moved.
- Series edits only regenerate future occurrences that were never federated. Rows that already went out to followers keep their identity, which is what the ActivityPub Update/Delete work will need later.
- Guard against pathological rules. Values without a DTSTART are rejected since rrule-go would silently default the start to "now" on every parse and break idempotency, and expansion is capped so something like FREQ=SECONDLY can't fill the table or burn unbounded CPU.
- Embed IANA timezone data (`time/tzdata`) so named-zone expansion works on Windows, which has no system tzdata. The timezone travels inside the rule on purpose: "Monday 18:00 in America/Los_Angeles" stays 18:00 wall time across DST transitions while the UTC instant shifts, and there's a test asserting the exact instants across the March 2027 spring-forward to prove it.

Added unit tests for all of the above, including the cancel and move lifecycles, the window queries the future scheduler tick will use, and negative controls for bad rules and unknown timezones.

```
go test ./persistence/scheduleeventsrepository/... ./services/schedule/...
ok  	github.com/owncast/owncast/persistence/scheduleeventsrepository	0.059s
ok  	github.com/owncast/owncast/services/schedule	0.120s
```

I also verified the planned ActivityPub Event shape against a real Gancio instance ahead of the federation work, so the federation-facing fields here (`federated_at`, `reminder_sent_at`, per-occurrence short ids that become object IRIs) are known to be the right shape.

## Required checklist


gabek:

- Both TestMains now pass the exit code through os.Exit.
- Removed the SetGlobalInstance/Get accessors from the new repository per the service migration pattern in AGENTS.md. The repository is now only reachable through constructor injection, and the follow-up API PR wires it through main.go.
- The occurrence insert now uses a targeted ON CONFLICT (series_id, original_start) DO NOTHING instead of INSERT OR IGNORE, so a primary key collision surfaces as an error instead of silently reporting the slot as pre-existing.

Also fixed the CI failure in this run. It was an unrelated pre-existing flake: the follower pagination tests insert 100 rows fast enough to share second-precision created_at timestamps and then expect insertion order back from an ORDER BY created_at DESC query, so a second boundary landing mid-loop on a slow runner shuffled the order. The fixture now pins distinct created_at values so the ordering is deterministic.

PR #5008, opened 2026-07-07
Scheduled streams: admin and public APIs plus the scheduler service

gabek:
This is the second slice of the scheduled streams work (#150, #2371), stacked on #5007. It adds the whole HTTP surface and the background scheduler. Everything stays invisible to viewers until the admin flips the new schedule toggle.

Admin API, spec-first in openapi.yaml with generated stubs:

- `GET /api/admin/schedule` returns every series and the occurrences worth managing.
- `POST /api/admin/schedule/event` creates or updates a one-off event (a start time) or a recurring series (an RFC 5545 recurrence value). Validation is strict: name required, start and recurrence are mutually exclusive, unknown timezones and bad rules get a 400, and a rejected update never half-applies. Omitted optional fields keep their stored values.
- `POST /api/admin/schedule/event/delete` deletes, or with `cancel: true` cancels. A cancelled occurrence is kept and shown as cancelled, and series edits never resurrect it.
- `POST /api/admin/schedule/preview` expands a recurrence rule server-side and returns the next five occurrences, so the upcoming admin form previews exactly what the materializer will produce. One parser owns both paths.
- `POST /api/admin/config/schedule/enabled` and `.../remindermessage` config toggles, both readable back via `/api/admin/serverconfig`.

Public API:

- `GET /api/schedule` with optional `from`/`to` range params, defaulting to the past week plus the next 90 days. Serves past occurrences too since the viewer calendar will browse backwards. Always an empty list while the feature is off.
- `/api/config` gains `schedule: { enabled }`.
- `/api/status` gains a small `scheduledEvent { id, name, startTime, chatOpen }` when something is current or upcoming. It keeps serving an event that has started but not run past its duration, so the future chat window won't slam shut when a stream starts a few minutes late.

Backend:

- A new schedule service ticks once a minute, materializing occurrences from recurring series out to a 30 day horizon. Every tick is idempotent and driven purely from table state, so restarts lose nothing. A short-lived cache answers the status poll, invalidated immediately on admin mutations.

Tests: a new `013_scheduledevents.test.js` integration spec (35 tests) covering defaults, auth, validation, mutations, range params, and the toggle staying inert, plus repository tests for the current-or-upcoming query including its end-time boundary. The spec honors an `OWNCAST_TEST_URL` override so it can run against an instance on any port, defaulting to 8080 for CI.

```
Test Suites: 1 passed, 1 total
Tests:       35 passed, 35 total
```

## Required checklist


gabek:

- The schedule service is now constructor-injected per the service migration pattern: New(Deps) with the repository injected, Start/Stop as methods, no package-level state, and no persistence Get() calls. main.go composes it and hands it to the handler and admin dependency sets.
- GetUpcomingEvent now caches failed lookups too, so a transient database error backs off for the cache TTL instead of re-querying and logging on every status poll.

Issue #2370, opened 2022-11-23
Admin interface for managing schedules


gabek:

germainelee:
Please share the components needed for this issue so that I can design accordingly. Thanks.

gabek:
For the admin, we try to keep components as simple as possible. Most of the time, we use the components directly from our UI library and make small tweaks only where necessary. That will also be the case here and save all the UI work for the primary frontend.

New events admin page:

- Enable/disable events feature
- List of events you can edit and delete from: https://ant.design/components/table
- Add event button
- Add event date picker: https://ant.design/components/date-picker
- Add event time picker: https://ant.design/components/time-picker
- Add event name and description: https://ant.design/components/input
- Calendar view of events as an overview: https://ant.design/components/calendar

The table view and calendar view are probably switchable via tabs https://ant.design/components/tabs, and both of them shouldn't be showing at the same time.

<blockquote><img src="https://gw.alipayobjects.com/zos/rmsportal/rlpTLlbMzTNYuZGGCVYM.png" width="48" align="right"><div><strong><a href="https://ant.design/components/table"> Table - Ant Design</a></strong></div><div>An enterprise-class UI design language and React UI library with a set of high-quality React components, one of best React UI library for enterprises</div></blockquote>
<blockquote><img src="https://gw.alipayobjects.com/zos/rmsportal/rlpTLlbMzTNYuZGGCVYM.png" width="48" align="right"><div><strong><a href="https://ant.design/components/date-picker"> DatePicker - Ant Design</a></strong></div><div>An enterprise-class UI design language and React UI library with a set of high-quality React components, one of best React UI library for enterprises</div></blockquote>
<blockquote><img src="https://gw.alipayobjects.com/zos/rmsportal/rlpTLlbMzTNYuZGGCVYM.png" width="48" align="right"><div><strong><a href="https://ant.design/components/time-picker"> TimePicker - Ant Design</a></strong></div><div>An enterprise-class UI design language and React UI library with a set of high-quality React components, one of best React UI library for enterprises</div></blockquote>
<blockquote><img src="https://gw.alipayobjects.com/zos/rmsportal/rlpTLlbMzTNYuZGGCVYM.png" width="48" align="right"><div><strong><a href="https://ant.design/components/input"> Input - Ant Design</a></strong></div><div>An enterprise-class UI design language and React UI library with a set of high-quality React components, one of best React UI library for enterprises</div></blockquote>
[...]

PR #5012, opened 2026-07-07
Scheduled streams: admin schedule management page

gabek:
This is the third slice of the scheduled streams work (#150, #2370), stacked on #5008. It adds the admin page for managing the schedule.

- New Schedule page in the admin sidebar with the feature toggle, the Fediverse reminder message field, a table of upcoming events and a table of recurring schedules.
- Add event modal handles both one-off events and weekly recurring schedules. The recurrence form is built from plain antd primitives (a repeats select, weekday chips, native date and time inputs) and maps onto the RFC 5545 rule the backend stores. The admin never sees RRULE text.
- Live preview under the form comes from the server-side preview endpoint, so the dates shown are expanded by the exact same parser the scheduler materializes with. One source of truth.
- Editing round-trips safely: rules this form didn't create (different frequencies, INTERVAL, COUNT, or unusual UNTIL formats from the raw API) degrade to details-only editing instead of silently rewriting the rule on save.
- Timezone handling is explicit. A series stores its IANA zone inside the rule, one-off events convert the picked wall time to UTC through a small Intl-based helper, so there's no timezone library and nothing new for the antd upgrade to migrate. That's also why the date and time inputs are native rather than antd's moment-based pickers.
- All new strings are localized.

Unit tests cover the recurrence compose/parse round-trip (including the negative-offset UNTIL edge that would otherwise drift an end date one day per edit), the wall-time conversion math across PST/PDT/half-hour zones and a DST gap, and the degrade-to-details cases.

```
Test Suites: 1 passed, 1 total
Tests:       26 passed, 26 total
```

Verified end to end in a real browser against a live backend: creating a weekly series through the modal with the preview showing, saving, and watching the occurrences appear in the tables. Screenshots below.

## Required checklist

## Screenshots

The schedule page with a one-off event and a weekly recurring schedule:

The add event modal in weekly mode with the server-side occurrence preview:


gabek:

- The When column and the recurrence preview now render wall times in the event's own timezone instead of the browser's, so the value always matches the zone label next to it.
- The sidebar label, table column titles and the duration unit are now localized.
- The weekday chips now get their labels from Intl.DateTimeFormat in the browser's locale rather than a hardcoded English table.

One thing I'm deliberately leaving for now: describeRecurrence still builds its "Weekly on Mon at 18:00" summary in English. Composing that sentence from localized parts is a bigger job than it looks and the admin has other English-only surfaces today, so I'd rather revisit it separately than fake it here.

gabek:
Localized recurrence summary. The IANA timezone now renders as text rather than an escaped interpolation.

Schedule recurrence summary

PR #5015, opened 2026-07-08
Add a central scheduler service for recurring background work

gabek:
This adds a central scheduler service for recurring background work, so periodic internal tasks stop hand-rolling their own tickers, goroutines and shutdown channels. It's independent infrastructure and based directly on develop, so it can land regardless of the scheduled streams work. Once both are in, the scheduled streams materializer moves onto it as a small follow-up.

The engine is gocron v2, which is actively maintained and widely used, wrapped in a small `services/scheduler` facade so the rest of the codebase depends on one Owncast-shaped API instead of a third-party surface. Every registered job gets the same conventions for free:

- Overlap protection. A run that outlives its interval causes following ticks to be skipped, never stacked or run concurrently.
- Panic recovery. A panicking run is logged and the job runs again next tick, it can't take the process down.
- Cooperative shutdown. Jobs receive a context cancelled at Stop, and Stop waits for in-flight runs to finish.
- A name in the logs and a Statuses listing for a future admin or debug surface.

Adding a new job is a service method plus one registration line in main.go. Jobs are expected to be idempotent sweeps that read their own state, the same contract the old tickers had.

Two tickers migrate in this PR to prove the shape:

- The stale featured servers check. This one also moves off the deprecated package-global pattern to constructor injection (removing two persistence Get() callers), and its federation-enabled gate moves inside the job body, which fixes a small pre-existing quirk: enabling federation at runtime used to leave the checker dead until a restart.
- The chat message pruner.

Good follow-up candidates once this lands: the scheduled streams materializer (#5008), the hourly DB backup, stream stats saving, the expired auth request pruners, follower validation and the metrics polling loops. Lifecycle-bound loops (thumbnail generation, the live outbox ping, websocket keepalives) deliberately stay where they are, they're not periodic app maintenance.

The facade contract is pinned by unit tests covering registration validation, repeated firing, run-at-start, overlap protection, panic recovery, shutdown semantics and late registration.

```
go test ./services/scheduler/ ./services/chat/
ok  	github.com/owncast/owncast/services/scheduler	1.337s
ok  	github.com/owncast/owncast/services/chat	0.111s
```

Boot verified with both jobs registered and running.

## Required checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.103.3
 	github.com/extism/go-sdk v1.7.1
 	github.com/go-chi/chi/v5 v5.2.4
+	github.com/go-co-op/gocron/v2 v2.21.2
 	github.com/go-fed/httpsig v1.1.0
 	github.com/gobwas/glob v0.2.3
 	github.com/gorilla/websocket v1.5.3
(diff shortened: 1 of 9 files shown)


gabek:

- Register after Stop now returns an explicit error instead of relying on the engine's behavior, with a test locking it in. Worth noting the underlying library actually no-ops silently there rather than erroring, so the explicit guard is what makes the contract real.
- The staleness sweep checks the context between servers, so shutdown isn't delayed by a long server list.
- Fixed the stale chat Start comment about persistence setup.

PR #5106, opened 2026-08-03
Report plugin mutation errors to plugins

### LINKED ISSUE #5092: Feature Request (Plugins): owncast_add_actions should error if payload represents misconfigured ActionButton

`ActionButton`s can either have `html` or `url` as a field for building them. Because it can be either, both are nullable. This means that the user can submit a misconfigured action button that has _neither_ `html` nor `url`, or that has _both_ `html` and `url`. When using `owncast_add_actions` the server logs an error in this case, but the plugin has no way of knowing when such an error occurs.

In order to notify the plugin that the `ActionButton` was configured incorrectly, `owncast_add_actions` should return `{ error?: string }`, and in the event that the call was successful, `error?` is `null`, and in the event that there was an error, then `error?` is the error message describing what the issue was. (Other error types should exist as well, which I will iterate on after the next paragraph.)

Another possible fix would be to replace the `html` and `url` fields with a single `content` field, and then have a new `contentType` field which can either be "html" or "url". This would prevent both types of misconfigured action buttons. The host would no longer have XOR for `html` or `url`, and instead only have to verify that `contentType` is a valid string. I think this is the better solution, as it makes the `ActionButton` payload more closely resemble how `ActionButton`s are manually created in the admin UI. In this scenario `owncast_add_actions` would still return `{ error? }`, but instead error? would be populated when the author creates a different type of error. Types of errors may include:

- Invalid contentType field.
- Malformed URL.
- Malformed HTML body.
- Malformed icon URL.

gabek:
Plugins can now tell when a runtime action was rejected instead of only leaving an error in the Owncast logs.

Changing these return values updates the shared WebAssembly host ABI. I included the other non-fire-and-forget mutation calls in the same engine rebuild so validation, permission, and persistence failures are not silently discarded.

- Return `{}` on success and `{"error":"..."}` on failure for runtime actions, KV writes, chat moderation, and user moderation
- Pass service errors back through the plugin host boundary
- Rebuild the JavaScript and Python engines with the updated SDK wrappers
- Cover invalid action payloads and missing `ui.modify` permission in both runtimes

For example, adding an action without `url` or `html` now throws with this host result:

```json
{"error":"actions[0]: exactly one of url or html is required"}
```

I ran the focused plugin and plugin-host Go tests, the SDK host-runtime tests against this checkout, and built and loaded the JavaScript and Python action-button examples.

Fixes #5092

## Required checklist


gabek:
The automated plugin test job currently fails because it clones the SDK's main branch, which still has the old void mutation signatures. The matching SDK change is owncast/plugin-sdk#12. Both JavaScript and Python example matrices pass there against this Owncast branch. This core check can pass after the SDK PR merges.

PR #5117, opened 2026-08-09
feat(plugins): namespace emitted custom events

### LINKED ISSUE #5093: Feature Request (Plugins): Host should namespace custom event hooks to avoid event spoofing

Currently, plugins are expected to namespace custom event types on their own. This not only means that a function can fail to do so and create unnamespaced or inappropriately namespaced event types, but they can impersonate other plugins' event hooks. For instance, if a plugin `plugin_one` has an event hook called `plugin_one.foo`, a plugin `plugin_two` can register an event hook under the same name, and events that were supposed to be sent to one plugin get sent to both.

To solve both of these issues, the host should automatically namespace plugins with the plugin's slug. For example, `plugin_one` would simply register an event hook called `foo`, but then once the host reads it in the manifest, it becomes `plugin_one.foo`. (`owncast_emit_event` would still expect the plugin author to provide the namespace of the plugin being called for obvious reasons.)

This would mean that it is impossible for one plugin to spoof another plugin's custom event hooks, and it would mean that plugins are forced to all namespace their custom events in a predictable way.

gabek:
Closes #5093

## The problem

Plugins namespace their own custom events today, on the honor system. Nothing stops `plugin_two` from emitting `plugin_one.foo`, so a subscriber has no way to know who really sent an event. Our own examples don't follow the convention: `relay` emits a bare `announcement.broadcast`.

## The fix

The host now owns the namespace. `owncast_emit_event` prefixes every emitted name with the calling plugin's slug, resolved host-side from the wasm call, so the guest can't supply or override it:

```js
// in the "relay" plugin
owncast.events.emit("announcement.broadcast", payload);
// dispatched as: relay.announcement.broadcast

// subscriber, in any other plugin
on: {
  "relay.announcement.broadcast"(payload) { ... }
}
```

Two consequences, which are exactly the two things #5093 asks for:

1. **No spoofing.** An event named `relay.*` can only have come from `relay`. If another plugin tries `emit("relay.announcement.broadcast")`, it dispatches as `attacker.relay.announcement.broadcast` and no subscriber of the real name hears it.
2. **Uniform namespacing.** Every custom event is `<slug>.<name>`, enforced rather than documented.

Core events are protected too. Slugs like `chat` are legal, so a plugin slugged `chat` emitting `message.received` would otherwise compose the real `chat.message.received`. The host rejects any composed name in `reservedEventTypes`.

Subscriptions stay literal: you subscribe to the sender's full `<slug>.<event>` name, which is what keeps plugin-to-plugin composition working.

## Relationship to the issue text

The issue suggests prefixing the *subscription* side (a plugin registers `foo`, the host stores `plugin_one.foo`) and keeping emit fully qualified. That variant gives each plugin exclusive ownership of its hook names, but it leaves emission anonymous: any plugin could still emit `plugin_one.foo` with a fake payload, and the receiver couldn't tell. Prefixing at emit closes that instead, and subscribing to `relay.announcement.broadcast` still gives the receiver a single trusted sender.

Not addressed here: subscriptions are public. Any plugin can listen to any custom event, the same as core events. Private or targeted delivery is an access-control feature, happy to open a separate issue if that's wanted.

## Tests

`services/plugins/event_namespacing_test.go`, driven through the real JS and Python engines:

- `TestEmitCannotSpoofAnotherPluginsEvents` runs the issue's scenario: a spoof attempt surfaces only under the attacker's own name.
- `TestEmitIsNamespacedByCallerSlug` asserts the exact delivery set for plain, dotted, impersonating, and core-colliding names.
- `TestEmitCannotComposeCoreEventName` covers the slug-`chat` composition case.
- `TestEmitFromPythonIsNamespaced`, `TestEmitWithoutPermissionDispatchesNothing`, `TestSubscriptionsAreNotRewritten`.

`go vet ./services/plugins` and `go test ./services/plugins -count=1` pass. Mutation-checked: removing the prefix or disabling the reserved-name check fails the corresponding tests.

## Follow-up in the SDK

owncast/plugin-sdk#15 updates the docs, types, and relay/announcer examples. It can only land after a release carrying this change is pinned in the SDK's `host-runtime/go.mod`.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -146,7 +146,12 @@ Owncast and plugins communicate through a shared dispatcher
   plugins' `filterChatMessage` handlers run on inbound messages before they're
   broadcast (redact, drop, rewrite).
 - Plugins can also **emit** custom events to each other; `env.Emit` is wired to
-  the runtime's live dispatcher after the manager starts.
+  the runtime's live dispatcher after the manager starts. `hostEmitEvent`
+  prefixes the emitted name with the calling plugin's slug (resolved by the
+  host, never supplied by the guest), so a plugin can only publish under
+  `<its-own-slug>.` and can neither forge a core event nor impersonate another
+  plugin. Subscriptions stay literal, so listening to
+  `other-plugin.something` still works.
 
 ## Lifecycle, persistence, and HTTP
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -149,8 +149,10 @@ Owncast and plugins communicate through a shared dispatcher
   the runtime's live dispatcher after the manager starts. `hostEmitEvent`
   prefixes the emitted name with the calling plugin's slug (resolved by the
   host, never supplied by the guest), so a plugin can only publish under
-  `<its-own-slug>.` and can neither forge a core event nor impersonate another
-  plugin. Subscriptions stay literal, so listening to
+  `<its-own-slug>.` and cannot impersonate another plugin. Composed names that
+  would equal a built-in event (a plugin slugged `chat` emitting
+  `message.received`) are rejected against `reservedEventTypes`, so core
+  events can't be forged either. Subscriptions stay literal, so listening to
   `other-plugin.something` still works.
 
 ## Lifecycle, persistence, and HTTP
--- a/services/plugins/dispatcher_test.go
+++ b/services/plugins/dispatcher_test.go
@@ -8,28 +8,6 @@ import (
 	"time"
 )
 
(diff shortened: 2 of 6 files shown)


gabek:
All four review comments flagged the same real gap: a plugin slugged `chat` emitting `message.received` composed the exact core event name `chat.message.received`, which would have fanned out to every subscriber as host-originated. Fixed in 5353939a5b: `reservedEventTypes` is restored and checked against the composed name after prefixing, with an engine-level regression test (`TestEmitCannotComposeCoreEventName`) proving the forged name is dropped while the genuine host dispatch still arrives. The docs and comments no longer claim the prefix alone makes collisions impossible.

PR #5145, opened 2026-09-01
Replace legacy Airbnb ESLint configuration

gabek:
This replaces the legacy `.eslintrc.js` and Airbnb setup with a flat ESLint configuration that works with ESLint 10. The old config blocks dependency installation as its plugin peer ranges stop before ESLint 10.

- Move the Next.js, TypeScript, Storybook, and project-specific rules into `web/eslint.config.mjs`.
- Remove the legacy root and scoped ESLint configs plus `.eslintignore`.
- Keep Prettier separate from ESLint and update the changed-file hooks and scripts to include `.mjs`.
- Remove the unused Airbnb and direct ESLint plugin dependencies, then regenerate `web/package-lock.json`.

The web build, TypeScript check, ESLint, Prettier, and Knip checks pass. The local pre-push web build and Go tests also passed. The plugin integration hook could not run because `cargo` is not installed, so it was skipped when pushing.

## Required checklist


gabek:
Addressed in 546052649d. `npm run lint` now retains the explicit `--ext .js,.ts,.tsx` option while linting the config file and the `types/`, `pages/`, and `components/` directories. The command passes locally and the updated CI run passed.

